PR #26, opened 2026-03-13
refactor: unify node-kind tables & fix marker budget

dean0x:
## Summary

- Unifies the two separate node-kind tables (`to_static_node_kind` in `structure.rs` and `score_node_kind` in `utils.rs`) into a single `node_kind_info()` function in `utils.rs`, eliminating table drift risk
- Fixes marker budget bug in `truncate.rs` where non-contiguous spans could cause output to exceed `max_lines` — replaces fixed `MARKER_RESERVE = 2` with dynamic select-then-trim algorithm
- Self-review cleanup: destructured tuples, removed redundant re-sort, simplified unreachable guard

## Test plan

- [x] All 254 tests pass (103 core + integration/CLI)
- [x] `test_node_kind_info_consistency` validates all 41 known kinds map correctly
- [x] `test_noncontiguous_spans_marker_accounting` covers the fixed bug
- [x] `test_count_markers_*` (4 tests) cover empty, contiguous, gap, and edge cases
- [x] `test_trim_prefers_dropping_low_priority` and `test_trim_tiebreak_drops_last_position` verify trim algorithm
- [x] Clippy zero warnings

### DIFF
--- a/crates/rskim-core/src/transform/truncate.rs
+++ b/crates/rskim-core/src/transform/truncate.rs
@@ -96,16 +96,15 @@ pub(crate) fn truncate_to_lines(
     }
 
     // Score and sort spans: priority desc, position asc (tie-break)
-    let mut scored: Vec<(u8, usize, &NodeSpan)> = valid_spans
+    let mut scored: Vec<(u8, &NodeSpan)> = valid_spans
         .iter()
-        .enumerate()
-        .map(|(idx, span)| (score_node_kind(span.node_kind), idx, *span))
+        .map(|span| (score_node_kind(span.node_kind), *span))
         .collect();
     scored.sort_by(|a, b| {
         b.0.cmp(&a.0).then_with(|| {
-            a.2.transformed_range
+            a.1.transformed_range
                 .start
-                .cmp(&b.2.transformed_range.start)
+                .cmp(&b.1.transformed_range.start)
         })
     });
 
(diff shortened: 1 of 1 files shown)


crates/rskim-core/src/transform/truncate.rs
```diff
@@ -597,4 +669,152 @@ mod tests {
...
+        // If too many, should drop import (prio 3) before fn (prio 4)
+        let result = truncate_to_lines(text, &spans, Language::TypeScript, 3).unwrap();
+
+        assert!(
+            result.contains("type A"),
+            "Should keep highest priority (type): {:?}",
+            result
+        );
+        // If trimming happened, import should be dropped before function
+        if !result.contains("import B") && result.contains("fn foo()") {
```
dean0x: **Weak test assertions** (MEDIUM - Blocking)

The test uses `if/else if` conditional branches to check behavior, but the `else` case (wrong priority behavior) is silently accepted. The test would pass even if the algorithm drops higher-priority spans before lower-priority ones.

**Fix**: Add an explicit failure on wrong behavior with an `else` branch that panics on incorrect priority ordering.

crates/rskim-core/src/transform/signatures.rs
```diff
@@ -118,7 +118,7 @@ fn collect_signatures_with_kinds(
 
     if is_signature_node(kind, node_types) {
         if let Some(sig) = extract_signature(node, source, node_types)? {
-            let static_kind = crate::transform::structure::to_static_node_kind(kind);
+            let static_kind = crate::transform::utils::to_static_node_kind(kind);
```
dean0x: **Inconsistent import style** (MEDIUM - Should Fix)

`to_static_node_kind` uses a fully-qualified inline path (`crate::transform::utils::to_static_node_kind`) while other cross-module references like `NodeSpan` are imported via `use` statements. Since this PR changed the import path, it's a good opportunity to normalize to the shorter style.

**Fix**: Add `use crate::transform::utils::to_static_node_kind;` at the top and use the short name instead.

Issue #28, opened 2026-03-14
Tech debt: token budget cascade follow-ups


dean0x:
## Summary

**Audit 2026-03-27**: 7 of 9 remaining items resolved or closed. 2 items remain as performance optimization opportunities (not release-blocking).

---

## Performance

- ~~**B4**: Redundant tokenization at cascade/fallback boundary — cascade counts tokens, then `truncate_to_token_budget` re-counts the same string. Pass the already-computed count through.~~ ✅ Resolved in #29
- ~~**B5**: Binary search in `truncate_to_token_budget` builds candidate strings via `join` on every iteration. Use precomputed byte offsets for slice-based indexing instead.~~ ✅ Resolved in #29
- **B8**: Cascade re-parses tree-sitter AST for each mode. Parse once and reuse across mode transformations (requires core API changes).
- ~~**S5 (partial)**: `cascade_from_here` returns `&'static [Mode]` now (FIXED), but the cascade loop still creates a config per iteration.~~ ✅ Resolved — config construction is O(1) and intentional; mode differs per iteration.
- **B11**: Per-iteration tokenization in binary search — `count_tokens` is called O(log N) times inside the binary search loop, each O(N). A prefix-sum approach could reduce total to O(N), with trade-off of cross-line token boundary accuracy.
- ~~**B12**: Fast-path `text.to_string()` allocation — when text fits within budget, the function copies the entire string. Could return `Cow<str>` to avoid this allocation (requires API change).~~ ⊗ Closed — allocation overhead negligible; benchmarks prove no impact (14.6ms for 3000-line files, 3x under target).
- ~~**B13**: Split-then-rejoin round-trip — `text.lines().collect()` then `lines.join(\"\\n\")` reconstructs the input. Could scan `text` directly for newline byte positions and use `&str` slices, eliminating both the `Vec<&str>` and the `joined` String.~~ ✅ Resolved — binary search now uses precomputed byte offsets with zero-copy slicing.

## Complexity

- **B6**: `process_file` spans 120 lines with 5 responsibilities. Extract `try_cached_result` and `run_transform` helpers. `main()` spans 178 lines with cyclomatic complexity ~14. Extract `validate_args()` and unify dispatch.
[...]

PR #31, opened 2026-03-16
refactor: wave 1 tech debt from issue #28

dean0x:
## Summary

- **S13**: Extract `validate_bounded_arg` helper — 3 one-liner calls replace 50 lines of duplicated validation
- **S6**: Compose `ProcessOptions` inside `MultiFileOptions` — eliminates 5 duplicated fields
- **B6**: Extract `validate_args`, `read_and_validate`, `try_cached_result`, `run_transform` helpers — `process_file` drops from ~135 to ~30 lines
- **S5**: Hoist config creation out of cascade loop — create once, mutate `mode` per iteration
- **S2**: Add `effective_mode` to cache metadata — observability when cascade selects a different mode
- Simplifier pass: noise comment removal, idiomatic Rust patterns, TOCTOU fix in cache read

All internal refactors with no behavioral changes. Only `crates/rskim/src/main.rs` and `crates/rskim/src/cache.rs` modified.

Resolves S13, S6, B6, S5, S2 from #28.

## Test plan

- [x] `cargo test` — all 262 tests pass
- [x] `cargo clippy -- -D warnings` — clean
- [x] Scrutinizer 9-pillar review — PASS on all pillars

### DIFF
--- a/crates/rskim/src/cache.rs
+++ b/crates/rskim/src/cache.rs
@@ -111,9 +111,8 @@ fn cache_key(
     let canonical_display = canonical_path.display();
     let max_lines_str = fmt_opt(max_lines);
     let token_budget_str = fmt_opt(token_budget);
-    let hash_input = format!(
-        "{canonical_display}|{mtime_secs}|{mode:?}|{max_lines_str}|{token_budget_str}",
-    );
+    let hash_input =
+        format!("{canonical_display}|{mtime_secs}|{mode:?}|{max_lines_str}|{token_budget_str}");
 
     let mut hasher = Sha256::new();
     hasher.update(hash_input.as_bytes());
--- a/crates/rskim/src/main.rs
+++ b/crates/rskim/src/main.rs
@@ -386,8 +386,7 @@ fn try_cached_result(
         return Ok(None);
     }
 
-    let Some(hit) =
-        cache::read_cache(path, options.mode, options.max_lines, options.token_budget)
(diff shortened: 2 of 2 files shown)


crates/rskim/src/cache.rs
```diff
@@ -86,42 +85,32 @@ fn cache_key(
...
-    let cache_file = cache_dir.join(format!("{}.json", key));
-
-    // Check if cache file exists
-    if !cache_file.exists() {
-        return None;
-    }
+    let cache_file = get_cache_dir().ok()?.join(format!("{key}.json"));
 
-    // Read cache file
     let cache_content = fs::read_to_string(&cache_file).ok()?;
```
dean0x: ⚠️ **Should fix: Removed explicit `exists()` check on cache miss path** 

The old code called `cache_file.exists()` before `fs::read_to_string()`. The refactored code relies on `fs::read_to_string().ok()?` returning `None` when the file doesn't exist.

**Impact:** On the cache-miss hot path (every uncached file), this trades a cheap `stat()` syscall for a more expensive `open() + ENOENT` syscall chain. While functionally equivalent and the subsequent transformation dominates, this is a micro-optimization regression.

**Suggested fix:** Consider restoring the explicit check for the common case:

```rust
let cache_file = get_cache_dir().ok()?.join(format!("{key}.json"));
if !cache_file.exists() {
    return None;  // Fast return on cache miss
}
let cache_content = fs::read_to_string(&cache_file).ok()?;
```

Alternatively, document this as a justified simplification trade-off (code clarity vs. micro-optimization).

PR #69, opened 2026-03-20
feat: add gitignore support with --no-ignore override

dean0x:
## Summary

- Replace `glob` crate with `ignore` (from ripgrep/BurntSushi) for directory walking and glob pattern matching
- `.gitignore`, `.git/info/exclude`, global gitignore, and hidden files now respected by default
- Add `--no-ignore` CLI flag to include all files regardless of ignore rules
- `.gitignore` respected even outside git repos (`require_git(false)`)

## Changes

- **Dependencies**: `glob` → `ignore` + `globset` (globset is already a transitive dep of ignore)
- **`multi.rs`**: Core rewrite — `WalkBuilder` + `GlobMatcher` replacing recursive `visit_dir()` and `glob::glob()`
- **`main.rs`**: `--no-ignore` flag, `{` added to glob metacharacter detection
- **Tests**: 11 new integration tests (gitignore respect, `--no-ignore` override, hidden files, hidden dirs, gitignore without .git/) + 10 unit tests for `glob_walk_root()`

## Behavioral changes

- Hidden files/directories now **excluded by default** (matches ripgrep/fd convention). Use `--no-ignore` to include them.
- `.ignore` files (ripgrep convention) are also respected by default.
- Error messages now include hint: "Use --no-ignore to include all files" when gitignore filtering may be hiding results.

## Test plan

- [x] All 405 tests pass (`cargo test --all-features`)
- [x] `cargo clippy --all-features -- -D warnings` — zero warnings
- [ ] Manual: create temp repo with `node_modules/` in `.gitignore`, confirm `rskim .` skips it
- [ ] Manual: confirm `rskim . --no-ignore` includes everything
- [ ] Manual: confirm `rskim '**/*.ts'` respects gitignore

### DIFF
--- a/crates/rskim/src/main.rs
+++ b/crates/rskim/src/main.rs
@@ -230,8 +230,12 @@ struct Args {
     )]
     jobs: Option<usize>,
 
-    /// Don't respect .gitignore rules when scanning directories or globs
-    #[arg(long, help = "Don't respect .gitignore rules (include all files)")]
+    /// Don't respect .gitignore rules when scanning directories or globs.
+    /// Also includes hidden files and directories (dotfiles) that are excluded by default.
+    #[arg(
+        long,
+        help = "Don't respect .gitignore rules (include all files, including hidden/dotfiles)"
+    )]
     no_ignore: bool,
 
     /// Disable caching (caching is enabled by default for performance)


crates/rskim/src/multi.rs
```diff
@@ -172,77 +252,89 @@ fn process_files(
...
+    let mut builder = WalkBuilder::new(walk_root);
+    configure_walker(&mut builder, options.no_ignore);
 
-    fn visit_dir(dir: &Path, files: &mut Vec<PathBuf>) -> anyhow::Result<()> {
-        if !dir.is_dir() {
-            return Ok(());
-        }
+    let walk_root_path = Path::new(walk_root)
+        .canonicalize()
+        .unwrap_or_else(|_| PathBuf::from(walk_root));
```
dean0x: **Security observation**: The `canonicalize()` call on each entry resolves symlinks to their real paths. Although the `ignore` walker has `follow_links(false)` configured (preventing symlink entries from being yielded), `canonicalize()` on a non-symlink path could still resolve through a symlinked parent directory, creating a subtle defense-in-depth gap.

**Suggested improvement**: Either:
1. Add a comment documenting that this is safe because symlinks are never yielded by the walker, or
2. Reorder to prefer `strip_prefix` first (no syscall, no symlink resolution), falling back to `canonicalize` only when needed:

```rust
path.strip_prefix(walk_root)
    .ok()
    .or_else(|_| path.canonicalize().ok().and_then(|c| c.strip_prefix(&walk_root_path).ok()))
    .map(|p| p.to_path_buf())
    .is_some_and(|rel| matcher.is_match(&rel))
```

This also improves performance by avoiding syscalls in the hot path.

PR #151, opened 2026-04-21
fix(init): bare skim hooks, unified stdin/pipe infra, non-interactive install

dean0x:
## Summary

Three interconnected changes: hook script portability, unified stdin/pipe infrastructure across all subcommands, and non-interactive install UX.

### Change A — Bare `skim` in hook scripts
- Hook scripts now emit `exec skim rewrite --hook` (PATH-resolved) instead of hardcoded absolute binary paths — eliminates "unknown git subcommand" failures after upgrading skim via npm
- Format-aware idempotency detects old-format hooks and forces regeneration even when version matches
- Removed `validate_shell_safe_path()`, `SHELL_UNSAFE_CHARS`, and 12 associated tests (attack surface no longer exists)

### Change B — Unified stdin/pipe reading infrastructure
- New shared primitives in `cmd/mod.rs`: `should_read_stdin()`, `read_bounded()`, `read_stdin_bounded()`, `obtain_output()`
- `obtain_output()` provides a single entry point for all subcommands: read piped stdin OR spawn the real command, with empty-stdin fallthrough
- Bounded reads with 10 MB cap and lossy UTF-8 fallback, matching `runner.rs` pipe behavior

### Change C — Test runner deduplication
- `go.rs`, `pytest.rs`, `vitest.rs` refactored to use shared `run_passthrough()` and `try_read_stdin()` from `test/shared.rs`
- Eliminated duplicated passthrough/stdin/error-hint logic across all test parsers
- `emit_failure_context` helper centralized in `shared.rs`

### Change D — Infra `--json` flag disambiguation
- New `extract_infra_json_flag()` distinguishes skim's `--json` (boolean format flag) from `gh --json fields` (value pair forwarded to the tool)
- Scoped to infra subcommand only — global `extract_json_flag` unchanged to avoid breaking lint's `["--json", "eslint"]` pattern

### Change E — `runner.rs` expansion
- `run_with_env()` method for environment variable overrides during spawn
- Cascading binary resolution support

### Change F — Non-interactive `skim init`
- Removed non-TTY guard, interactive scope/marketplace prompts, and confirmation prompt from install flow
- `--yes` still accepted for backward compatibility (used by uninstall only)
- Removed `prompt_choice()` helper (dead code)

### Change G — New E2E tests
- `cli_test_pytest.rs` — pytest parser integration tests
- `cli_test_vitest.rs` — vitest parser integration tests
- 3 new unit tests for hook format detection (new/old/missing)

## Metrics

- 40 files changed, +1,228 / −816 (net −195 when excluding new test files)
- Test count: 2,800 → verified passing

## Test plan

- [x] `cargo build --all-features` — passes
- [x] `cargo clippy -- -D warnings` — zero warnings
- [x] `cargo test --all-features` — all tests pass
- [x] `skim init` detects old-format hook and regenerates
- [x] `skim init` (second run) reports "Already up to date"
- [x] Non-TTY install succeeds without `--yes`
- [x] Piped stdin flows through `obtain_output` correctly
- [x] `skim infra gh --json fields pr list` preserves `--json fields` for gh
- [x] Uninstall still prompts for confirmation


dean0x:
The `trim().is_empty()` call scans the full stdin buffer from both ends before deciding whether to use it. For the empty-pipe case this PR targets, this is harmless. For large legitimate inputs, this adds unnecessary O(n) work. 

Replace with a short-circuit pattern that avoids the full scan:

```rust
// Instead of: stdin_buf.trim().is_empty()
// Use: (no non-whitespace bytes exist)
if stdin_buf.bytes().any(|b| !b.is_ascii_whitespace()) {
    stdin_content = Some(stdin_buf);
}
```

This short-circuits on the first non-whitespace byte, making the empty case O(1).

**Performance Review Confidence**: 85%

dean0x:
The comparison `bytes_read as u64 >= MAX_STDIN_BYTES` mixes types: `usize` from `read_to_string` with a `u64` constant. The cast is safe in practice but obscures intent and adds cognitive overhead.

Fix by making `MAX_STDIN_BYTES` a `usize` to match the return type:

```rust
const MAX_STDIN_BYTES: usize = 64 * 1024 * 1024;
// ...
if bytes_read >= MAX_STDIN_BYTES {
    anyhow::bail!("stdin input exceeded 64 MiB limit");
}
```

Note: `take()` still needs `u64`, so the cast moves there: `.take(MAX_STDIN_BYTES as u64)`

**Rust Review Confidence**: 82%

dean0x:
The new tests at lines 187-206 use bare `Command::cargo_bin("skim")` without isolating environment variables. This is inconsistent with the 10 other test files in this PR that added `env_remove("SKIM_PASSTHROUGH")` and `env_remove("SKIM_DEBUG")`.

If `SKIM_PASSTHROUGH=1` is set in the developer's shell during `cargo test`, these tests may exercise the passthrough code path instead of the intended compression path, potentially masking regressions.

Add env isolation:
```rust
Command::cargo_bin("skim")
    .unwrap()
    .env_remove("SKIM_PASSTHROUGH")
    .env_remove("SKIM_DEBUG")
    .arg("test")
    .arg("vitest")
    .arg("run")
    // ...
```

**Consistency & Rust Review Confidence**: 85%

dean0x:
The `should_read_stdin()` helper was extracted here but the same `!std::io::stdin().is_terminal() && args.is_empty()` logic remains duplicated across the codebase:
- `test/cargo.rs:72` (inline expression)
- `file/mod.rs:125` (inline expression)
- `lint/mod.rs:153` (inline expression)
- `pkg/mod.rs:96` (inline expression)
- `infra/mod.rs:117` (as separate `should_use_stdin()` function)

This creates a consistency risk: a future fix to stdin detection logic must be applied in 6 separate locations. All six sites currently agree, but consolidation reduces drift surface.

**Recommendation**: Promote `should_read_stdin()` to `cmd/mod.rs` as `pub(crate)` and consolidate all call sites. At minimum, migrate `test/cargo.rs` (same `test/` module sibling) in this PR to use the shared helper.

**Architecture & Consistency Review Confidence**: 90%

dean0x:
Vitest has `test_vitest_with_args_does_not_read_stdin` validating the args-guard fix via the CLI boundary. Pytest received the identical fix (switching from bare `is_terminal()` to `should_read_stdin(args)` at both `pytest.rs:52` and `pytest.rs:74`), but pytest has no corresponding integration test in `cli_test_pytest.rs`.

The unit tests in `vitest.rs:845-875` cover the pure function, but they cannot exercise the pytest-specific code paths that call `should_read_stdin`. Add an integration test for pytest:

```rust
#[test]
fn test_pytest_with_args_does_not_read_stdin() {
    Command::cargo_bin("skim")
        .unwrap()
        .env_remove("SKIM_PASSTHROUGH")
        .env_remove("SKIM_DEBUG")
        .arg("test")
        .arg("pytest")
        .arg("-v")
        .assert()
        .failure()
        .stdout(predicate::str::is_empty().not());
}
```

The whole purpose of this branch is fixing the stdin detection bug—pytest is one of two direct consumers of the fix and should have end-to-end coverage.

**Testing Review Confidence**: 92%

dean0x:
The `should_read_stdin` function lives in `shared.rs` (line 75) and is used by both vitest and pytest. Its 3 unit tests (`test_should_read_stdin_false_when_args_present`, `test_should_read_stdin_args_gate_short_circuits`, `test_should_read_stdin_empty_args_defers_to_terminal`) are defined in `vitest.rs::tests` (lines 844-875) rather than in `shared.rs::tests` where the function is defined.

This violates the co-location principle: if someone modifies `should_read_stdin` in `shared.rs`, they would look at `shared.rs::tests` for coverage and would not realize the tests are in a different file. The `shared.rs::tests` module already has a `#[cfg(test)] mod tests` block (around line 200) with tests for `scrape_failures` and `last_n_lines`—`should_read_stdin` belongs there with the other shared helper tests.

Move the three `should_read_stdin` tests from `vitest.rs::tests` to `shared.rs::tests`.

**Consistency & Testing Review Confidence**: 85%

dean0x:
## PR Review Summary

**Overall Recommendation**: CHANGES_REQUESTED

This PR addresses a real stdin detection bug in test parsers (vitest, pytest) and the shared `run_parsed_command_with_mode` dispatcher. The core architectural decision—extracting `should_read_stdin()` as a shared helper and adding two-phase stdin detection with empty-stdin fallback—is sound and solves the bug correctly.

### Blocking Issues (6 findings, 80%+ confidence)

All 6 critical findings have been posted as separate comments above. Key themes:

1. **Type & Performance Cleanup** (2 issues)
   - Empty-stdin check: Replace `trim().is_empty()` with `bytes().any()` short-circuit
   - Type consistency: Change `MAX_STDIN_BYTES` from `u64` to `usize`

2. **Incomplete Consolidation** (2 issues)
   - Stdin detection logic duplicated across 5 other modules—promote `should_read_stdin()` to `cmd/mod.rs` and consolidate all call sites (or at minimum `test/cargo.rs` in this PR)
   - Tests placed in wrong file: Move `should_read_stdin` tests from `vitest.rs` to `shared.rs::tests`

3. **Test Coverage Gaps** (2 issues)
   - Missing pytest integration test (92% confidence)
   - Missing env isolation in new vitest tests (85% confidence)

### Score Summary by Reviewer
- **Security**: 9/10 — Approved
- **Architecture**: 7/10 — Changes requested
- **Consistency**: 5/10 — Changes requested
- **Testing**: 6/10 — Changes requested
- **Complexity**: 7/10 — Changes requested
- **Regression**: 8/10 — Approved with conditions
- **Rust**: 7/10 — Changes requested
- **Performance**: 7/10 — Approved with conditions

---

These are improvements worth considering but not blocking:

1. **Inconsistent empty-stdin fallback semantics** (82% confidence, MEDIUM) — `run_parsed_command_with_mode` uses `Option<String>` to gate the branch, while vitest/pytest use explicit fallback calls. Behavior is identical but implementation structure differs. Add a comment in `shared.rs` documenting the two codepaths and why they are separate to prevent accidental drift.

2. **Test helper env isolation inconsistent across all test files** (85% confidence, PRE-EXISTING) — After this PR, 10 test files have env-isolated `skim_cmd()` helpers but 8 still use bare `Command::cargo_bin("skim")`. Consider extracting a shared test utility crate function in a follow-up.

3. **Vitest/pytest passthrough path duplication** (85% confidence, MEDIUM) — Both files have the stdin-read-then-fallback pattern duplicated in passthrough and main paths. Extract a `try_read_stdin()` helper in `shared.rs` to eliminate duplication.

4. **bytes_read >= check is always true/false** (Performance note) — The `.take(MAX_STDIN_BYTES)` already caps the read, so `read_to_string` will never return MORE than `MAX_STDIN_BYTES`. The check only triggers if exactly `MAX_STDIN_BYTES` was read (stream was truncated). Comment this to clarify semantics.

---

### Verification Checklist

Before approving, confirm:
- [ ] All 6 inline comments addressed
- [ ] `test/cargo.rs:72` migrated to use `should_read_stdin()` 
- [ ] `should_read_stdin` tests moved to `shared.rs::tests`
- [ ] pytest integration test added to `cli_test_pytest.rs`
- [ ] New vitest tests have `env_remove` calls
- [ ] `trim().is_empty()` replaced with `bytes().any()` optimization
- [ ] `MAX_STDIN_BYTES` changed to `usize`
- [ ] All tests pass: `cargo test --all-features`

---

**Claude Code Review** | Comprehensive Multi-Agent Review (8 specialized reviewers across security, architecture, consistency, testing, complexity, regression, Rust, performance)

crates/rskim/src/cmd/mod.rs
```diff
@@ -222,15 +222,84 @@ pub(crate) struct ParsedCommandConfig<'a> {
...
+    install_hint: &str,
+    use_stdin: bool,
+) -> anyhow::Result<Option<CommandOutput>> {
+    const MAX_STDIN_BYTES: usize = 64 * 1024 * 1024;
+
+    if use_stdin {
+        let mut stdin_buf = String::new();
+        let bytes_read = io::stdin()
+            .take(MAX_STDIN_BYTES as u64)
+            .read_to_string(&mut stdin_buf)?;
```
dean0x: BLOCKING: Inconsistent UTF-8 handling

obtain_output uses io::stdin().take().read_to_string() which errors on invalid UTF-8, while read_stdin_raw (shared.rs:107) uses String::from_utf8 with lossy fallback. Both paths read piped stdin but have different failure modes.

Fix: Replace with the same chunked-read + lossy-UTF-8 pattern used in read_stdin_raw, or call it directly.

Confidence: 90% (flagged in security, architecture, performance, consistency, rust, testing reviews)

crates/rskim/src/cmd/mod.rs
```diff
@@ -222,15 +222,84 @@ pub(crate) struct ParsedCommandConfig<'a> {
...
+        if bytes_read >= MAX_STDIN_BYTES {
+            anyhow::bail!("stdin input exceeded 64 MiB limit");
```
dean0x: Inconsistent comparison operator for size limit

This location uses >= but shared.rs:102 uses > to check the same 64 MiB limit. The chunked-read path (>) is more correct since it rejects inputs that would exceed the buffer.

Confidence: 85% (flagged in security, architecture, consistency reviews)

crates/rskim/src/cmd/mod.rs
```diff
@@ -358,7 +363,6 @@ where
         crate::process::report_token_stats(orig, comp, "");
     }
```
dean0x, line 429: BLOCKING: Misleading hint for passthrough-tier results

Removing !result.is_passthrough() guard means the hint now fires on ALL non-zero exits, including passthrough results where output was NOT compressed. The message 'compressed output' is factually inaccurate for Tier 3 passthrough.

Fix: Use tier-neutral message instead:
if code != 0 {
    eprintln!("[skim] processed output (exit {code}). SKIM_PASSTHROUGH=1 for raw output.");
}

Confidence: 85% (flagged in architecture, consistency, regression, testing reviews)

dean0x:
## Code Review Summary — PR #151

**Review Coverage**: 8 specialized reviewers (security, architecture, performance, complexity, consistency, regression, testing, rust)

### Issues Requiring Action

#### BLOCKING (2 issues — must fix before merge)

1. **Inconsistent UTF-8 handling between stdin readers** (90% confidence)
   - `obtain_output` (mod.rs:242): Uses `read_to_string()` → errors on invalid UTF-8
   - `read_stdin_raw` (shared.rs:107): Uses lossy UTF-8 fallback
   - Impact: Subcommands using `obtain_output` (build, lint, pkg, git, infra, file, log) will error on non-UTF-8 stdin while test parsers handle it gracefully
   - Fix: Consolidate both to use the same chunked-read + lossy pattern

2. **Passthrough-tier results show misleading "compressed" message** (85% confidence)
   - Location: mod.rs:364 (uncommitted change)
   - Problem: Removing `!result.is_passthrough()` guard means hint fires on ALL non-zero exits, including passthrough results where output was NOT compressed
   - Impact: Agents may unnecessarily retry with `SKIM_PASSTHROUGH=1` when output was already uncompressed
   - Fix: Use tier-neutral message text that doesn't claim compression occurred

#### SHOULD FIX (2 issues — fix for consistency, not blocking)

3. **Inconsistent size-limit comparison operators** (85% confidence)
   - mod.rs:244 uses `>=` vs shared.rs:102 uses `>`
   - The chunked-read path (>) is more idiomatic
   - Fix: Align both to use `>`

4. **Inconsistent error messages for stdin size limit** (82% confidence)
   - mod.rs:244: `"stdin input exceeded 64 MiB limit"`
   - shared.rs:103: `"stdin exceeded 67108864 byte limit"` (raw bytes, different prefix)
   - log.rs:153: `"stdin input exceeded 64 MiB limit"`
   - Fix: Standardize all three to human-readable format

#### CONSIDERATIONS (Lower confidence suggestions for discussion)

- **Index-based while loop in `extract_infra_json_flag`** (65-70%) — Consider iterator-based approach with `peekable()` for idiomatic Rust
- **Hardcoded 300s timeout in `obtain_output`** (62%) — Could be externalized via env var for flexibility
- **Test comment references line numbers** (70%) — Use structural descriptions instead of line numbers to avoid drift
- **Missing unit test for whitespace-only stdin** (85%) — Consider extracting whitespace check into pure function for testability

### Summary Statistics

| Category | Count |
|----------|-------|
| Blocking issues | 2 |
| Should fix | 2 |
| Pre-existing (not blocking) | 3 |
| Suggestions (≤79% confidence) | 4 |
| **Total Issues Identified** | **11** |

### Reviewers & Consensus

All HIGH-confidence (≥80%) blocking issues were flagged by 3+ independent reviewers:
- UTF-8 inconsistency: 6 reviewers (security, arch, perf, consistency, rust, testing)
- Passthrough hint: 4 reviewers (arch, consistency, regression, testing)
- Size-limit operators: 3 reviewers (security, arch, consistency)
- Error message text: 3 reviewers (security, arch, consistency)

### Overall Assessment

**Status**: CHANGES_REQUESTED

The refactoring achieves solid deduplication (consolidating 4 duplicate stdin implementations into `try_read_stdin`) and extracting `obtain_output`/`render_output` for clarity. The zero-copy UTF-8 path is a genuine improvement. However, the two blocking issues must be resolved before merge: the inconsistent UTF-8 handling creates different failure modes across subcommand families, and the passthrough hint text is factually misleading. Both are high-confidence findings with broad consensus across reviewers.

---

*Review completed by Claude Code on 2026-04-23*
*Inline comments added to specific lines. See PR comments for detailed feedback.*

dean0x:
## Code Review Summary — PR #151 High-Confidence Findings

This is a clean consolidation refactoring with excellent architecture. All 2,802 tests pass. I've reviewed 8 specialized review reports and identified the following findings with ≥80% confidence:

### HIGH CONFIDENCE ISSUES (≥80%)

#### 1. Vitest Passthrough Behavior Change (85% confidence)
**File**: `crates/rskim/src/cmd/test/vitest.rs:41-48`

The vitest passthrough stdin path changed from `println!("{raw_output}")` to `print!("{raw}")` via the new `shared::run_passthrough()`. This removes a trailing newline.

**Assessment**: This is actually a **bug fix**, not a regression. The pytest path already used `print!` (correct behavior), and passthrough should not modify content. The vitest `println!` was arguably adding unwanted output.

**Recommendation**: Document this as intentional behavior alignment in the commit message if not already done.

---

#### 2. Missing Unit Tests for New Shared Functions (80-82% confidence)

Three new consolidation points lack unit test coverage:

**a) `read_stdin_bounded()` (cmd/mod.rs:67-83)**
- Contains non-trivial logic: chunked 8 KiB reads, boundary check, try-valid-then-lossy UTF-8
- Should mirror the test suite for `read_pipe()` in runner.rs
- **Suggestion**: Extract into generic `read_bounded(impl Read, max_bytes: usize) -> Result<String>` and add tests for:
  - Under limit case
  - Over limit failure
  - Invalid UTF-8 fallback path

**b) `shared::run_passthrough()` (cmd/test/shared.rs:104-119)**
- New consolidation point with two branches (stdin vs spawn) and exit-code logic
- **Suggestion**: Add unit tests for both branches using injected fakes

**c) `is_spawn_error()` tests incomplete (runner.rs:803-820)**
- Tests cover only 2 of 5 RunnerError variants
- **Suggestion**: Add tests for Io, PipeCaptureFailed, ReaderPanicked variants

---

#### 3. Parallel Constants (82% confidence)
**Files**: `crates/rskim/src/cmd/mod.rs:61` vs `crates/rskim/src/runner.rs:247`

Both `MAX_STDIN_BYTES` and `MAX_OUTPUT_BYTES` are 64 MiB with identical purpose and structure. The doc comment "mirrors" relationship has no compiler enforcement.

**Recommendation**: Extract a single constant (e.g., `pub(crate) const MAX_PIPE_BYTES: usize = 64 * 1024 * 1024;`) in runner.rs and re-export/use it consistently.

---

#### 4. FromUtf8Error Handling (82% confidence)
**File**: `crates/rskim/src/cmd/mod.rs:82`

Uses `e.as_bytes()` while the parallel `read_pipe()` in runner.rs:280 uses `e.into_bytes()`. The `as_bytes()` path forces an extra allocation.

**Fix**: Change to `&e.into_bytes()` to match runner.rs pattern and avoid allocation on invalid-UTF-8 path.

---

### ASSESSMENT

✅ **All 2,802 tests pass**
✅ Pure consolidation refactoring with zero regressions
✅ Security improvements: boundary check fix, timeout enforcement, type-safe error detection
✅ Architecture improvements: DRY consolidation, consistent patterns

The findings above are improvement suggestions, not blockers. The code is production-ready.

**Recommendation**: Address high-confidence items (parallel constants, UTF-8 handling, missing unit tests) in a follow-up PR or before merge if they're important for your quality bar.

---

*Review by Claude Code — 8 specialized analyses (security, architecture, performance, testing, rust, complexity, consistency, regression)*

PR #154, opened 2026-04-26
feat(discover,learn): responsive table truncation with --no-truncate flag

dean0x:
## Summary

Add terminal-width-aware table truncation for discover and learn subcommands to improve readability in narrow terminals.

## Changes

- Auto-detect terminal width and truncate wide columns to fit viewport
- Add --no-truncate flag to disable truncation and preserve full content
- Graceful fallback to untruncated output for non-TTY environments

## Breaking Changes

None

## Testing

- Verify --no-truncate flag works in discover command
- Verify --no-truncate flag works in learn command
- Test output in narrow terminals to confirm truncation
- Test in pipes/non-TTY to confirm graceful fallback

## Related Issues

None


dean0x:
## Code Review: PR #154

### ⚠️ HIGH (85% confidence)

**File: Cargo.toml:47** — `crossterm` default features
`crossterm` is added with default features (`events`, `bracketed-paste`), which pulls in `mio`, `signal-hook`, `signal-hook-mio`, and `parking_lot`. Only `crossterm::terminal::size()` is used, which doesn't require these features.

**Fix**: Disable default features:
```toml
crossterm = { version = "0.29", default-features = false }
```

dean0x:
### 🔴 BLOCKING TESTING ISSUES (90% confidence)

**Files: discover.rs, learn.rs** — Test coverage gaps

Two unit tests are missing critical assertions:

1. **`test_parse_args_defaults` in discover.rs** (line ~740) and **learn.rs** (line ~1452):
   - Missing: `assert!(!config.no_truncate);`
   - These tests verify all config field defaults but don't assert the new `no_truncate` field
   - If default were accidentally changed to `true`, no test would catch it

2. **`test_print_indented_table_indents_every_line` in ux.rs** (line ~201):
   - The test doesn't call the function under test after its signature changed
   - After adding `truncate: bool` parameter, the test still doesn't exercise the truncate logic
   - Needs to test both `truncate=true` (width constraint) and `truncate=false` (no constraint) paths

**Recommended fix**:
- Add `assert!(!config.no_truncate);` to both defaults tests
- Rewrite the table test to call `print_indented_table` with both `truncate` values and verify behavior

dean0x:
### ⚠️ MEDIUM ISSUES (80-85% confidence)

**Performance: Redundant `terminal_width()` syscalls** (discover.rs — 85% confidence)
- Called 3 separate times in `print_text_report()`: line 363, line 403, and inside `print_indented_table`
- Each call invokes `ioctl(TIOCGWINSZ)` syscall on Unix
- **Fix**: Compute once at the top of `print_text_report()` and pass down:
```rust
fn print_text_report(analysis: &DiscoverAnalysis, config: &DiscoverConfig) {
    let term_width = if config.no_truncate { 0 } else { 
        crate::cmd::ux::terminal_width() as usize 
    };
    // pass term_width to helper functions instead of recomputing
}
```

**Consistency: Hint placement differs between discover and learn** (90% confidence)
- discover.rs (line ~314-316): hint printed inline without blank-line separator
- learn.rs (line ~798-801): hint gets its own blank line before the table
- **Fix**: Adopt one pattern. The learn style (hint + blank line) is cleaner. Apply in discover too.

**Memory: Unnecessary String allocations in no-op path** (ux.rs, 82% confidence)
- `truncate_str()` and `truncate_path_middle()` return `s.to_string()` when `max == 0`
- When `--no-truncate` is set (common case), every call allocates unnecessarily
- **Fix**: Return `Cow<'_, str>` to avoid allocation in no-op case:
```rust
pub(crate) fn truncate_str<'a>(s: &'a str, max: usize) -> Cow<'a, str> {
    if max == 0 { return Cow::Borrowed(s); }
    // ... rest of logic
}
```

dean0x:
### 📋 MAINTAINABILITY: Duplicated truncation budget logic (85% confidence)

**Files: discover.rs (lines 358-366, 398-413), learn.rs (lines 803-814)**

Three near-identical blocks compute truncation budgets independently:
- Each recalculates overhead constants (23, 17, etc.) ad-hoc
- Each has its own inline comment explaining the math
- Each duplicates the `if truncate { ... } else { 0 }` pattern

Adding truncation to a fourth subcommand would require copy-pasting again, increasing drift risk.

**Suggested extraction** (to ux.rs):
```rust
/// Compute the per-column character budget for truncation.
/// Returns 0 (the no-op sentinel) when `active` is false.
/// `overhead` is indent + borders + padding + fixed columns.
pub(crate) fn column_budget(active: bool, overhead: usize, splits: &[usize]) -> Vec<usize> {
    if !active {
        return vec![0; splits.len()];
    }
    let available = (terminal_width() as usize).saturating_sub(overhead);
    let total_parts: usize = splits.iter().sum();
    splits.iter().map(|&s| (available * s / total_parts).max(1)).collect()
}
```

Then each call site becomes one line:
```rust
let [path_max] = column_budget(!config.no_truncate, 23, &[1])[..] else { unreachable!() };
```

This is not blocking (current code is correct), but is worth a follow-up refactor before a third consumer adopts truncation.

dean0x:
## Summary

**Comments Created**: 4
- 1 HIGH (85%): crossterm default features
- 1 BLOCKING (90%): Testing coverage gaps  
- 3 MEDIUM (80-85%): Performance, consistency, memory, maintainability

**Skipped (60-79% confidence)**: 
- Unicode width counting (68%)
- Dual-purpose max==0 sentinel (65%)  
- Hint indentation style (65%)
- Alternative terminal-size crate (65%)

**Pre-existing issues not addressed**:
- 57-line `print_commands_section` function (82% pre-existing, not blocking)
- No test for truncate_path_middle max==2 boundary (65%)
- No test for truncate_str multi-byte input (65%)
- Integration test fixture contract not documented (60%)

---

**Overall Assessment**: 
The feature is well-designed with strong architecture and comprehensive test coverage. All findings are either maintainability concerns or efficiency improvements — no correctness issues. The two BLOCKING testing items should be addressed before merge. The HIGH crossterm default-features issue is a quick one-line fix.

dean0x:
## Finding: Unnecessary heap allocation in `env_vars_contain_passthrough`

**File**: `crates/rskim/src/cmd/rewrite/engine.rs:169`
**Confidence**: 85-90% (flagged by architecture, rust, and performance reviewers)

### Issue
The call `check_passthrough_value(Some(v.to_string()))` allocates a `String` from an `&str` slice solely to match its `Option<String>` parameter signature. This runs in the hot path of every rewrite evaluation for commands with env-var prefixes.

The value `v` is already a `&str` from token splitting. Wrapping it in `Some(v.to_string())` forces an unnecessary heap allocation, and then `check_passthrough_value` immediately calls `.to_lowercase()` which allocates again.

### Recommended Fix
Add a `check_passthrough_str(val: &str) -> bool` overload in `cmd/mod.rs` and call it here:

```rust
fn env_vars_contain_passthrough(env_vars: &[&str]) -> bool {
    env_vars.iter().any(|token| {
        token
            .strip_prefix("SKIM_PASSTHROUGH=")
            .is_some_and(|v| crate::cmd::check_passthrough_str(v))
    })
}
```

Then add in `cmd/mod.rs`:
```rust
pub(crate) fn check_passthrough_str(val: &str) -> bool {
    matches!(val.to_ascii_lowercase().as_str(), "1" | "true" | "yes")
}
```

This eliminates both allocations while keeping the truthy-logic definition in one place.

---

*Code review feedback from feat/responsive-table-truncation*

dean0x:
## Finding: Inconsistent Cow deref pattern in table.add_row calls

**Files**: `crates/rskim/src/cmd/discover.rs:381`, `discover.rs:436-438`, `learn.rs:837-839`
**Confidence**: 85% (consistency reviewer flagged this)

### Issue
The `truncate_str` and `truncate_path_middle` functions now return `Cow<'_, str>`, but the code uses inconsistent patterns to convert the `Cow` into values for `table.add_row`:

- `discover.rs:381` uses `.as_ref()` explicitly
- `learn.rs:837-839` uses `.as_ref()` explicitly  
- `discover.rs:436-438` passes the `Cow` directly into a `vec![]` without `.as_ref()`

While both patterns work (because `Cow<str>` implements `Into<Cell>`), mixing them within the same PR creates a pattern deviation.

### Recommended Fix
Pick one convention and apply uniformly across all three call sites:

**Option A** (recommended - simpler): Drop all `.as_ref()` calls since `Cow<str>` works directly:
```rust
table.add_row(vec![truncate_str(...), ...])
```

**Option B** (explicit): Add `.as_ref()` to the `discover.rs:436-438` site:
```rust
table.add_row(vec![truncate_str(...).as_ref(), ...])
```

Consistency matters for code that will be extended by multiple developers.

---

*Code review feedback from feat/responsive-table-truncation*

dean0x:
## Finding: Missing test coverage for gh list skip flags

**File**: `crates/rskim/src/cmd/rewrite/tests.rs`
**Confidence**: 82% (testing reviewer flagged this)

### Issue
The new skip-flag combinations added to `gh list` commands are missing test coverage:

1. **`gh issue list --web`** — Now skipped by the new rules, but no test verifies this behavior
2. **`gh issue list --template`** and **`gh run list --template`** — Added to skip lists, but only `gh pr list --template` and `gh release list --template` have tests

Example of missing coverage:
- `gh issue list --web` should skip (new behavior)
- `gh run list --template '...'` should skip (new behavior)

### Recommended Tests
Add these test cases to `crates/rskim/src/cmd/rewrite/tests.rs`:

```rust
#[test]
fn test_gh_issue_list_web_skips() {
    assert_eq!(
        would_rewrite("gh issue list --web"),
        None,
        "gh issue list --web must skip rewrite (opens browser)"
    );
}

#[test]
fn test_gh_issue_list_template_skips() {
    assert_eq!(
        would_rewrite("gh issue list --template '{{.title}}'"),
        None,
        "gh issue list --template must skip rewrite"
    );
}

#[test]
fn test_gh_run_list_template_skips() {
    assert_eq!(
        would_rewrite("gh run list --template '{{.name}}'"),
        None,
        "gh run list --template must skip rewrite"
    );
}

/// gh run list does NOT support --web (not in skip list by design)
#[test]
fn test_gh_run_list_web_still_rewrites() {
    let result = would_rewrite("gh run list --web");
    assert!(
        result.is_some(),
        "gh run list --web must rewrite: --web is not supported by gh run list"
    );
}
```

---

*Code review feedback from feat/responsive-table-truncation*

dean0x:
## Finding: Unchecked cast in `print_indented_table`

**File**: `crates/rskim/src/cmd/ux.rs:192`
**Confidence**: 82% (Rust reviewer flagged this)

### Issue
The code uses `indent as u16` which is an unchecked integer narrowing cast. On 64-bit systems, if `indent > 65535` (unlikely for indentation but possible in pathological cases), the cast silently truncates.

The rest of the function defensively uses `saturating_sub`, but this cast is undefended.

### Recommended Fix
Use defensive casting consistent with the rest of the function:

```rust
let available = term_width.saturating_sub(u16::try_from(indent).unwrap_or(u16::MAX));
```

This makes the intent clear: if indent is somehow larger than `u16::MAX`, cap it at the maximum rather than silently truncating.

---

*Code review feedback from feat/responsive-table-truncation*

dean0x:
## Finding: Weak test coverage for `print_indented_table` branching

**File**: `crates/rskim/src/cmd/ux.rs:257-275`
**Confidence**: 85% (testing reviewer flagged this)

### Issue
The two new tests for `print_indented_table` only assert that the function doesn't panic:
- `test_print_indented_table_term_width_constrained_does_not_panic`
- `test_print_indented_table_no_truncate_does_not_panic`

This is the weakest form of smoke test. A no-panic test would still pass if the branching logic (`if term_width > 0`) were accidentally inverted or removed entirely.

### Recommended Fix
Capture actual table output and assert observable differences between branches:

```rust
#[test]
fn test_print_indented_table_constrained_limits_width() {
    let mut table = comfy_table::Table::new();
    table.set_header(["Col"]);
    table.add_row([&"x".repeat(200)]);
    
    // After set_width(76) = 80-4, lines should be shorter than unset
    let mut constrained = table.clone();
    constrained.set_width(76);
    let constrained_width = constrained.to_string().lines()
        .map(|l| l.chars().count()).max().unwrap_or(0);
    assert!(constrained_width < 200, "constrained table should limit line width");
}
```

This validates that:
1. When `term_width > 0`, `set_width()` is actually called
2. When `term_width == 0`, the full string renders without wrapping

---

*Code review feedback from feat/responsive-table-truncation*

dean0x:
## Summary: Lower-Confidence Suggestions (60-79%)

These findings fell below the 80% confidence threshold for inline comments but are worth mentioning for future improvements:

### 1. Extract repeated `term_width` computation pattern (82% confidence, but low urgency)
**Location**: `discover.rs:323-327`, `learn.rs:805-809`

Both files independently implement the same pattern:
```rust
let term_width: u16 = if config.no_truncate { 0 } else { terminal_width() };
```

Only 2 current occurrences, so not yet urgent. However, if a third subcommand adds table truncation, this pattern will be copied a third time. Consider extracting a helper like `resolve_term_width(no_truncate: bool) -> u16` in `ux.rs` before reaching 3+ call sites.

### 2. `truncate_str` Cow ergonomics (65% confidence)
**Location**: `learn.rs:837-839`, `discover.rs:381`

The `Cow<str>` return type is architecturally correct for zero-allocation on no-op. However, callers immediately call `.as_ref()` to convert to `&str` for `table.add_row()`. The allocation savings are real, but API ergonomics could be improved. This is minor and not blocking—just noted for future API design iterations.

### 3. Type consistency for terminal width (60% confidence)
**Location**: `ux.rs:172`

The `column_budget` function uses mixed types (`u16` for `term_width`, `usize` for `overhead` and return), requiring an internal cast. This is safe and well-documented by the sentinel convention, but a type alias like `type TermWidth = u16` would make the convention self-documenting at call sites. Low priority.

---

**Overall Assessment**: The PR is architecturally sound with solid refactoring patterns. The 5 high-confidence findings above are straightforward to address. Addressing them will bring the PR to "ready to merge" status.

*Code review summary from feat/responsive-table-truncation*

PR #163, opened 2026-05-07
feat: skim heatmap -- git history risk/coupling analysis

dean0x:
## Summary

- Adds `skim heatmap` subcommand that mines git history into a structured risk profile with 6 metrics: file churn, co-change coupling (blast radius), stability scores, author concentration (bus factor), fix-after-touch risk, and module encapsulation health
- Supports adaptive dual windowing (max of 90 days / 200 commits), auto-exclusion of lock files and build artifacts, JSON and text output, path scoping, and configurable thresholds
- Pure computation layer (zero I/O in metrics) with trait-based git abstraction for testability

## New files (7)
- `crates/rskim/src/cmd/heatmap/types.rs` — data structures
- `crates/rskim/src/cmd/heatmap/excludes.rs` — exclude patterns + GlobSet
- `crates/rskim/src/cmd/heatmap/git_source.rs` — GitDataSource trait + CLI impl
- `crates/rskim/src/cmd/heatmap/metrics.rs` — 6 pure metric functions
- `crates/rskim/src/cmd/heatmap/output.rs` — JSON + text rendering
- `crates/rskim/src/cmd/heatmap/mod.rs` — orchestration entry point
- `crates/rskim/tests/cli_heatmap.rs` — 16 integration tests

## Test plan

- [x] `cargo test --all-features` — 3,190 tests pass (87 new heatmap tests)
- [x] `cargo clippy -- -D warnings` — clean
- [x] `skim heatmap --json | jq .` — valid JSON with all 6 metrics
- [x] `skim heatmap` — text output with 5 sections
- [x] `skim heatmap --top 3 --json` — limits files in JSON output
- [x] `skim heatmap --debug` — diagnostics to stderr
- [x] Non-git-repo → error with exit 1
- [x] Empty repo → error with exit 1
- [x] `skim heatmap --window half` / `--window all` — presets work
- [x] Dual default windowing — `window.mode: "dual"` in JSON

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- **`skim heatmap` subcommand** — Git history risk and coupling analysis. Mines git log to produce 6 metrics: file churn, co-change coupling (blast radius), stability scores, author concentration (bus factor), fix-after-touch risk, and module encapsulation health. Supports adaptive dual windowing (max of 90 days / 200 commits), auto-exclusion of lock files and build artifacts, JSON and text output, and path scoping. 87 new tests.
+
 ### Changed
 
 ### Fixed
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -13,7 +13,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 ✅ **PHASE 3 COMPLETE** (100% of original roadmap)
 
 **What's Complete (Phases 1 & 2):**
-- ✅ Full Rust project with comprehensive test suite (3,103 tests passing)
+- ✅ Full Rust project with comprehensive test suite (3,190 tests passing)
 - ✅ 17 languages supported: TypeScript, JavaScript, Python, Rust, Go, Java, C, C++, C#, Ruby, SQL, Kotlin, Swift, Markdown, JSON, YAML, TOML
 - ✅ 4 transformation modes: structure, signatures, types, full
 - ✅ CLI with stdin/stdout streaming support
@@ -159,6 +159,7 @@ cargo fmt -- --check           # Format check
 - `learn` — Detect CLI error-retry patterns in agent sessions and generate correction rules (`--generate`, `--agent`, `--dry-run`, `--no-truncate`)
 - `rewrite` — Rewrite developer commands into skim equivalents (`--hook` for agent integration)
 - `stats` — Token analytics dashboard with per-session tracking (`--since`, `--format json`, `--verbose`, `--clear`)
+- `heatmap` — Git history risk/coupling analysis: churn, co-change coupling, stability scores, author concentration, fix-after-touch, module encapsulation (`--json`, `--since`, `--last`, `--window`, `--path`, `--top`)
 
 **Multi-category dispatchers:**
 - `cargo` — Rust toolchain compression: test, build, clippy, audit, nextest
@@ -568,7 +569,7 @@ Cross-platform builds require different GitHub Actions runners:
 5. Create test fixtures
 6. Validate AST access works
 
-**NOTE:** All phases complete (100%). Phase 3 features (multi-file glob, caching, parallel processing, token counting) are fully implemented and tested with 3,103 tests passing. See README.md for full usage guide.
+**NOTE:** All phases complete (100%). Phase 3 features (multi-file glob, caching, parallel processing, token counting) are fully implemented and tested with 3,190 tests passing. See README.md for full usage guide.
 
 ### Critical First File: `src/parser.rs`
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -159,6 +159,8 @@ cargo fmt -- --check           # Format check
 - `learn` — Detect CLI error-retry patterns in agent sessions and generate correction rules (`--generate`, `--agent`, `--dry-run`, `--no-truncate`)
 - `rewrite` — Rewrite developer commands into skim equivalents (`--hook` for agent integration)
 - `stats` — Token analytics dashboard with per-session tracking (`--since`, `--format json`, `--verbose`, `--clear`)
+
+**Analysis:**
 - `heatmap` — Git history risk/coupling analysis: churn, co-change coupling, stability scores, author concentration, fix-after-touch, module encapsulation (`--json`, `--since`, `--last`, `--window`, `--path`, `--top`)
 
 **Multi-category dispatchers:**
--- a/crates/rskim/src/analytics/mod.rs
+++ b/crates/rskim/src/analytics/mod.rs
@@ -44,6 +44,7 @@ pub(crate) enum CommandType {
     Infra,
     FileOps,
     Log,
+    Heatmap,
 }
 
 impl CommandType {
@@ -58,6 +59,7 @@ impl CommandType {
             CommandType::Infra => "infra",
             CommandType::FileOps => "fileops",
             CommandType::Log => "log",
+            CommandType::Heatmap => "heatmap",
         }
     }
 }
--- a/crates/rskim/src/cmd/heatmap/excludes.rs
+++ b/crates/rskim/src/cmd/heatmap/excludes.rs
@@ -1,0 +1,160 @@
+//! Default exclusion patterns and GlobSet construction for `skim heatmap`.
+
+use globset::{Glob, GlobSet, GlobSetBuilder};
+
+// ============================================================================
+// Default exclusion patterns
+// ============================================================================
+
+/// Hardcoded patterns for files that should be excluded from heatmap analysis
+/// by default (lock files, build artifacts, minified bundles, logs, etc.).
+pub(crate) const DEFAULT_EXCLUDES: &[&str] = &[
+    // Lock files
+    "Cargo.lock",
+    "package-lock.json",
+    "yarn.lock",
+    "pnpm-lock.yaml",
+    "poetry.lock",
+    "Pipfile.lock",
+    "composer.lock",
+    "Gemfile.lock",
+    "go.sum",
+    "*.lock",
+    // Build / dist artifacts
+    "dist/**",
+    "build/**",
+    "target/**",
+    "out/**",
+    ".next/**",
+    ".nuxt/**",
+    "__pycache__/**",
+    "*.egg-info/**",
+    // Minified / generated bundles
+    "*.min.js",
+    "*.min.css",
+    "*.bundle.js",
+    "*.chunk.js",
+    // Log files
+    "*.log",
+    "logs/**",
+    // Vendored dependencies
+    "vendor/**",
+    "node_modules/**",
+    // IDE / OS metadata
+    ".idea/**",
+    ".vscode/**",
+    ".DS_Store",
+    // Coverage / test artifacts
+    "coverage/**",
+    ".nyc_output/**",
+    // Migration generated files
+    "migrations/**/*.sql",
+];
+
+// ============================================================================
+// GlobSet construction
+// ============================================================================
+
+/// Build a [`GlobSet`] from the default excludes plus any extra patterns.
+///
+/// Returns an empty set when `no_exclude` is `true` so that no files are
+/// filtered out.
+pub(crate) fn build_exclude_set(no_exclude: bool, extra: &[String]) -> GlobSet {
+    if no_exclude {
+        return GlobSet::empty();
+    }
+
+    let mut builder = GlobSetBuilder::new();
+
+    for pattern in DEFAULT_EXCLUDES {
+        if let Ok(glob) = Glob::new(pattern) {
+            builder.add(glob);
+        }
+    }
+
+    for pattern in extra {
+        if let Ok(glob) = Glob::new(pattern) {
+            builder.add(glob);
+        }
+    }
+
+    builder.build().unwrap_or_else(|_| GlobSet::empty())
+}
+
+/// Return `true` when `path` matches any pattern in `set`.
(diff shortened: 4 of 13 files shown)


dean0x:
**File: `crates/rskim/src/cmd/heatmap/metrics.rs`** (line 100)

The nested loop here generates O(n²) pairs for each commit, with 4 string clones per pair. For a 500-file commit, this creates 250,000 pairs and ~1M allocations. Consider capping files-per-commit (e.g., skip coupling for >50 files) or using indexed file IDs instead of string keys to eliminate allocations.

---
*Claude Code Review* • Confidence: 80%+ only

dean0x:
**File: `crates/rskim/src/cmd/heatmap/mod.rs`** (line 506)

The error message says 'requires a positive integer', but `usize::from_str("0")` succeeds. Add validation after parsing to reject zero values.

---
*Claude Code Review* • Confidence: 80%+ only

dean0x:
**File: `crates/rskim/src/cmd/heatmap/mod.rs`** (line 567)

Positional arguments (non-flag args) are silently ignored. Running `skim heatmap src/` discards `src/` with no error. Either reject unexpected positional args or treat the first as `--path`.

---
*Claude Code Review* • Confidence: 80%+ only

dean0x:
**File: `crates/rskim/tests/cli_heatmap.rs`** (line 406)

This test wraps all assertions inside `if output.status.success()` which means it passes silently if the command fails. Assert the success unconditionally before validating the output:
```rust
assert!(output.status.success(), "stderr: {}", String::from_utf8_lossy(&output.stderr));
```

---
*Claude Code Review* • Confidence: 80%+ only

dean0x:
**File: `crates/rskim/tests/cli_heatmap.rs`** (line 48)

The `git_commit()` helper only checks if the process spawned, not if git exited 0. If git fails silently, the test repo is incomplete. Add exit code verification:
```rust
assert!(output.status.success(), "git commit failed: {}", String::from_utf8_lossy(&output.stderr));
```

---
*Claude Code Review* • Confidence: 80%+ only

dean0x:
**File: `crates/rskim/src/cmd/heatmap/mod.rs`** (line 484)

The heatmap `--debug` flag sets `config.debug = true` but doesn't initialize from `crate::debug::is_debug_enabled()` or call `crate::debug::force_enable_debug()`. Other subcommands (e.g., discover.rs) do both. This means `SKIM_DEBUG=1 skim heatmap` silently ignores the env var.

---
*Claude Code Review* • Confidence: 80%+ only

dean0x:
**File: `crates/rskim/src/cmd/heatmap/output.rs`** (line 12)

The custom `no_color()` checks only the `NO_COLOR` env var, but the `colored` crate already has built-in detection for `NO_COLOR`, `TERM=dumb`, piped output, and other edge cases. The stats.rs module doesn't have this custom function -- it relies on `colored`'s automatic behavior. Consider removing this function and letting `colored` handle color detection.

---
*Claude Code Review* • Confidence: 80%+ only

dean0x:
**File: `crates/rskim/src/cmd/heatmap/mod.rs`** (line 88)

This function accepts `&CliGitSource` for infrastructure checks (is_git_repo, get_repo_root, etc.) and `&dyn GitDataSource` for data fetching. Tests cannot mock the infrastructure path. Extract the infrastructure methods into a trait (e.g., GitInfraSource) so the full orchestration can be tested without a real git repo.

---
*Claude Code Review* • Confidence: 80%+ only

dean0x:
**File: `crates/rskim/tests/cli_heatmap.rs`** (line 332)

This test asserts that `new.rs` appears but never asserts that the old commit (`old.rs` at epoch 1,580,000,000) is absent. Add a negative assertion to verify the --since filter actually excludes old commits:
```rust
.stdout(predicate::str::contains("new.rs"))
.stdout(predicate::str::contains("old.rs").not());
```

---
*Claude Code Review* • Confidence: 80%+ only

dean0x:
**File: `crates/rskim/src/cmd/heatmap/mod.rs`** (line 189)

SystemTime::now() is called independently at lines 42, 189, 376, and 419. Each call can return a different value, so the 'now' used for stability scoring differs from the 'now' used for window calculation. Capture once at the top of `run()` and thread a single `now_epoch: u64` through all downstream functions.

---
*Claude Code Review* • Confidence: 80%+ only

dean0x:
**File: `crates/rskim/src/cmd/heatmap/output.rs`** (line 31)

This 205-line function has 5 rendering sections, each with duplicated `if use_color { ... } else { ... }` branches. The color and no-color paths produce nearly identical strings. Extract a helper to conditionally apply color:
```rust
fn colorize(text: &str, color_fn: fn(&str) -> colored::ColoredString, use_color: bool) -> String {
    if use_color { color_fn(text).to_string() } else { text.to_string() }
}
```

---
*Claude Code Review* • Confidence: 80%+ only

dean0x:
**File: `crates/rskim/src/cmd/heatmap/mod.rs`** (line 88)

COMPLEXITY: This orchestration function spans 228 lines (88-315) with 11 numbered steps inlined into one function body. Extract each step into a named helper: `validate_git_env`, `fetch_and_filter_commits`, `compute_all_metrics`, `assemble_file_metrics`. This brings each under 30 lines and makes the pipeline clear.

---
*Claude Code Review* • Confidence: 80%+ only

dean0x:
**File: `crates/rskim/src/cmd/heatmap/mod.rs`** (line 584)

When `--since` is the last argument without a value (e.g., `skim heatmap --since`), `extract_value` returns `None` and the error becomes 'unknown flag: --since' instead of 'missing value for --since'. Add a post-check in `parse_args` before the 'unknown flag' error to detect and report this properly.

---
*Claude Code Review* • Confidence: 80%+ only

dean0x:
**File: `crates/rskim/src/cmd/heatmap/mod.rs`** (line 324)

The `resolve_effective_config` function contains significant logic (flag precedence, dual-mode defaults, conflicting-flag warnings) but has no unit tests. The `GitDataSource` trait was introduced explicitly for testability, but no test double exists. Create a `FakeGitSource` and test the precedence logic, warning generation, and dual-mode behavior directly.

---
*Claude Code Review* • Confidence: 80%+ only

dean0x:
**File: `crates/rskim/tests/cli_heatmap.rs`** (line 376)

The assertion `.stderr(predicate::str::is_empty().or(predicate::str::contains("No commits")))` accepts both success and failure. Since the test creates commits 7 days ago (within the 14-day sprint window), the command should always succeed. Remove the `.or(contains("No commits"))` branch and assert success unconditionally.

---
*Claude Code Review* • Confidence: 80%+ only

dean0x:
## Code Review Summary (Lower-Confidence Findings 60-79%)

These suggestions are below the inline-comment threshold but worth considering:

### Performance (70% confidence)
- **Coupling pairs could use canonical ordering** — The `support_count` HashMap is symmetric, so using canonical ordering (e.g., `(a, b) where a <= b`) would halve entries without changing semantics.
- **Consider HashMap capacity pre-allocation** — Several metric functions create HashMaps with default capacity. Pre-allocating with `HashMap::with_capacity()` (estimated from input size) would reduce rehashing.

### Architecture (70% confidence)
- **Coupling computation cap suggestion** — For commits touching >50 files, consider skipping coupling analysis entirely (still counted in weighted_total). Auto-formatter commits produce noisy coupling signal anyway.

### Testing (70% confidence)
- **Missing test for `--format json` vs `--json` equivalence** — Both set `format_json = true`, but no test verifies they produce identical output.
- **No test for coupling weight discount** — The `1.0 / sqrt(files_in_commit)` weighting is a key design choice but has no test verifying it actually discounts large commits.

### Documentation & Consistency (70% confidence)
- **Analytics records empty strings** — The analytics call passes `String::new()` for both raw/compressed text. Since heatmap doesn't compress output, consider whether this zero-token row should be omitted entirely (like discover/learn/agents).
- **Coupling cross-module dependency** — The heatmap module imports `crate::cmd::session::types::parse_duration_ago`. Consider promoting to `cmd/mod.rs` for shared use.

---
*15 inline comments posted (92% high-confidence findings from architecture, rust, testing, performance, and consistency reviews)*

dean0x (commented):
## Code Review Summary

This PR includes 3 incremental commits (style/clippy fixes) that are mechanically sound but have a few documentation and testing gaps worth addressing.

### ✅ What's Good

- **Clippy improvements** (`sort_by_key` usage): Idiomatic Rust that improves readability (commits `0f30522`)
- **UFCS-to-method simplification** (`fd7591a`): Correct and idiomatic
- **Disjoint set optimization** (`compute_fix_after_touch`): Mathematically sound (divides the problem correctly)
- **All tests passing** (96 heatmap unit tests)

---

### 📋 Findings Requiring Action

#### 1. **Doc comment inaccuracy** (Confidence: 85% — SHOULD FIX)
**File**: `crates/rskim/src/cmd/heatmap/mod.rs`, lines 242-244

The doc comment on `compute_heatmap` states the function is "pure" (no I/O, no side effects), but the function:
- Calls `Instant::now()` (line 255) for timing measurement
- Conditionally writes to stderr via `eprintln!` (lines 268-277) when `config.debug=true`

While the previous batch's removal of the `debug: bool` parameter was positive (reducing redundancy), the remaining doc comment is now misleading. Consider updating to clarify the contract:

```rust
/// Compute all six risk metrics from commits and assemble a `HeatmapResult`.
///
/// No git I/O — all repository data is pre-fetched. Accepting `now_epoch` as a
/// parameter keeps computation deterministic and testable.
///
/// Debug timing is emitted to stderr when `config.debug` is enabled.
```

**Related review**: Architecture Review, section "Doc comment claims purity but function has side effects."

---

#### 2. **Missing test for fix-after-touch combined signal** (Confidence: 82% — SHOULD FIX)
**File**: `crates/rskim/src/cmd/heatmap/metrics.rs`, lines 376-379

The change from `keyword_set.union(&proximity_set).count()` to `keyword_count + proximity_set.len()` relies on a disjointness invariant (keyword indices vs. non-fix indices). The existing tests cover each signal independently but never test both firing simultaneously.

**Suggested test** (validates the arithmetic equivalence):
```rust
#[test]
fn test_fix_risk_combined_keyword_and_proximity() {
    // commit 0: feat touching a.rs (non-fix)
    // commit 1: fix touching a.rs (fix — keyword && within window of 0)
    // commit 2: feat touching a.rs (no fix following — not in proximity)
    let commits = vec![
        make_commit("h1", "Alice", 1, "feat: add", &["a.rs"]),
        make_commit("h2", "Alice", 2, "fix: bug", &["a.rs"]),
        make_commit("h3", "Alice", 3, "feat: more", &["a.rs"]),
    ];
    let fix_re = build_fix_regex();
    let result = compute_fix_after_touch(&commits, &fix_re, 5);
    let m = &result["a.rs"];
    // keyword_count=1 (commit 1), proximity={0} (commit 0→fix at 1)
    // combined = 1 + 1 = 2 out of 3 = 66.67%
    assert!((m.combined_pct - 66.666666).abs() < 0.01);
}
```

**Related review**: Testing Review, section "No unit test for `compute_fix_after_touch` combined_pct."

---

#### 3. **No unit tests for `build_window_info`** (Confidence: 80% — SHOULD FIX)
**File**: `crates/rskim/src/cmd/heatmap/mod.rs`, lines 422-466

The `build_window_info` function simplification removed tuple destructuring but has branching logic (dual mode, preset, count, time) that lacks direct unit tests. Only exercised indirectly through E2E CLI tests.

**Suggested tests** (major branches):
```rust
#[test]
fn test_build_window_info_dual_mode() {
    let config = HeatmapConfig {
        dual_mode: true,
        dual_time_since: Some(100),
        dual_count_since: Some(200),
        ..HeatmapConfig::default()
    };
    let info = build_window_info(&config, 50, 1_000_000);
    assert_eq!(info.mode, "dual");
    assert!(info.time_commits.is_none());
    assert!(info.count_commits.is_none());
}
```

**Related review**: Testing Review, section "No unit tests for `build_window_info`."

---

### 📌 Lower-Confidence Suggestions (60-79%)

- **Commit message scope** (`76a3ebc`, Confidence: 70%): The "style: apply rustfmt formatting fixes" commit also includes non-formatting refactorings (debug param removal, tuple simplification, union optimization). Consider splitting into separate commits for easier bisecting.
- **Sort idiom consistency** (`output.rs:50`, `mod.rs:311`, Confidence: 82%): The selective `sort_by_key` application is correct (remaining `sort_by` calls use f64/partial_cmp and can't use sort_by_key). Consider adding a brief inline comment explaining the f64 limitation to prevent confusion.

---

### Pre-existing Issues Not Blocking This PR

- **`WindowInfo` dead fields** (`mod.rs:462-463`, Confidence: 82%): `time_commits` and `count_commits` are always `None` in JSON output. Worth resolving before schema stabilizes.
- **Redundant `extract_top_dir` calls** (`metrics.rs:433-459`, Confidence: 85%): Called twice per file in the coupling loop. Informational.
- **String allocation in coupling pairs** (`metrics.rs:107-116`, Confidence: 82%): O(n²) loop allocates 4900+ strings for 50-file commits. Informational.

---

### Recommendation

**APPROVED WITH CONDITIONS**

The three incremental commits are solid cleanup. The actionable items are:
1. Update the `compute_heatmap` doc comment (10 minutes)
2. Add test for combined keyword+proximity fix-risk signal (15 minutes)
3. Add unit tests for `build_window_info` (20 minutes)

These strengthen the codebase before heatmap JSON output stabilizes. All 96 existing tests pass; no regressions detected.

---

**Reviewed by**: Code Review Agents (Architecture, Consistency, Performance, Testing, Rust, Regression)
**Confidence levels**: BLOCKING (100%), SHOULD_FIX (80%+), Pre-existing (82%+)

Issue #182, opened 2026-05-09
Wave 1d: Index builder (file walker, incremental updates)


dean0x:
## Overview
Build the indexing pipeline: walk codebase, parse with tree-sitter, classify fields, extract n-grams, write index.

## Pipeline
File walker (ignore crate, respects .gitignore) → Language detection (Language::from_path) → tree-sitter parse → Field extraction (walk AST, classify into 7 fields) → N-gram extraction per field → Postings accumulation → Index serialization (two-file write)

## Reuse existing infrastructure
- File walking: `ignore` crate (already used in multi.rs)
- Language detection: `Language::from_path` from rskim-core
- Parallelism: rayon for per-file processing
- Cache pattern: SHA256+mtime from cache.rs

## Incremental updates
- Metadata file: `~/.cache/skim/search/<repo-hash>/meta.json` with file→(mtime, content_hash)
- On re-index: compare, only re-process changed files
- Content hash (SHA256) for branch-switch correctness (not just mtime)

## Edge cases
- Skip files >5MB, skip non-UTF-8, skip minified (avg line >500 chars)
- Atomic writes: build to temp dir, atomic rename
- Signal handling: clean up temp on Ctrl+C

## Dependencies
- #6 (index format to write to)

## Acceptance Criteria
- [ ] Index 1000 files in < 5 seconds
- [ ] Incremental re-index (10 changed files) in < 500ms
- [ ] Respects .gitignore, skips binaries
- [ ] Index size < 10% of source
- [ ] All 12 languages handled (9 tree-sitter + 3 serde as full-text)

## Implementation Guidelines Checklist
- [ ] No file exceeds 400 lines
- [ ] Every public function has /// doc comment
- [ ] All dependencies injected
- [ ] All fallible operations return Result<T, SearchError>
- [ ] Module has #[cfg(test)] mod tests with: happy path, edge case, error path
- [ ] No hardcoded language-specific match arms

Issue #168, opened 2026-05-07
v2.5: dig/nslookup DNS output compression


dean0x:
# v2.5: dig/nslookup — DNS Output Compression

**Parent:** #96 (Command Coverage Expansion)
**Priority:** Low — moderate agent frequency, high noise-to-signal ratio
**Depends on:** Nothing (v2.x stable codebase)

## Overview

DNS diagnostic tools produce structured but verbose output. `dig` in particular has multiple sections (QUESTION, ANSWER, AUTHORITY, ADDITIONAL) plus statistics that are mostly noise for agents debugging connectivity.

### `dig`
- **Output:** Sectioned text (`;; QUESTION SECTION:`, `;; ANSWER SECTION:`, etc.)
- **Tier 1:** Extract ANSWER section records, query time, server — skip AUTHORITY/ADDITIONAL/stats
- **Tier 2:** Passthrough
- **Key insight:** ANSWER section is almost always the only thing agents care about

### `nslookup`
- **Output:** `Server: ...`, `Address: ...`, `Name: ...` lines
- **Tier 1:** Extract resolved name/address pairs
- **Tier 2:** Passthrough

## Implementation Notes

- New module at `cmd/infra/dns.rs` (shared for both tools)
- Rewrite rules: `dig` → `skim dig`, `nslookup` → `skim nslookup`
- dig sections are delimited by `;; SECTION_NAME SECTION:` markers — reliable regex targets
- For `dig +short`, output is already minimal — passthrough is correct

## Acceptance Criteria

- [ ] dig: ANSWER section extracted, noise sections stripped
- [ ] nslookup: name/address pairs extracted
- [ ] `dig +short` passes through unchanged (already compact)
- [ ] Rewrite rules added
- [ ] Test fixtures for each tool
- [ ] All existing tests pass

dean0x:
Implemented in a686e78 (PR #228)

Issue #167, opened 2026-05-07
v2.5: make build output compression


dean0x:
# v2.5: make — Build Output Compression

**Parent:** #96 (Command Coverage Expansion)
**Priority:** Medium — agents use make extensively in C/C++/Go projects
**Depends on:** Nothing (v2.x stable codebase)

## Overview

`make` produces verbose build output that is mostly noise for agents: compiler invocation lines, linker flags, entering/leaving directory messages. The valuable signal is errors, warnings, and the final success/failure status.

## Output Characteristics

- Recursive make: `make[N]: Entering directory ...` / `Leaving directory ...` noise
- Compiler lines: `gcc -c -o foo.o foo.c -Wall -Wextra -I/usr/include ...` (long flag chains)
- Errors: `file.c:42:10: error: ...` (GCC/Clang format)
- Warnings: `file.c:42:10: warning: ...`
- Nothing-to-do: `make: Nothing to be done for 'all'.`
- Failure: `make: *** [Makefile:17: target] Error 1`

## Tiers

- **Tier 1 (Full):** Extract error/warning lines with `file:line:col` format, count total targets built, surface make failure lines
- **Tier 2 (Degraded):** Strip entering/leaving directory noise, strip compiler invocation lines, keep errors + last N lines
- **Tier 3 (Passthrough):** Raw output

## Implementation Notes

- New module at `cmd/build/make.rs`
- Rewrite rule: `make` → `skim make`, `make -j` → `skim make`
- GCC/Clang error format regex (`file:line:col: error/warning:`) is well-standardized
- Could share error-line regex with future `gcc`/`clang` direct parsers
- Recursive make directory messages are pure noise — always strip

## Acceptance Criteria

- [ ] Tier 1 extracts errors and warnings with file:line:col
- [ ] Recursive make noise stripped
- [ ] Compiler invocation lines stripped (keep only errors/warnings/summary)
- [ ] Rewrite rule added
- [ ] Test fixtures: clean build, build with errors, build with warnings, nothing-to-do, recursive make
- [ ] All existing tests pass

dean0x:
Closed — merged via PR #223 (feat(#167): GNU Make build output compression).

PR #229, opened 2026-05-17
feat(#182): index builder pipeline — walk, manifest, BM25F indexing, incremental updates

dean0x:
## Problem

Search index requires a pipeline to walk codebases, classify files, and build the n-gram index with incremental update support.

## Key Changes

New `cmd/search/` module with pipeline, manifest, and walker. JSONL sidecar for FileId→path mapping and cached field maps. Two-phase processing: parallel tree-sitter classification, sequential builder accumulation.

## Breaking Changes

None.

## Reviewer Focus

Incremental diff logic in pipeline.rs, atomic write ordering (.skpost → .skidx → .skfiles), skip condition thresholds, fail-soft error handling.


dean0x:
### ⚠️ BLOCKING: Help Delegation Bug in Search Dispatcher

**File**: `crates/rskim/src/cmd/search/mod.rs` (line 34)
**Confidence**: 95% (Regression review)

`skim search index --help` is intercepted by the parent dispatcher and prints the wrong help text. The `args.iter().any(...)` scan finds `--help` at position 1 (the subcommand help) and prints parent help instead of delegating to `index::run(["--help"])`.

**Fix**: Dispatch to subcommands first, then check for parent-level `--help`:

```rust
if args.is_empty() {
    print_help();
    return Ok(ExitCode::SUCCESS);
}

// Dispatch subcommands first — let them handle their own --help
if args.first().is_some_and(|a| a == "index") {
    let rest = &args[1..];
    return index::run(rest);
}

// Top-level help (only when not a subcommand)
if args.iter().any(|a| matches!(a.as_str(), "--help" | "-h")) {
    print_help();
    return Ok(ExitCode::SUCCESS);
}
```

dean0x:
### ⚠️ BLOCKING: Unsafe `as u32` Cast on FileId (Rust/Reliability HIGH)

**File**: `crates/rskim/src/cmd/search/index.rs` (line 217)
**Confidence**: 90% (Rust), 85% (Reliability)

The cast `FileId(idx as u32)` silently truncates if `idx >= u32::MAX`. While the default `max_files` is 50,000, the `--max-files` flag accepts arbitrary `usize` values. A user could pass `--max-files=5000000000`, causing the cast to wrap and produce duplicate `FileId` values, corrupting the index.

**Fix**: Use `try_from` with explicit error handling:

```rust
let file_id = FileId(
    u32::try_from(idx)
        .map_err(|_| anyhow::anyhow!("file index {idx} exceeds u32::MAX"))?
);
builder.add_file_classified(file_id, &rf.content, rf.lang, field_map)?;
```

This aligns with how `file_count` and `skipped_count` are already handled on lines 172, 211, 236.

dean0x:
### ⚠️ BLOCKING: `--max-files=0` Validation (Security/Reliability MEDIUM)

**File**: `crates/rskim/src/cmd/search/index.rs` (lines 97-101)
**Confidence**: 85% (Security), 82% (Reliability)

Zero is accepted as a valid value despite the error message saying "requires a positive integer". Passing `--max-files=0` causes an empty index to be silently created with no error.

**Fix**: Add zero validation after parsing:

```rust
} else if let Some(val) = next_value(args, &mut i, "--max-files")? {
    let n = val.parse::<usize>()
        .map_err(|_| anyhow::anyhow!("--max-files requires a positive integer"))?;
    if n == 0 {
        anyhow::bail!("--max-files requires a positive integer");
    }
    max_files = Some(n);
```

dean0x:
### ⚠️ BLOCKING: Fragile Debug-Based Language Serialization (Rust/Performance HIGH)

**File**: `crates/rskim/src/cmd/search/index.rs` (line 230)
**Confidence**: 85% (Rust), 82% (Performance)

`format!("{:?}", rf.lang).to_lowercase()` relies on the `Debug` trait output of the `Language` enum for the manifest `lang` field. The `Debug` output is not a stable API — if a variant is renamed (e.g., `TypeScript` → `Typescript`), the manifest will silently produce different strings, breaking incremental cache hits for all existing manifests.

**Fix**: Add a stable method to `Language`:

```rust
// In types.rs or rskim-core:
impl Language {
    pub fn as_str(&self) -> &'static str {
        match self {
            Language::TypeScript => "typescript",
            Language::Python => "python",
            // ... etc
        }
    }
}

// Then in index.rs line 230:
lang: rf.lang.as_str().to_owned(),
```

dean0x:
### ⚠️ BLOCKING: TOCTOU Race in File Size Check (Security HIGH)

**File**: `crates/rskim/src/cmd/search/walk.rs` (lines 147-166)
**Confidence**: 82% (Security)

`fs::metadata(abs_path)` checks file size at line 147, then `fs::read_to_string(abs_path)` reads at line 166. Between these calls, a file could be replaced with a much larger one (symlink swap, rapid write). Since `read_to_string` reads the entire file into memory without a size limit at the time of read, this could cause OOM.

**Fix**: Check size on the open file handle:

```rust
let file = fs::File::open(abs_path)?;
let meta = file.metadata()?;
if meta.len() > MAX_FILE_BYTES {
    skipped.push(SkipReason::TooLarge { ... });
    continue;
}
let mut content = String::with_capacity(meta.len() as usize);
use std::io::Read;
file.take(MAX_FILE_BYTES + 1).read_to_string(&mut content)?;
if content.len() as u64 > MAX_FILE_BYTES {
    // File grew between metadata and read
    skipped.push(SkipReason::TooLarge { ... });
    continue;
}
```

dean0x:
### ⚠️ BLOCKING: Mixed Error Return Types (Architecture/Consistency HIGH)

**File**: `crates/rskim/src/cmd/search/manifest.rs` (lines 109, 205)
**Confidence**: 84% (Architecture), 88% (Consistency)

`FileManifest::load()` returns `std::io::Result<Self>` while `FileManifest::save()` returns `anyhow::Result<()>`. The rest of the CLI codebase uses `anyhow::Result` uniformly (see `heatmap/mod.rs`, `discover.rs`, etc.). Mixing error types forces callers to handle two error domains and is inconsistent with established patterns.

**Fix**: Unify both methods to return `anyhow::Result`:

```rust
// manifest.rs
pub(super) fn load(project_root: PathBuf, cache_dir: PathBuf) -> anyhow::Result<Self> {
    // ... same logic, but anyhow wraps io::Error automatically via ?
}

pub(super) fn save(&self) -> anyhow::Result<()> { ... }

// walk.rs
pub(super) fn discover_project_root(start: &Path) -> anyhow::Result<PathBuf> { ... }
pub(super) fn walk_and_read(...) -> anyhow::Result<(Vec<ReadFile>, Vec<SkipReason>)> { ... }
```

dean0x:
### ⚠️ BLOCKING: Silent Error Swallowing in Classification (Architecture HIGH)

**File**: `crates/rskim/src/cmd/search/index.rs` (lines 250-256)
**Confidence**: 83% (Architecture), 82% (Rust)

`run_classify(content, lang).unwrap_or_default()` silently discards classification errors, returning an empty field map. Files that fail classification (grammar regressions, edge cases) are indexed with zero field annotations, degrading search ranking quality with no diagnostic signal.

**Fix**: Log classification failures when debug mode is enabled and return both field map and success flag:

```rust
fn run_classify(
    content: &str,
    lang: rskim_core::Language,
) -> (FieldMap, bool) {
    match classify_source(content, lang) {
        Ok(fm) => (fm, true),
        Err(e) => {
            if crate::debug_enabled() {
                eprintln!("skim: classify failed for {:?}: {e}", lang);
            }
            (Vec::new(), false)
        }
    }
}
```

dean0x:
### ⚠️ BLOCKING: Hand-Rolled Argument Parser vs Clap (Architecture HIGH)

**File**: `crates/rskim/src/cmd/search/index.rs` (lines 83-127)
**Confidence**: 85% (Architecture)

The rest of the CLI uses `clap` with derive API (documented in CLAUDE.md). This subcommand introduces a hand-rolled `parse_args` with manual `--flag=val` / `--flag val` parsing and a `next_value` helper. This creates an inconsistent pattern — future contributors must maintain two parsing paradigms.

**Fix**: Migrate to clap derive:

```rust
#[derive(clap::Parser)]
struct IndexArgs {
    #[arg(long)]
    root: Option<PathBuf>,
    #[arg(long)]
    force: bool,
    #[arg(long)]
    max_files: Option<usize>,
    #[arg(long, hide = true)]
    index_dir: Option<PathBuf>,
}

// Then in run():
let args = IndexArgs::parse_from(args);
let config = IndexConfig { ... };
```

dean0x:
### ⚠️ BLOCKING: All File Contents Held in Memory (Performance HIGH)

**File**: `crates/rskim/src/cmd/search/index.rs` (line 171), `walk.rs` (lines 89-198)
**Confidence**: 90% (Performance)

`walk_and_read` reads every file into memory as `ReadFile.content: String` and returns them all in a `Vec`. With 50,000 files averaging 10 KB each, this is ~500 MB of heap-resident strings. The entire `read_files` vector lives alongside the `classified` vector until `build_index` completes, causing significant memory pressure and poor cache locality.

**Fix**: Process files in batches or separate walk (collect paths + SHA-256) from read (just-in-time during classification):

```rust
// Minimal mitigation: pre-allocate and batch process
let mut files: Vec<ReadFile> = Vec::with_capacity(max_files.min(4096));
// Consider batching: process 1,000 files at a time
```

This is a future optimization opportunity. The current approach is correct but will show memory impact at scale.

dean0x:
### ⚠️ BLOCKING: Redundant `fs::metadata` Syscalls (Performance HIGH)

**File**: `crates/rskim/src/cmd/search/walk.rs` (line 147)
**Confidence**: 92% (Performance)

`walk_and_read` calls `fs::metadata(abs_path)` for every file to check size. However, the `ignore::DirEntry` from the walker already provides a `.metadata()` method that may reuse cached metadata from directory traversal. On a 50,000-file project, this is 50,000 unnecessary syscalls.

**Fix**: Use the walker's metadata:

```rust
// Before (redundant syscall):
let metadata = match fs::metadata(abs_path) { ... };

// After (reuse walker metadata):
let metadata = match entry.metadata() {
    Ok(Ok(m)) => m,
    _ => {
        skipped.push(SkipReason::ReadError { ... });
        continue;
    }
};
```

dean0x:
### ⚠️ BLOCKING: Function Complexity Thresholds (Complexity HIGH - TWO FUNCTIONS)

**File 1**: `crates/rskim/src/cmd/search/index.rs` (lines 159-244, 85 lines)
**File 2**: `crates/rskim/src/cmd/search/walk.rs` (lines 89-198, 106 lines)
**Confidence**: 85% (index.rs), 82% (walk.rs)

Both functions exceed the 50-line complexity threshold with cyclomatic complexity around 8-12. While each step is cohesive, combining them into one function means readers must hold all 6-7 stages in working memory simultaneously.

**Suggestions**:
- **index.rs**: Extract classify phase and manifest-write phase into named helpers
- **walk.rs**: Extract per-file processing logic into `try_read_file()` helper that returns `Result<ReadFile, SkipReason>`

Example for index.rs:

```rust
fn classify_files(read_files: &[ReadFile], manifest: &FileManifest) -> Vec<ClassifiedFile> {
    // ... classification logic
}

fn write_manifest(config: &IndexConfig, cache_dir: PathBuf, 
                  read_files: &[ReadFile], classified: &[ClassifiedFile]) -> anyhow::Result<()> {
    // ... manifest writing logic
}
```

This would bring `build_index` down to ~40 lines as a clean orchestration sequence.

dean0x:
### ⚠️ BLOCKING: Missing Cache Hit Verification (Testing HIGH)

**File**: `crates/rskim/src/cmd/search/index_tests.rs` (lines 112-124)
**Confidence**: 90% (Testing)

`test_index_incremental_second_build_faster_or_same()` only asserts `ExitCode::SUCCESS` but never verifies that cache hits actually occurred. The `IndexResult` containing `cache_hits`, `file_count`, `skipped` is discarded before reaching tests. The incremental build path is not actually verified — a second build that re-classified every file would still pass.

**Fix**: Expose `build_index` results or inspect manifest file to verify entries match:

```rust
#[test]
fn test_index_incremental_second_build_has_cache_hits() {
    let project = make_project();
    let cache = tempfile::tempdir().unwrap();
    let args = index_args(project.path(), cache.path());

    run(&args).unwrap(); // first build
    run(&args).unwrap(); // second build — should have cache hits

    // Verify manifest exists with entries for all source files
    let manifest_path = find_file_path_with_ext(cache.path(), "skfiles").unwrap();
    let contents = fs::read_to_string(manifest_path).unwrap();
    let lines: Vec<&str> = contents.lines().collect();
    // 1 header + 3 source files
    assert!(lines.len() >= 4, "manifest should have header + 3 entries");
}
```

dean0x:
### ⚠️ BLOCKING: No Direct Test for Minification Detection (Testing HIGH)

**File**: `crates/rskim/src/cmd/search/walk.rs` (lines 217-228)
**Confidence**: 85% (Testing)

The `is_minified()` function is a non-trivial heuristic (probe first 8KB, count newlines, check average line length > 500 bytes) but has no direct test. A regression here would silently index minified bundles, degrading index quality.

**Fix**: Add a test in `walk_tests.rs`:

```rust
#[test]
fn test_walk_skips_minified_files() {
    let dir = tempfile::tempdir().unwrap();
    let root = dir.path();
    fs::create_dir_all(root.join(".git")).unwrap();

    // Create a "minified" JS file: one long line > 500 avg bytes/line
    let minified = "var a=1;".repeat(200); // 1600 bytes, 0 newlines
    fs::write(root.join("bundle.js"), &minified).unwrap();
    fs::write(root.join("normal.rs"), "fn f() {}\n").unwrap();

    let (files, _skipped) = walk_and_read(&root.canonicalize().unwrap(), 50_000).unwrap();
    let paths: Vec<_> = files.iter().map(|f| f.rel_path.clone()).collect();

    assert!(!paths.iter().any(|p| p.ends_with("bundle.js")), "minified should be skipped");
    assert!(paths.iter().any(|p| p.ends_with("normal.rs")), "normal should be included");
}
```

dean0x:
## Summary: Code Review Findings

### Overall Recommendation: **CHANGES REQUESTED**

Comprehensive review across 9 dimensions (Security, Architecture, Performance, Complexity, Consistency, Regression, Testing, Reliability, Rust) identified **13 blocking issues** at ≥80% confidence. Below is the consolidated assessment.

---

### Blocking Issues by Category

| Category | Count | Severity | Action Required |
|----------|-------|----------|-----------------|
| **Security** | 1 | HIGH | TOCTOU race in walk.rs:147-166 |
| **Reliability** | 2 | HIGH | u32 cast truncation, --max-files=0 validation |
| **Architecture** | 2 | HIGH | Hand-rolled parser vs clap, mixed error types |
| **Rust** | 2 | HIGH | Unsafe cast, fragile Debug serialization |
| **Performance** | 2 | HIGH | Full memory residency, redundant syscalls |
| **Complexity** | 2 | HIGH | build_index (85L), walk_and_read (106L) |
| **Regression** | 1 | HIGH | Help delegation bug (mod.rs:34) |
| **Testing** | 2 | HIGH | Cache hit verification, minification skip test |
| **Consistency** | 1 | HIGH | Mixed error return types |

### Lower-Confidence Findings (60-79%)

These are **NOT required for merge** but represent quality improvements:

1. **Duplicate test helpers** (consistency, 90%) — manifest_tests.rs re-implements encode/decode instead of importing
2. **Misattributed I/O errors** (rust, 85%) — read_to_string errors classified as NonUtf8 instead of distinguishing
3. **SHA-256 encoding pattern** (consistency, 85%) — two divergent patterns for hex encoding
4. **Error handling UX** (consistency, 82%) — parse errors not user-friendly like heatmap pattern
5. **Test organization** (consistency, 80%) — new #[path] convention differs from existing inline-test pattern

### Comments Posted

**Highest Confidence (95%)**: Help delegation bug (mod.rs line 34)
**By Severity**: 13 inline comments covering all blocking issues with suggested fixes

---

### Path to Merge

1. **CRITICAL (95% confidence)**: Fix help delegation in `mod.rs` — prevents `skim search index --help` from working
2. **HIGH (90%+ confidence)**: Fix u32 truncation, TOCTOU race, --max-files=0 validation
3. **HIGH (85%+ confidence)**: Unify error types, migrate to clap, add test coverage
4. **MEDIUM (80-85% confidence)**: Fix fragile Debug serialization, silent error swallowing, complexity extraction

---

### Test Status

All 4006 workspace tests pass. New search module tests (38 tests) all pass. Recommendations add test coverage for:
- Incremental build cache hits (currently only asserts ExitCode)
- Minified file skip path (currently untested)
- Error handling paths (--max-files validation, parse errors)
- File deletion in incremental builds

---

### Deduplication Note

Many reviewers flagged the same issues independently, confirming critical paths:
- `as u32` cast flagged by Rust, Reliability reviewers
- --max-files=0 flagged by Security, Reliability reviewers
- Memory residency flagged by Performance, Reliability reviewers
- Error type mixing flagged by Architecture, Consistency reviewers
- Duplicate test helpers flagged by Consistency, Testing reviewers

Consolidated into single well-scoped comments above.

**Next**: Address HIGH-confidence blocking issues in order of confidence. Each comment includes working code samples.

dean0x:
## PR Review Summary

**Overall Recommendation**: APPROVED_WITH_CONDITIONS

### Review Scores
- Rust: 9/10 — APPROVED_WITH_CONDITIONS
- Reliability: 8/10 — APPROVED_WITH_CONDITIONS  
- Security: 9/10 — APPROVED
- Complexity: 8/10 — APPROVED_WITH_CONDITIONS
- Testing: 7/10 — APPROVED_WITH_CONDITIONS

### Key Blocking Issue: String-Matching Error Discrimination

**Location**: `crates/rskim/src/cmd/search/walk.rs:183-184`  
**Confidence**: 82-85% (flagged by Rust, Reliability, Complexity, and Testing reviewers)

The code uses `e.to_string().contains("too large")` to discriminate between error types, creating a fragile coupling to the exact error message string. This pattern has three problems:

1. **Silent breakage**: If the error message changes in a future refactor, the classification will silently degrade (misclassifying as `ReadError` instead of `TooLarge`).
2. **No test coverage**: The TOCTOU fallback path (file grew between pre-screen and open) is untested, so a message change could go undetected.
3. **Maintenance burden**: As error variants grow, this pattern becomes harder to maintain.

**Recommended Fix**: Replace string-matching with a typed error enum:

```rust
enum ReadError {
    TooLarge,
    NonUtf8,
    Io(io::Error),
}

fn open_and_read(path: &Path) -> Result<String, ReadError> { ... }
```

This makes the contract explicit and eliminates the string-matching heuristic entirely.

---

### Secondary Issues (Should Fix)

**`open_and_read` u64→usize cast lacks overflow guard** (`walk.rs:252`)  
**Confidence**: 82% (Reliability)

The cast `(size as usize).saturating_add(1)` is safe today (MAX_FILE_BYTES = 5 MB fits in 32-bit usize), but the pattern is fragile. Add a compile-time guard:

```rust
const _: () = assert!(MAX_FILE_BYTES <= usize::MAX as u64);
```

**Error handling nesting and complexity** (`walk.rs:179-196`)  
**Confidence**: 85% (Complexity)

The error-handling branch has 4 levels of nesting with the string-matching heuristic. Once the error type is refactored to a typed enum (see main issue), this nesting will collapse naturally.

---

These were not flagged as inline comments per the review guidelines, but should be considered for follow-up:

- **`#[must_use]` on `Language::as_str()`** (`types.rs:129`) — Confidence: 65%
- **`HashMap::with_capacity` hardcoded to 1024** (`manifest.rs:158`) — Confidence: 62%
- **Manifest entry count unbounded on parse** (`manifest.rs:158-169`) — Confidence: 65%
- **Cache directory hash collision risk** (`index.rs:295-305`) — Confidence: 62%

---

### What's Working Well

1. **TOCTOU fix**: The new `open_and_read()` function correctly uses the file handle for both metadata and read operations, eliminating the stat-then-read race.
2. **Bounded iteration**: `discover_project_root()` uses `MAX_ANCESTORS = 256` instead of an unbounded loop.
3. **Atomic writes**: Manifest persistence uses `NamedTempFile` + rename for crash safety.
4. **Error handling**: Consistent use of `Result` types with `.with_context()` for rich diagnostics.
5. **Type safety**: `u32::try_from()` guards prevent silent truncation.
6. **Symlink protection**: Walker is configured with `.follow_links(false)`.

---

### Merge Checklist

Before merging:
- [ ] Address string-matching error discrimination (primary blocker)
- [ ] Add compile-time guard for u64→usize cast
- [ ] Add test for TOCTOU fallback path in `open_and_read()`
- [ ] Consider `#[must_use]` on `Language::as_str()` (low priority)

---

**Generated by**: Claude Code Review Agent  
**Timestamp**: 2026-05-17

dean0x:
**[Consistency]** (95% confidence) — SKIM_DEBUG bypass

The index pipeline reads SKIM_DEBUG directly via std::env::var_os, bypassing the centralized `crate::debug::is_debug_enabled()` module used everywhere else in the codebase. This introduces a semantic difference: `is_debug_enabled()` checks truthiness (`1`/`true`/`yes`), while the index check only checks presence — setting `SKIM_DEBUG=false` would trigger debug output here but not elsewhere.

**Fix:** Replace the check on index.rs:196 with:
```rust
let debug_enabled = crate::debug::is_debug_enabled();
```

This is also a fast atomic load, so the stated optimization goal of hoisting the env-var check is already met.

---
*Review automation — consolidated from Consistency, Architecture, and Performance reviews*

dean0x:
**[Rust/Reliability]** (90% confidence) — Remove unsafe block in sha256_hex

The `String::from_utf8_unchecked` is correct and the safety invariant (NIBBLES contains only ASCII hex) is sound. However, the safe alternative `String::from_utf8(hex).expect("hex nibbles are always valid UTF-8")` provides identical performance on a 64-byte string called once per file. The performance difference is negligible (one bounds check on UTF-8 validation).

**Fix:** Replace walk.rs:332 with the safe version:
```rust
String::from_utf8(hex).expect("hex nibbles are always valid UTF-8")
```

This eliminates the `unsafe` block and its audit burden while maintaining identical behavior and performance.

---
*Review automation — flagged by Rust, Reliability, Security, and Consistency reviewers*

dean0x:
**[Security]** (82% confidence) — Validate manifest paths to prevent poisoning

The `FileManifest::load` deserializes `ManifestEntry.path` without validation. A crafted `index.skfiles` containing entries with `../` path segments or absolute paths would persist indefinitely when re-saved. While not exploitable in current code (lookups use keys computed from walked files), this violates defense-in-depth: future code iterating entries or using the path field to construct file paths would become a path traversal vector.

**Fix:** Add validation in FileManifest::load (manifest.rs:165):
```rust
if let Ok(entry) = serde_json::from_str::<ManifestEntry>(&line) {
    // Defense-in-depth: reject paths that could escape the project root
    if entry.path.contains("../") || entry.path.starts_with('/') {
        continue;
    }
    entries.insert(entry.path.clone(), entry);
}
```

---
*Review automation — flagged by Security reviewer*

dean0x:
**[Architecture]** (82% confidence) — Decouple build_index orchestration

The `build_index` function (93 lines, index.rs:150-243) handles six responsibilities sequentially: cache resolution, file walking, manifest loading, parallel classification, sequential index building, and persistence. While each step is delegated to helpers, the orchestration function itself couples concerns that could be tested independently.

For example, cache directory resolution is only tested through the full pipeline. As the pipeline grows (temporal scoring, query index, etc.), this function will accumulate more steps.

**Fix:** Extract a `Pipeline` struct holding config and cache_dir, with methods for each phase:
```rust
struct Pipeline { config: Config, cache_dir: PathBuf }
impl Pipeline {
    fn walk(&self) -> Result<Vec<ReadFile>> { ... }
    fn classify(&self, files: Vec<ReadFile>) -> Result<Vec<ClassifiedFile>> { ... }
    fn build(&self, classified: Vec<ClassifiedFile>) -> Result<IndexResult> { ... }
}
```

This makes individual phases unit-testable without running the full pipeline. Not blocking because current size is within tolerable range, but this is the right time to establish the seam before complexity grows.

---
*Review automation — flagged by Architecture and Complexity reviewers*

dean0x:
**[Architecture]** (85% confidence) — Replace std::mem::take with explicit ownership transfer

The `path_keys` vector is consumed via `std::mem::take(&mut path_keys[idx])` inside the sequential builder loop (index.rs:225), leaving empty strings at consumed indices. While this works because each index is visited exactly once, it creates a fragile coupling: the loop must process indices in order, and no future code may access `path_keys`. Any refactor that re-orders or parallelizes the builder would silently produce empty path keys.

**Fix:** Replace the take pattern with explicit ownership transfer:
```rust
for (idx, path_key) in path_keys.into_iter().enumerate() {
    // path_key is owned, no take needed
    new_manifest.insert(ManifestEntry { path: path_key, ... });
}
```

This makes the ownership transfer explicit and eliminates the subtle ordering invariant. The `take` pattern saved one allocation per file, but introduces a correctness hazard.

---
*Review automation — flagged by Architecture reviewer*

dean0x:
**[Performance]** (90% confidence) — Document memory ceiling for all-files-in-memory design

The pipeline reads every accepted file into a `Vec<ReadFile>` where each owns its full content. With a 50,000-file cap at 5MB each, theoretical peak RSS is ~250GB. On a typical mid-size repo (10,000 files, 20KB average), this is ~200MB held simultaneously. Content must be retained until the build loop completes.

**For v1:** This is acceptable given the file count and size caps, but should be documented as a known memory ceiling. Add a doc comment to `walk_and_read`:
```rust
/// Walks and reads files into memory.
///
/// # Memory Note
/// All file contents are held in memory simultaneously. With the default cap of 50,000 files
/// at up to 5 MB each, peak RSS could reach ~250 GB in theory. For typical repos (10,000
/// files at 20 KB avg), expect ~200 MB. Content is released after the build phase completes.
```

**For future iterations:** Consider a streaming two-pass approach: pass 1 walks + hashes (no content), pass 2 reads content on-demand during classify+build. Alternatively, use `memmap2` to let the OS page manager handle pressure.

---
*Review automation — flagged by Performance reviewer*

dean0x:
**[Performance]** (82% confidence) — Re-enable parallel walker traversal

The WalkBuilder::sort_by_file_path disables the ignore crate's parallel directory traversal. The I/O-heavy walk+read phase (file open, metadata, read, SHA-256) is entirely single-threaded and I/O-bound, while the classify phase is already parallelized with rayon.

If deterministic ordering is required for the manifest, sort after collection instead of disabling parallelization:

**Fix:** In walk.rs, remove the sort from the walker and sort the results after:
```rust
// builder.sort_by_file_path(...);  // REMOVE THIS

let mut files = walk_and_read_impl(&config.root, max_files)?;
files.0.sort_by(|a, b| a.rel_path.cmp(&b.rel_path));  // Sort after
```

This re-enables build_parallel for concurrent directory traversal and file reading. The downstream classify phase is already parallelized, so the walker can leverage parallel I/O.

---
*Review automation — flagged by Performance reviewer*

dean0x:
**[Performance]** (85% confidence) — Add mtime-based pre-screening for incremental builds

SHA-256 is computed for every file on every build, even in the incremental path. For large repos where most files are unchanged, this is redundant CPU work. A cheaper pre-screen using file mtime (already available from entry.metadata()) could skip hash computation for files whose mtime matches the manifest cache.

**Fix:** Add mtime fields to ManifestEntry and check them before re-hashing:
```rust
// In ManifestEntry:
pub mtime_secs: i64,
pub mtime_nanos: u32,

// In walk phase, skip read+hash when mtime matches:
if let Some(entry) = manifest.lookup(path_key)
    && entry.mtime_secs == meta_mtime_secs
    && entry.mtime_nanos == meta_mtime_nanos
{
    // Reuse cached hash and field_map — skip file read entirely
}
```

This converts the incremental no-op case from O(n * file_size) read + O(n * file_size) SHA-256 to O(n * stat), dramatically faster for unchanged repos.

---
*Review automation — flagged by Performance reviewer*

dean0x:
**[Complexity]** (92% confidence) — Split dns.rs into separate parser modules

The dns.rs file is 1034 lines, more than double the 500-line critical threshold. It contains two fully independent parser implementations (dig and nslookup), each with their own regex definitions, three-tier parse logic, helpers, and test suite — all in one file. The two parsers share only combine_stdout_stderr, so the coupling that justifies co-location is minimal.

**Fix:** Split into three files following the existing docker/ and gh/ patterns:
- dns/mod.rs — re-exports, shared regex utilities, combine_stdout_stderr
- dns/dig.rs — dig config, regex, parse chain, helpers
- dns/nslookup.rs — nslookup config, regex, parse chain, helpers
- dns/dig_tests.rs and dns/nslookup_tests.rs for tests

This mirrors the directory-module pattern already used in the infra module and makes each parser independently navigable.

---
*Review automation — flagged by Complexity reviewer*

dean0x:
**[Consistency]** (85% confidence) — Consolidate hex-encoding approaches

Two hex-encoding functions exist in the same module pair with different approaches:
- sha256_hex (walk.rs): nibble lookup table with unsafe from_utf8_unchecked
- project_root_hash (index.rs): write! macro with unwrap

The unsafe block in walk.rs is the only occurrence of from_utf8_unchecked in the entire codebase. While the safety invariant is correctly maintained, the inconsistency within a single PR is notable.

**Fix:** Either use the same approach in both functions (the write! approach in index.rs is consistent with the rest of the codebase's avoidance of unsafe), or extract a shared hex-encoding helper. If the nibble table is preferred for performance, add a brief comment justifying why the approach differs:
```rust
// sha256_hex uses a nibble table for hot-path performance
// project_root_hash uses write! for consistency with the rest of the codebase
```

Recommended: Replace sha256_hex unsafe with safe version (per Rust review comment) and use the same write! approach in both, or consolidate into a shared hex_encode(digest: &[u8]) -> String helper.

---
*Review automation — flagged by Consistency and Rust reviewers*

dean0x:
**[Reliability]** (85% confidence) — Add upper bound to manifest entry parsing

FileManifest::load reads every line from index.skfiles into a HashMap without any upper bound. While the file is locally generated (not untrusted external input), a corrupted or hand-edited manifest with millions of lines would cause unbounded memory growth during loading. Per the Iron Law: "every loop must have a fixed upper bound."

**Fix:** Add a manifest entry cap consistent with the index file cap:
```rust
const MAX_MANIFEST_ENTRIES: usize = 60_000;  // slightly above DEFAULT_MAX_FILES
let mut entries = HashMap::with_capacity(1024);
for line_result in lines {
    if entries.len() >= MAX_MANIFEST_ENTRIES {
        break;
    }
    // ... existing parsing logic ...
}
```

Also consider adding a file size gate (256 MiB sanity cap) to prevent BufReader::lines from reading unbounded-length single lines into memory.

---
*Review automation — flagged by Reliability reviewer*

dean0x:
**[Documentation]** (95% confidence) — Add CHANGELOG entry for index builder pipeline

The [Unreleased] section documents dig/nslookup (#168) and make (#167) but has no entry for this PR's core feature: the index builder pipeline (#182). The PR adds skim search index with walk/manifest/pipeline orchestration, incremental builds, and parallel classification — all undocumented in the CHANGELOG. This is a significant user-facing feature addition.

**Fix:** Add to CHANGELOG.md under [Unreleased] > ### Added:
```markdown
- **`skim search index` subcommand** -- Build or update the n-gram search index for the current project. Walk/classify/build pipeline with parallel tree-sitter classification (rayon), JSONL manifest sidecar for incremental builds (SHA-256 cache hits skip re-classification), atomic write ordering (.skpost -> .skidx -> .skfiles), minified file detection, and 50K file cap. `--force` flag for full rebuild, `--root` for explicit project root, `--max-files` override. (#182)
```

---
*Review automation — flagged by Documentation reviewer*

dean0x:
**[Documentation]** (90% confidence) — Clarify help text for unimplemented query path

The skim search help text lists query options (--lang, --ast, --json, --limit) and query examples as if they work, but skim search <QUERY> returns "not yet implemented". A user running any of these documented examples will get a confusing failure message. The help text presents these as working features when they are not.

**Fix:** Either remove unimplemented options/examples from help until implemented, or clearly mark them as upcoming. Recommended:

```rust
fn print_help() {
    println!("\
Usage: skim search <SUBCOMMAND> [OPTIONS]

Search code using layered n-gram indexing.

Subcommands:
  index    Build or update the search index for the current project

Options:
  -h, --help       Print this help message

Examples:
  skim search index              Build the search index
  skim search index --force      Rebuild from scratch

Query mode (skim search <QUERY>) is not yet implemented.");
}
```

This prevents user confusion and sets correct expectations.

---
*Review automation — flagged by Documentation reviewer*

dean0x:
## Code Review Summary

**13 inline comments created for high-confidence findings (≥80% confidence)**

### Overall Scores by Discipline

| Reviewer | Score | Recommendation |
|----------|-------|-----------------|
| Security | 9/10 | APPROVED_WITH_CONDITIONS |
| Architecture | 8/10 | APPROVED_WITH_CONDITIONS |
| Performance | 7/10 | APPROVED_WITH_CONDITIONS |
| Complexity | 7/10 | APPROVED_WITH_CONDITIONS |
| Consistency | 7/10 | CHANGES_REQUESTED |
| Regression | 9/10 | APPROVED |
| Testing | 7/10 | CHANGES_REQUESTED |
| Reliability | 7/10 | CHANGES_REQUESTED |
| Rust | 8/10 | APPROVED_WITH_CONDITIONS |
| Dependencies | 9/10 | APPROVED |
| Documentation | 6/10 | CHANGES_REQUESTED |

---

### Lower-Confidence Suggestions (60-79%) for Discussion

**Performance (70-75% confidence)**
- Manifest entry path cloned on insert (80% — consolidated into inline comment)
- decode_field_map allocates on every cache hit (80% — see Performance comment)
- Sorted manifest save may be unnecessary (65%) — consider skipping if not required for correctness
- to_string_lossy().replace allocates twice on Windows paths (70%) — use conditional replace

**Testing (65-80%)**
- Missing mtime-based incremental pre-screening coverage (would require perf feature implementation)
- Redundant tests inflate count without coverage — consider merging test_index_incremental_cache_hits_verified_via_manifest and test_index_incremental_manifest_correctness
- walk_and_read determinism test relies on implicit ordering assumption (82%) — add comment documenting sort guarantee
- No negative test for manifest format version mismatch (70%)
- Missing CapReached skip reason reporting test (65%)

**Architecture (65-70%)**
- FileManifest uses HashMap with cloned keys (65%) — minor allocation concern
- No trait abstraction for manifest store (70%) — aligns with integration test preference per CLAUDE.md
- ClassifiedFile bare tuple (62%) — small struct would improve readability

**Consistency (60-65%)**
- Language::name() vs Language::as_str() naming (70%) — distinction is documented, semantically sound
- SkipReason dead_code allow on entire enum (65%) — minor style preference
- Help text println vs eprintln (60%) — actually correct (help to stdout, errors to stderr)

**Documentation (65-80%)**
- Language::as_str not in CLAUDE.md API reference (80%) — add to "Adding New Language" section
- CLAUDE.md missing search subcommand (85%) — add to Analysis or new section
- README.md not updated with search subcommand (82%) — can defer to query-mode release

**Reliability (60-70%)**
- Manifest field_map ranges not validated (70%) — decode_field_map already filters unknowns
- project_root_hash truncation (65%) — 64 bits acceptable for local cache with documentation
- find_file_with_ext unbounded recursion in tests (62%) — low risk on tempdir

---

### Deduplication Summary

**Single findings flagged by multiple reviewers (consolidated into one inline comment each):**

1. **Unsafe in sha256_hex** — Flagged by Rust (90%), Reliability (85%), Security (82%), Architecture (80%)
2. **Manifest path validation** — Flagged by Security (82%), Architecture (concept), Consistency (concept)
3. **Performance all-files-in-memory** — Primary finding with documentation note
4. **Inconsistent hex encoding** — Flagged by Consistency (85%), Rust (75%), Performance (implicit)
5. **Manifest unbounded parsing** — Flagged by Reliability (85%), core design concern

**Total blocking findings across all reviewers:** 13 (all inline comments above)
**Total pre-existing (not blocking):** 1 (rskim-core version drift in dependencies review)

---

### Next Steps for Maintainer

1. Address the 13 inline comments (primary focus: unsafe removal, SKIM_DEBUG, path validation, manifest parsing bounds, manifest persistence fsync, documentation)
2. Consider (not blocking) the medium-confidence suggestions: mtime-based pre-screening, DNS file splitting, test consolidation
3. Update CHANGELOG, CLAUDE.md, README.md for search subcommand discovery
4. All 4059 tests currently pass; ensure test suite passes after addressing inline comments

---
*Code review automation — consolidated findings from 11 discipline reviewers (security, architecture, performance, complexity, consistency, regression, testing, reliability, rust, dependencies, documentation)*

dean0x:
## Code Review Summary: Index Builder Pipeline

**Report Date**: 2026-05-17  
**Files Changed**: 15  
**Total LOC Added**: 2,770

### Findings by Severity

#### BLOCKING: 7 items requiring fixes before merge

**Security & Reliability - Mutex Poisoning (85% confidence)**
- **Location**: `walk.rs` - parallel walker closure (lines 261, 272, 275, 287)
- **Issue**: Multiple `.lock().unwrap()` calls on `Mutex<Vec<>>`. If any thread panics, the mutex becomes poisoned and subsequent `.unwrap()` calls cascade into process termination.
- **Impact**: Production CLI can abort unexpectedly mid-run
- **Fix**: Use poison recovery: `.lock().unwrap_or_else(|e| e.into_inner())`

**Reliability - Arc::try_unwrap Panic (82% confidence)**
- **Location**: `walk.rs:300-307`
- **Issue**: `Arc::try_unwrap(...).expect(...)` will panic if walker threads don't complete cleanly (e.g., due to panic in thread pool). The expect message claims threads completed, but that invariant is never verified.
- **Fix**: Use error handling instead of panic:
  ```rust
  Arc::try_unwrap(files)
    .map_err(|_| anyhow::anyhow!("walker threads incomplete"))?
  ```

**Architecture - FileManifest Trait Abstraction (82% confidence)**
- **Location**: `manifest.rs:87-285`
- **Issue**: FileManifest couples persistence, in-memory state, and validation. Pipeline depends directly on concrete type, not trait abstraction, making it untestable without filesystem I/O. Violates DIP.
- **Suggestion**: Extract `ManifestStore` trait with `lookup/insert/save` methods. Enables in-memory mocks for testing.

**Architecture - build_index Orchestration (84% confidence)**
- **Location**: `index.rs:150-249`
- **Issue**: 100-line function mixes 6 responsibilities (cache resolution, directory creation, walking, manifest loading, parallel classification, sequential building, persistence). Multiple reasons to change = SRP violation.
- **Suggestion**: Extract phases into composable functions (resolve dirs, classify files, build+persist) when future phases are added.

**Architecture - Error Propagation Inconsistency (85% confidence)**
- **Location**: `index.rs:230`
- **Issue**: `add_file_classified` error via `?` aborts entire 50K-file build. Contradicts stated fail-soft philosophy in module docs.
- **Fix**: Skip failed files instead:
  ```rust
  if let Err(e) = builder.add_file_classified(...) {
    if debug_enabled { eprintln!(...); }
    continue;  // Skip, don't abort
  }
  ```

**Consistency - Missing Analytics Parameter (85% confidence)**
- **Location**: `index.rs:60` in `run()` signature
- **Issue**: Every other `run()` entry point accepts `&AnalyticsConfig`, but index::run drops this parameter. Breaks interface consistency and requires signature change when analytics recording is added.
- **Fix**: Add parameter even if unused: `pub(super) fn run(args: &[String], _analytics: &crate::analytics::AnalyticsConfig)`

**Complexity - Walker Closure Nesting (85% confidence)**
- **Location**: `walk.rs:252-298`
- **Issue**: 4-level nesting in parallel walker callback. Outer closure returns Box containing inner closure with nested matches. Dense and hard to follow.
- **Suggestion**: Extract body into named `handle_entry()` function. Parallel API requires Box shape, but body can delegate.

#### SHOULD-FIX: 4 items for quality/performance improvement

**Performance - sync_data Latency (82% confidence)**
- **Location**: `manifest.rs:277`
- **Issue**: `sync_data()` forces expensive fsync on every save (5-50ms per build). Manifest is a cache file, regenerable from source. Atomic rename already provides crash consistency.
- **Suggestion**: Consider removing `sync_data()` — the rename-based crash safety is sufficient for a developer tool cache.

**Testing - Missing Error Path Coverage (82% confidence)**
- **Location**: `walk_tests.rs:358`
- **Issue**: `test_walk_skips_minified_js_file` verifies file absence but doesn't assert the `SkipReason::Minified` is in the diagnostics list (unlike other skip tests). Silent regression if minification logic breaks.
- **Fix**: Add assertion: `assert!(skipped.iter().any(|r| matches!(r, SkipReason::Minified(_))))`

**Testing - SHA Determinism Lang Field (82% confidence)**
- **Location**: `walk_tests.rs:186`
- **Issue**: `test_walk_sha256_is_deterministic` only compares `rel_path` and `sha256`, not `lang` or `content`. Parallel walk refactoring is a major concurrency change — adding `lang` field to assertion would strengthen regression detection.
- **Fix**: Add `assert_eq!(f1.lang, f2.lang)` in the zip loop.

**Performance - Mutex Contention in Hot Path (85% confidence)**
- **Location**: `walk.rs:270-272`
- **Issue**: Every accepted file acquires the shared mutex. On large repos, this serializes the hot path and negates parallelism benefits.
- **Suggestion**: Accumulate results in thread-local storage, then merge after walk. Current approach is correct but sub-optimal. Channel-based approach could improve scaling.

#### SUGGESTIONS: Low-confidence items (60-79%) consolidated

- FileManifest trait extraction would improve testability (noted above)
- Consider `parking_lot::Mutex` for non-poisoning semantics
- `write!` to String at `index.rs:316` is safe but inconsistent with walk.rs pattern
- Minified file detection `is_minified()` lacks boundary-value unit tests
- Cap-checking with Relaxed ordering allows minor over-collection (handled by truncate)

### Test Coverage Status

- ✅ 2,359+ tests pass with 0 failures
- ✅ Incremental cache hits verified
- ✅ SHA-256 determinism validated (after parallel refactor)
- ✅ Skip conditions tested across non-UTF8, oversized, minified files
- ✅ No regressions detected from sequential→parallel walker conversion
- ⚠️ Missing: Direct test of walker I/O error propagation path
- ⚠️ Missing: Direct test of SkipReason variant consistency in minified path
- ⚠️ Missing: Boundary tests for `is_minified()` edge cases

### Overall Assessment

**Architecture Score**: 7/10 — Well-structured module boundaries, but FileManifest trait abstraction and orchestrator decomposition would improve testability.

**Reliability**: 6.5/10 — Strong bounds checking and safety limits, but mutex poisoning handling in parallel walker needs defensive improvement.

**Performance**: 7/10 — Parallel walker is solid improvement, but mutex contention and unnecessary fsync are addressable. Overall design appropriate for 50K file cap.

**Testing**: 7/10 — Good happy-path coverage and incremental build validation, but error paths and edge cases could be more comprehensive.

**Recommendation**: **APPROVED_WITH_CONDITIONS**

1. **Must fix before merge**: Mutex poisoning handling (lines 261,272,275,287 in walk.rs)
2. **Must fix before merge**: Arc::try_unwrap error handling (lines 300-307 in walk.rs)
3. **Should fix before merge**: Error propagation inconsistency in index.rs:230
4. **Should fix before merge**: Analytics parameter consistency in index.rs:60

Items 3-4 are softer but important for interface consistency. Quality-of-life improvements (trait abstraction, test coverage, performance) can be addressed post-merge if time-constrained.

---
**Claude Code Review Agent** | Processed 9 review reports | 47 findings consolidated

Issue #195, opened 2026-05-09
Research: BM25F parameter tuning via benchmarking


dean0x:
## Goal
After Wave 1 is functional, benchmark the 7-field BM25F config against baselines.

## Benchmarks
1. Standard BM25 (no fields) — baseline
2. Sourcegraph's 2-field approach (symbols vs content)
3. Our 7-field with default boosts
4. Tuned 7-field with optimized boosts

## Methodology
- Benchmark on 3+ real-world repos of different sizes
- Measure: precision@5, precision@10, MRR (mean reciprocal rank)
- Test queries: 50 representative search queries per repo

## Output
- Tuned parameters: k1, b per field, boost per field
- Benchmark report with comparison tables
- Published as comment on Research Archive #1

## Dependencies
- #10 (functional search system)

PR #247, opened 2026-05-20
feat(#195): rskim-bench crate — BM25F parameter tuning harness

dean0x:
## Summary

BM25F field boosts in rskim-search are intuition-based defaults with no empirical validation. This PR adds a new `rskim-bench` crate that provides a programmatic IR benchmark harness to measure and tune these parameters against real source code corpora.

- New `crates/rskim-bench` crate (`publish = false`) — zero changes to any existing crate
- Language-specific AST symbol extractors (Rust, Python, Go) generate ground-truth relevance judgments from the defining file of each symbol
- 4-config comparison (uniform, sourcegraph_style, default_8field, custom tuned) with MRR and Precision@K
- Coordinate descent tuner finds better BM25F parameters in ≤3 passes
- JSON + Markdown report output for CI integration

## Changes

**New crate: `crates/rskim-bench/`**
- `src/extract/` — tree-sitter AST symbol extractors for Rust, Python, Go
- `src/qrel.rs` — relevance judgment generation: extract → filter (DF≤5) → deduplicate → stratify
- `src/metrics.rs` — pure IR metrics: MRR, Precision@K, reciprocal rank
- `src/configs.rs` — 3 named BM25F configurations that all pass `validate()`
- `src/split.rs` — deterministic 60/40 train/test split via SHA-256 hash
- `src/harness.rs` — orchestrates index build → qrel gen → multi-config evaluation
- `src/tuning.rs` — coordinate descent: k1, per-field boosts, b values; converges ≤3 passes
- `src/report.rs` — JSON and Markdown output with per-repo + aggregate metrics
- `src/main.rs` — CLI: `bench`, `tune`, `qrels` subcommands
- `tests/integration.rs` — 12 integration tests covering all acceptance criteria

**Workspace change:**
- `Cargo.toml` — adds `crates/rskim-bench` to workspace members

## Breaking Changes

None. This is an additive new crate only.

## Reviewer Focus Areas

- **Qrel generation methodology** (`src/qrel.rs`): DF filter at ≤5, name length ≥4 bytes, first-occurrence deduplication, stratification targets
- **Language-specific extractors** (`src/extract/`): tree-sitter AST node kinds chosen for each language — are these the right node types?
- **Tuning convergence** (`src/tuning.rs`): coordinate descent sweep ranges and CONVERGENCE_THRESHOLD=0.001 — appropriate for the corpus size?
- **Train/test split** (`src/split.rs`): SHA-256 hash-based assignment guarantees determinism across runs
- **AC8 note**: default_8field > uniform on test MRR is verifiable once run against real corpora (sharkdp/fd, pallets/flask, gin-gonic/gin); the unit tests confirm the pipeline produces valid metrics


crates/rskim-bench/src/tuning.rs
```diff
@@ -0,0 +1,303 @@
...
+                    to_value: k1_candidate as f64,
+                    mrr_improvement: candidate_mrr - current_mrr,
+                });
+                current = candidate;
+                current_mrr = candidate_mrr;
+            }
+        }
+
+        // -- Sweep per-field boosts --
+        for field_idx in 0..8 {
```
dean0x: **[HIGH]** Hardcoded field count duplicates upstream constant

The literal `8` appears in loop bounds, array sizes, and field-name arrays. The upstream `rskim-search` defines `SearchField::count()` and `SearchField::ALL` as the single source of truth. If fields are added/removed, this code will silently produce incorrect results (truncated parameter arrays, skipped fields) rather than failing at compile time.

**Suggestion:** Use `SearchField::count()` or `SearchField::ALL.len()` for loop bounds. Define a local constant: `const N: usize = SearchField::count();` and use `[0.0f32; N]` for array sizes.

_Sources: Architecture, Complexity_

crates/rskim-bench/src/tuning.rs
```diff
@@ -0,0 +1,303 @@
...
+    }
+}
+
+/// Reconstruct a `BM25FConfig` from a `TuningResult`.
+///
+/// # Errors
+///
+/// Returns an error if the reconstructed config fails validation (shouldn't
+/// happen with well-formed tuning results).
+pub fn result_to_config(result: &TuningResult) -> anyhow::Result<BM25FConfig> {
```
dean0x, line 194: **[CRITICAL]** result_to_config silently produces wrong config when Vec lengths are < 8

`result_to_config` copies `best_field_boosts` and `best_field_b` using `.iter().take(8)`. If Vecs have fewer than 8 elements (from truncated JSON or future code changes), remaining array slots stay at `0.0`. Since 0.0 is valid per BM25F rules, validation passes silently. Result is a config that zeros out fields the caller never intended to zero, producing misleading benchmark numbers with no error or warning.

**Suggestion:** Assert the Vec lengths are exactly 8 before copying:
```rust
if result.best_field_boosts.len() != 8 || result.best_field_b.len() != 8 {
    bail!("TuningResult has wrong field count: boosts={}, b={} (expected 8 each)",
          result.best_field_boosts.len(), result.best_field_b.len());
}
```

_Sources: Reliability_

crates/rskim-bench/src/main.rs
```diff
@@ -0,0 +1,416 @@
...
+        }
+        _ => {
+            print!("{}", report::to_markdown(&bench_result, None));
+        }
+    }
+
+    Ok(())
+}
+
+fn run_tune(args: TuneArgs) -> anyhow::Result<()> {
```
dean0x: **[HIGH]** run_tune is a 134-line god function with 7+ distinct responsibilities

`run_tune` handles: corpus loading, file collection with ID assignment, qrel input construction, qrel generation, index building, train-split filtering, coordinate descent invocation with inline closure, tuning result conversion, final benchmark evaluation, and output formatting. This far exceeds the 50-line threshold and has cyclomatic complexity around 8.

**Suggestion:** Extract into named helpers:
- `load_all_repo_files()` (lines 224-248, also duplicated in run_bench/run_qrels)
- `build_index()` (lines 265-278)
- `run_tuning_evaluation()` (lines 289-306)
- `run_final_comparison()` (lines 314-338)
- `render_output()` (lines 340-350, also duplicated)

_Sources: Complexity_

crates/rskim-bench/src/main.rs
```diff
@@ -0,0 +1,416 @@
...
+        .iter()
+        .filter(|q| rskim_bench::split::assign_split(&q.query) == rskim_bench::split::Split::Train)
+        .cloned()
+        .collect();
+
+    eprintln!("Tuning on {} train qrels", train_qrels.len());
+
+    let idx_path = index_dir.path().to_path_buf();
+
+    let tuning_result = coordinate_descent(None, move |cfg: rskim_search::BM25FConfig| {
```
dean0x, line 397: **[CRITICAL]** Search errors silently swallowed in evaluate_split

`layer.search(&query).unwrap_or_default()` converts any search error (index corruption, I/O failure, deserialization error) into an empty result set. This means a broken index produces 0.0 MRR for every query, which looks identical to "no results found" rather than "the index is broken." For a benchmarking harness whose entire purpose is measuring search quality, this silently masks search failures producing misleading metrics.

**Suggestion:** Propagate the error or at minimum count and report error rates:
```rust
let results = layer.search(&query)
    .with_context(|| format!("searching for query '{}'", qrel.query))?;
```

_Sources: Reliability_

crates/rskim-bench/src/main.rs
```diff
@@ -0,0 +1,416 @@
...
+struct BenchArgs {
+    /// Path to the corpus directory (repos are cloned here if absent).
+    #[arg(long, default_value = ".bench-corpus")]
+    corpus_dir: PathBuf,
+
+    /// Path to corpus.toml (default: crates/rskim-research/corpus.toml).
+    #[arg(long)]
+    corpus_config: Option<PathBuf>,
+
+    /// Output format: json or markdown.
```
dean0x: **[HIGH]** Inconsistent format flag name across subcommands

`BenchArgs` and `TuneArgs` use `--output` (field name `output`) for the format flag, while `ReportArgs` uses `--format` (field name `format`) for the identical concept. Users must remember two different flag names for the same thing. Additionally, `ReportArgs` doc comment at line 99 references `bench --output json`, further highlighting the inconsistency.

**Suggestion:** Standardise on `--format` (matches the `skim stats --format json` convention used by main binary). Rename `output` to `format` in `BenchArgs` and `TuneArgs`.

_Sources: Consistency_

crates/rskim-bench/src/tuning.rs
```diff
@@ -0,0 +1,303 @@
...
+/// Run coordinate descent tuning.
+///
+/// # Arguments
+/// * `initial_config` — starting point (defaults to `BM25FConfig::default()` if None)
+/// * `evaluate` — closure that accepts a `BM25FConfig` and returns train-split MRR
+///
+/// # Returns
+///
+/// `TuningResult` with the best config found, convergence trace, and pass count.
+pub fn coordinate_descent<F>(initial_config: Option<BM25FConfig>, mut evaluate: F) -> TuningResult
```
dean0x, line 106: **[HIGH]** coordinate_descent is 109-line function with 4-level nesting and repetitive sweep blocks

The function contains three structurally identical "sweep" blocks (k1 sweep at lines 59-79, boost sweep at lines 82-107, b sweep at lines 111-136). Each follows the same pattern: iterate candidates, build config, validate, evaluate, compare, update if better. Nesting reaches 4 levels (pass loop > field loop > candidate loop > improvement check). Exceeds 50-line threshold by 2x.

**Suggestion:** Extract a generic `sweep_parameter` helper that takes candidates, a config-builder closure, and current best, returning the updated best. This would reduce three sweep blocks to three one-line calls.

_Sources: Complexity_

crates/rskim-bench/src/harness.rs
```diff
@@ -0,0 +1,389 @@
...
+    let mut rrs: Vec<f64> = Vec::with_capacity(qrels.len());
+    let mut p_at_5_sum = 0.0f64;
+    let mut p_at_10_sum = 0.0f64;
+    let mut found_at_rank_1 = 0usize;
+
+    for qrel in qrels {
+        let mut query = SearchQuery::new(&qrel.query);
+        query.limit = Some(TOP_K);
+
+        let results = layer.search(&query).unwrap_or_default();
```
dean0x: **[HIGH]** Index re-opened from disk on every tuning evaluation

The `coordinate_descent` evaluator closure calls `NgramIndexReader::open_with_config(&idx_path, cfg)` for every candidate configuration. Coordinate descent evaluates: 6 (k1) + 8*9 (boosts) + 2*5 (b) = 88 evaluations per pass, up to 3 passes = ~264 index opens. Each `open_with_config` re-reads index files from disk and reconstructs in-memory structures. Since the index data never changes (only scoring parameters), this is redundant I/O and the single biggest performance bottleneck in the tuning workflow.

**Suggestion:** Add a method like `with_config(cfg)` that swaps scoring parameters on an already-loaded reader without re-reading from disk. This is a **blocking** performance issue for larger corpora.

_Sources: Performance_

dean0x (commented):
## Code Review Summary

Comprehensive review completed across 10 focus areas. **26 issues identified** with 80%+ confidence, spanning architecture, reliability, testing, and best practices.

**Critical blocking items:**
- Hardcoded field count duplication (85% confidence)
- Silent config corruption in result_to_config (92% confidence)  
- Search errors swallowed in harness evaluation (85% confidence)
- Tautological integration test (95% confidence)

**High-impact should-fix items:**
- run_tune god function at 134 lines (92% confidence)
- coordinate_descent at 109 lines with 4-level nesting (88% confidence)
- Index re-opened per evaluation in tuning loop (88% confidence)
- Inconsistent format flag naming across subcommands (95% confidence)

Detailed inline comments posted for each finding. See individual comment threads for specific line references and fix suggestions.

crates/rskim-bench/src/types.rs
```diff
@@ -0,0 +1,113 @@
...
+    /// MRR improvement gained.
+    pub mrr_improvement: f64,
+}
+
+/// Result of the coordinate descent tuning run.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct TuningResult {
+    /// Best BM25F configuration found.
+    pub best_k1: f32,
+    pub best_field_boosts: Vec<f32>,
```
dean0x: **[MEDIUM]** TuningResult uses Vec for fixed-size-8 field arrays

`best_field_boosts` and `best_field_b` are `Vec<f32>` but always contain exactly 8 elements (one per field). This forces `result_to_config()` in `tuning.rs:165-172` to perform fallible conversion with manual index-by-index copying and `.take(8)` guards. The report renderer in `report.rs:93-94` uses `.get(i).copied().unwrap_or(0.0)`. The Vec type communicates "variable length" when the domain invariant is "exactly FIELD_COUNT elements."

**Suggestion:** Change to `[f32; 8]`. This removes the fallible conversion in `result_to_config()`, eliminates the defensive `.get()` calls in the report renderer, and makes the invariant compile-time enforced.

_Sources: Architecture, Rust, Reliability_

crates/rskim-bench/src/harness.rs
```diff
@@ -0,0 +1,389 @@
...
+    BM25FConfig, FileId, LayerBuilder, NgramIndexBuilder, SearchLayer, SearchQuery,
+};
+
+use crate::metrics::{mrr, precision_at_k, rank_of, reciprocal_rank};
+use crate::qrel::{QrelInput, generate_qrels, validate_qrel_coverage};
+use crate::split::partition;
+use crate::types::{BenchResult, ConfigMetrics, IndexedFile, Qrel, RepoBenchResult};
+
+/// Configuration for a single benchmark run.
+pub struct BenchConfig {
```
dean0x, line 18: **[MEDIUM]** BenchConfig missing #[derive(Debug)]

Every other public struct in this crate derives `Debug` (`Qrel`, `EvalResult`, `ConfigMetrics`, `RepoBenchResult`, `BenchResult`, `TuningResult`, `IndexedFile`, `ExtractedSymbol`, `QrelInput`). `BenchConfig` is the sole exception. This breaks the workspace pattern where all public types implement `Debug` (enforced in rskim-core, rskim-search, rskim-research). It also makes debugging harder -- `BenchConfig` values cannot be `dbg!()` printed.

**Suggestion:** Add `#[derive(Debug)]` to `BenchConfig`.

_Sources: Consistency, Rust_

crates/rskim-bench/src/main.rs
```diff
@@ -0,0 +1,416 @@
...
+    PathBuf::from(env!("CARGO_MANIFEST_DIR"))
+        .parent()
+        .map(|p| p.join("rskim-research").join("corpus.toml"))
+        .unwrap_or_else(|| PathBuf::from("corpus.toml"))
+}
+
+/// Load corpus config and prepare the GitCloneSource, creating the corpus dir if absent.
+fn open_corpus(
+    corpus_config: Option<PathBuf>,
+    corpus_dir: &PathBuf,
```
dean0x: **[MEDIUM]** Function parameter should accept &Path, not &PathBuf (C-BORROW)

`open_corpus` accepts `corpus_dir: &PathBuf` instead of `&Path`. Per Rust API guidelines, functions should accept the borrowed form of a type (`&Path`) rather than a reference to the owned form (`&PathBuf`). While `&PathBuf` auto-derefs to `&Path` at call sites (not a functional bug), this is non-idiomatic and constrains callers unnecessarily.

**Suggestion:** Change signature to `corpus_dir: &std::path::Path`. The `.clone()` on line 136 would then use `corpus_dir.to_path_buf()` instead.

_Sources: Rust_

crates/rskim-bench/src/report.rs
```diff
@@ -0,0 +1,248 @@
...
+/// Serialise a `BenchResult` (and optional tuning result) to a JSON string.
+///
+/// # Errors
+///
+/// Returns `serde_json::Error` if serialisation fails (should not happen in
+/// practice with the types used here).
+pub fn to_json(
+    result: &BenchResult,
+    tuning: Option<&TuningResult>,
+) -> Result<String, serde_json::Error> {
```
dean0x: **[MEDIUM]** to_json returns wrong error type instead of anyhow::Result

Every other fallible public function in this crate returns `anyhow::Result<T>` (`generate_qrels`, `validate_qrel_coverage`, `run_on_files`, `evaluate_split`, `result_to_config`). `to_json` is the only function returning a raw library error type (`serde_json::Error`). While callers in `main.rs` add `?` inside an `anyhow::Result` context so it auto-converts, this forces callers outside that context to handle a different error type than the rest of the crate's API.

**Suggestion:** Change to `anyhow::Result<String>` for consistency. The `?` operator on `serde_json` calls will auto-convert via `anyhow::Error::from`.

_Sources: Consistency_

crates/rskim-bench/src/types.rs
```diff
@@ -0,0 +1,113 @@
...
+    /// The symbol name used as the query text.
+    pub query: String,
+    /// The `FileId` of the file that defines this symbol (ground truth).
+    pub relevant_file_id: FileId,
+    /// The field type that produced this symbol.
+    pub field: SearchField,
+}
+
+/// Per-query evaluation result for a single configuration.
+#[derive(Debug, Clone, Serialize, Deserialize)]
```
dean0x: **[MEDIUM]** Dead public type EvalResult has no callers

`EvalResult` is defined as a public struct with `Serialize`/`Deserialize` derives but is never referenced anywhere in the crate (not in harness, report, main, tests, or integration tests). The workspace convention is to delete dead code rather than keep it around -- CLAUDE.md explicitly states "Delete dead code -- commented-out code is not version control."

**Suggestion:** Remove `EvalResult` from `types.rs`. If it is intended for future use, leave a `// TODO:` comment documenting the planned use case.

_Sources: Consistency, Testing_

crates/rskim-bench/Cargo.toml
```diff
@@ -0,0 +1,41 @@
...
+tree-sitter = { workspace = true }
+tree-sitter-rust = { workspace = true }
+tree-sitter-python = { workspace = true }
+tree-sitter-go = { workspace = true }
+anyhow = { workspace = true }
+clap = { workspace = true }
+serde = { workspace = true }
+serde_json = { workspace = true }
+sha2 = { workspace = true }
+tempfile = { workspace = true }
```
dean0x, line 33: **[MEDIUM]** Redundant tempfile in [dev-dependencies]

`tempfile` is listed in both `[dependencies]` (line 32) and `[dev-dependencies]` (line 35). Since production code in `main.rs` uses `tempfile::tempdir()` for benchmark index directories, the `[dependencies]` entry is correct and necessary. The `[dev-dependencies]` entry is completely redundant -- Cargo resolves `tempfile` from `[dependencies]` for both production and test compilation. This adds noise and could cause confusion.

**Suggestion:** Remove the entire `[dev-dependencies]` section (since `tempfile` is its only entry).

_Sources: Dependencies, Rust_

crates/rskim-bench/src/extract/rust_lang.rs
```diff
@@ -0,0 +1,269 @@
...
+//! - `impl_item`     → type name (if simple identifier) → SymbolName
+
+use std::path::Path;
+
+use rskim_search::SearchField;
+
+use super::ExtractedSymbol;
+
+/// Extract named symbols from Rust source using tree-sitter.
+pub fn extract(path: &Path, content: &str) -> Vec<ExtractedSymbol> {
```
dean0x, line 19: **[MEDIUM]** Tree-sitter Parser allocated per file in extractors

Each call to `extract()` creates a new `tree_sitter::Parser`, sets its language, parses the content, then drops the parser. When processing a corpus with hundreds or thousands of files, this means hundreds of parser allocations and language-setting operations. Parser creation is not free. Since the extractors are always called in a loop over files (in `generate_qrels`), the parser could be created once and reused.

**Suggestion:** Accept a `&mut Parser` parameter in each `extract()` function. Have `extract_symbols()` in `mod.rs` maintain one parser per language. Or use a thread-local or pass a pre-configured parser from the caller.

_Sources: Performance_

crates/rskim-bench/src/main.rs
```diff
@@ -0,0 +1,416 @@
...
+    #[arg(long, default_value = ".bench-corpus")]
+    corpus_dir: PathBuf,
+
+    /// Path to corpus.toml (default: crates/rskim-research/corpus.toml).
+    #[arg(long)]
+    corpus_config: Option<PathBuf>,
+
+    /// Output format: json or markdown.
+    #[arg(long, default_value = "markdown")]
+    output: String,
```
dean0x: **[MEDIUM]** Output format fields use String instead of ValueEnum

The `output` field on `BenchArgs`/`TuneArgs` and `format` field on `ReportArgs` accept arbitrary `String` values. Invalid inputs like `"xml"` or `"csv"` silently fall through to the markdown default via the `_ =>` match arm. This violates the principle of making illegal states unrepresentable. Clap's `ValueEnum` derive provides compile-time exhaustive matching and proper error messages for free.

**Suggestion:** Define an `OutputFormat` enum with `#[derive(ValueEnum)]`:
```rust
#[derive(Debug, Clone, clap::ValueEnum)]
enum OutputFormat { Json, Markdown }
```
Then use `#[arg(long, default_value_t = OutputFormat::Markdown)] output: OutputFormat`.

_Sources: Rust_

crates/rskim-bench/src/main.rs
```diff
@@ -0,0 +1,416 @@
...
+    Ok(())
+}
+
+fn run_tune(args: TuneArgs) -> anyhow::Result<()> {
+    let (corpus, source) = open_corpus(args.corpus_config, &args.corpus_dir)?;
+
+    // Load all files from all repos for tuning
+    let mut all_indexed: Vec<IndexedFile> = Vec::new();
+    let mut all_contents: HashMap<FileId, String> = HashMap::new();
+    let mut file_id_counter = 0u32;
```
dean0x: **[MEDIUM]** file_id_counter overflow unchecked

`file_id_counter` is a `u32` incremented once per file across all repos. If the corpus contains more than `u32::MAX` (~4.3 billion) files, this wraps silently. While practically unlikely today, the counter also lacks an overflow assertion which would catch misconfiguration or bugs in file enumeration. The `i as u32` cast in `run_bench` (line 183) and `run_qrels` (line 399) would panic on debug builds but silently truncate on release builds if a single repo has > 4B files.

**Suggestion:** Use checked arithmetic:
```rust
let fid = FileId(file_id_counter.checked_add(1)
    .context("file_id_counter overflow: too many files")?);
file_id_counter += 1;
```
For the `i as u32` casts, use `u32::try_from(i).context("too many files")?`.

_Sources: Reliability, Rust_

crates/rskim-bench/src/harness.rs
```diff
@@ -0,0 +1,389 @@
...
+    }
+}
+
+fn macro_average<F>(
+    repos: &[RepoBenchResult],
+    config_names: &[String],
+    get_metrics: F,
+) -> Vec<ConfigMetrics>
+where
+    F: Fn(&RepoBenchResult) -> &Vec<ConfigMetrics>,
```
dean0x: **[MEDIUM]** Return type should be &[ConfigMetrics], not &Vec<ConfigMetrics> (C-BORROW)

The `macro_average` function's trait bound specifies `F: Fn(&RepoBenchResult) -> &Vec<ConfigMetrics>`. Returning `&Vec<T>` instead of `&[T]` is non-idiomatic Rust -- it exposes the concrete container type when only a slice is needed. This limits the trait bound unnecessarily (callers must return a reference to a `Vec`, not any slice-like container).

**Suggestion:** Change to `F: Fn(&RepoBenchResult) -> &[ConfigMetrics]`.

_Sources: Rust_

crates/rskim-bench/src/main.rs
```diff
@@ -0,0 +1,416 @@
...
+        },
+        BenchConfig {
+            name: "default_8field".to_string(),
+            bm25f: configs::default_8field(),
+        },
+    ];
+
+    let mut repo_results = Vec::new();
+
+    for repo_entry in &corpus.repos {
```
dean0x: **[MEDIUM]** Duplicate file-loading and indexing logic across 3 subcommands

Three CLI handlers each repeat: open corpus, iterate repos, sort files by path, assign FileIds, build HashMap of contents, construct QrelInput/IndexedFile vectors. The sort-then-enumerate-then-map pattern appears in three places with minor variations (`run_bench` creates per-repo IDs starting at 0, `run_tune` uses global counter, `run_qrels` starts at 0 per repo). This is a classic SRP violation -- the "load and prepare files from corpus" concern is duplicated.

**Suggestion:** Extract a shared helper:
```rust
fn load_repo_files(source: &dyn FileSource, repo: &RepoEntry) 
    -> Result<(Vec<IndexedFile>, HashMap<FileId, String>)>
```
The global-counter variant for `run_tune` could take a `starting_id: u32` parameter.

_Sources: Architecture, Complexity_

dean0x: **[MEDIUM]** Sequential repo processing can be parallelized

The `run_bench` function iterates over repos sequentially (`for repo_entry in &corpus.repos`). Each repo involves git clone (or cache hit), file reading, indexing, and evaluation. These per-repo operations are independent and could be parallelized with rayon, which is already a dependency of the parent workspace. Similarly, `run_tune` loads all repos sequentially.

**Suggestion:** Use `rayon::prelude::*` and `.par_iter()` for the repo loop in `run_bench`. Each repo gets its own tempdir and index, so there are no shared-state conflicts. This would give near-linear speedup on multi-core machines.

_Sources: Performance_

crates/rskim-bench/src/extract/rust_lang.rs
dean0x, line 19: **[MEDIUM]** Extractor boilerplate duplicated across 3 modules

All three extractors (go.rs, python.rs, rust_lang.rs) share an identical `extract()` function body (19 lines): create parser, set language, parse, get bytes/root/cursor, call `walk_node`, return symbols. They also share identical `walk_node` recursion boilerplate (~10 lines each). That is approximately 30 lines duplicated verbatim across 3 files (90 lines total identical code). When a fourth language extractor is added, this pattern will be copied again.

**Suggestion:** Extract a shared `parse_and_walk` function in `extract/mod.rs` that takes the tree-sitter `Language` and a node-visitor closure. Each language module would supply only the language constant and the match arms for its specific node kinds.

_Sources: Complexity_

crates/rskim-bench/tests/integration.rs
```diff
@@ -0,0 +1,376 @@
...
+            );
+        }
+    }
+}
+
+// ============================================================================
+// AC24: FileId assignment is deterministic (files sorted by path)
+// ============================================================================
+
+#[test]
```
dean0x, line 207: **[BLOCKING]** Tautological test: file_id_assignment_deterministic_when_sorted

This test runs the exact same deterministic `enumerate().map()` expression twice on the same input and asserts the two results are equal. It is a tautology -- it can never fail regardless of how `FileId` assignment changes in production code. It does not exercise any production function; it only tests that `Vec::iter().enumerate()` is deterministic, which is a property of the standard library. If someone reorders the sort or changes the assignment scheme, this test will still pass.

**Suggestion:** Replace with a test that invokes the actual production path (`run_on_files`) twice with the same files in different initial orderings and asserts the resulting `qrel_count`, metrics, or qrel file IDs are identical across both runs.

_Sources: Testing_

crates/rskim-bench/src/extract/go.rs
```diff
@@ -0,0 +1,275 @@
...
+//! - `import_spec`           → last path segment → ImportExport
+
+use std::path::Path;
+
+use rskim_search::SearchField;
+
+use super::ExtractedSymbol;
+
+/// Extract named symbols from Go source using tree-sitter.
+pub fn extract(path: &Path, content: &str) -> Vec<ExtractedSymbol> {
```
dean0x, line 17: **[MEDIUM]** No integration test for extract_symbols dispatch

The `extract_symbols` dispatch function in `extract/mod.rs` is tested indirectly (through `generate_qrels` which only exercises Rust content). There are no tests that call `extract_symbols` directly with Python or Go content and verify the dispatch routes correctly. A bug in the `match` dispatch (e.g., routing Go to the Python extractor) would not be caught. Additionally, the fallback for unsupported languages (returning empty `Vec`) is untested.

**Suggestion:** Add a small integration test that calls `extract_symbols` with known content for each supported language (Rust, Python, Go) plus one unsupported language, asserting non-empty results for supported and empty for unsupported.

_Sources: Testing_

crates/rskim-bench/src/harness.rs
```diff
@@ -0,0 +1,389 @@
...
+/// # Arguments
+/// * `files` — files with assigned IDs, sorted by path for determinism
+/// * `configs` — named BM25F configurations to compare
+/// * `index_dir` — writable directory for the index files
+///
+/// # Errors
+///
+/// Returns an error if indexing fails or no qrels can be generated.
+pub fn run_on_files(
+    files: &[IndexedFile],
```
dean0x, line 39: **[MEDIUM]** Missing error-path test for run_on_files

`run_on_files` has multiple failure points (index builder creation, per-file indexing, qrel generation, qrel coverage validation, index opening). Only the happy path is tested. There are no tests verifying that meaningful errors propagate when, for example, no qrels can be generated from the input files (insufficient symbols), or when `contents` is missing entries for indexed files. The `qrel.rs` unit tests cover the "too few qrels" error, but the harness-level error propagation through `anyhow::Context` is never exercised.

**Suggestion:** Add at least one test that passes insufficient content to `run_on_files` (e.g., a single file with `"fn x() {}"``) and asserts the error is `Err` with a descriptive message containing "generating qrels".

_Sources: Testing_

crates/rskim-bench/tests/integration.rs
```diff
@@ -0,0 +1,376 @@
...
+    );
+    assert!(md.contains("repo"), "markdown should reference repo name");
+}
+
+// ============================================================================
+// Full pipeline smoke test: index → qrel → eval
+// ============================================================================
+
+#[test]
+fn full_pipeline_produces_non_zero_metrics() {
```
dean0x, line 306: **[MEDIUM]** Weak assertion in full_pipeline_produces_non_zero_metrics

The assertion `any_nonzero_mrr` uses an OR condition (`m.mrr > 0.0 || m.query_count > 0`) that will pass even if MRR is 0 for all configs as long as any config has `query_count > 0`. Since `query_count` will almost always be non-zero (there are enough symbols in the synthetic data), this test cannot detect a regression where search stops returning relevant results. The test name promises "non-zero metrics" but only verifies query presence.

**Suggestion:** Split into two assertions: (1) assert `query_count > 0` for at least one split, and (2) assert `mrr > 0.0` for at least one config+split combination. This ensures the search engine actually finds relevant documents.

_Sources: Testing_

crates/rskim-bench/Cargo.toml
```diff
@@ -0,0 +1,41 @@
...
+[[bin]]
+name = "rskim-bench"
+path = "src/main.rs"
+
+[lib]
+name = "rskim_bench"
+path = "src/lib.rs"
+
+[dependencies]
+rskim-search = { path = "../rskim-search" }
```
dean0x, line 20: **[MEDIUM]** Internal path dependencies missing version field

The three internal path dependencies (`rskim-search`, `rskim-core`, `rskim-research`) use path-only references without a `version` field. While the crate is `publish = false` so this has no functional impact, other workspace crates follow a convention of including the version alongside the path. For example, `crates/rskim/Cargo.toml` uses `rskim-core = { version = "2.10.0", path = "../rskim-core" }`. Matching the convention makes dependencies self-documenting and prevents surprises if `publish = false` is ever removed.

**Suggestion:** Add version fields to match workspace convention:
```toml
rskim-search = { version = "0.1.0", path = "../rskim-search" }
rskim-core = { version = "2.10.0", path = "../rskim-core" }
rskim-research = { version = "0.1.0", path = "../rskim-research" }
```

_Sources: Dependencies_

crates/rskim-bench/src/tuning.rs
```diff
@@ -0,0 +1,319 @@
...
+/// * `current_mrr` — MRR of current best (mutated in-place on improvement)
+/// * `history` — convergence trace (extended on improvement)
+/// * `pass` — current pass index for trace labelling
+/// * `param_name` — display name for the parameter being swept
+/// * `candidates` — candidate values to try
+/// * `get_value` — extracts the current parameter value from a config
+/// * `make_candidate` — produces a new config with the candidate value applied
+/// * `evaluate` — closure that scores a config and returns MRR
+#[allow(clippy::too_many_arguments)]
+fn sweep_parameter<G>(
```
dean0x, line 64: **[HIGH: Cognitive Complexity]** `sweep_parameter` has 9 parameters

The function accepts 9 parameters, exceeding the 5-parameter cognitive threshold. While the extraction correctly eliminates duplicated sweep logic (a net improvement), the parameter count creates overhead.

**Fix:** Group mutable state into a `SweepState` struct, reducing parameters from 9 to 6.

_Confidence: 85% · Architecture, Complexity, Rust_

crates/rskim-bench/src/tuning.rs
```diff
@@ -0,0 +1,319 @@
...
+    pass: usize,
+    param_name: &str,
+    candidates: &[f32],
+    get_value: impl Fn(&BM25FConfig) -> f32,
+    make_candidate: impl Fn(&BM25FConfig, f32) -> BM25FConfig,
+    evaluate: &mut G,
+) where
+    G: FnMut(BM25FConfig) -> f64,
+{
+    let from_value = get_value(current) as f64;
```
dean0x: **[HIGH: Data Accuracy]** `from_value` captures stale value across multi-step sweeps

`from_value` is captured once at function entry. If a candidate improves and mutates `current`, subsequent candidates in the same sweep record the original `from_value` rather than the intermediate value. This means the convergence trace shows inaccurate pre-improvement values for the 2nd+ improvements within one sweep call.

**Fix:** Move `from_value` capture inside the improvement branch:
```rust
if candidate_mrr > *current_mrr {
    let from_value = get_value(current) as f64; // capture just before mutation
    history.push(...);
    *current = candidate;
}
```

_Confidence: 82% · Rust_

crates/rskim-bench/src/extract/mod.rs
```diff
@@ -0,0 +1,128 @@
...
+{
+    let mut parser = tree_sitter::Parser::new();
+    if parser.set_language(&ts_language).is_err() {
+        return vec![];
+    }
+    walk_ast_with_parser(&mut parser, content, visit)
+}
+
+/// Recursive node visitor used by `walk_ast`.
+fn walk_nodes<F>(
```
dean0x, line 95: **[HIGH: Reliability]** Recursive `walk_nodes` has no depth bound

The function recurses on every child node with no explicit depth limit. While tree-sitter ASTs are typically bounded by source nesting, pathological or maliciously crafted files (deeply nested expressions, macro-generated code) could overflow the stack. The corpus is fetched from external repos, so input is not fully trusted.

**Fix:** Add a depth parameter with upper bound (e.g., 256):
```rust
fn walk_nodes<F>(
    node: tree_sitter::Node<'_>,
    cursor: &mut tree_sitter::TreeCursor<'_>,
    bytes: &[u8],
    symbols: &mut Vec<ExtractedSymbol>,
    visit: &mut F,
    depth: usize,
) where
    F: FnMut(...),
{
    const MAX_DEPTH: usize = 256;
    if depth >= MAX_DEPTH { return; }
    // ... rest of function ...
}
```
Update call sites: `walk_nodes(..., depth + 1)`

_Confidence: 85% · Reliability, Rust_

crates/rskim-bench/src/report.rs
```diff
@@ -0,0 +1,262 @@
...
+        if let serde_json::Value::Object(ref mut map) = obj {
+            map.insert("tuning".to_string(), tuning_val);
+        }
+    }
+
+    Ok(serde_json::to_string_pretty(&obj)?)
+}
+
+/// Map a `SearchField` to its PascalCase display name.
+fn field_display_name(field: SearchField) -> &'static str {
```
dean0x: **[HIGH: Architecture]** `field_display_name` duplicates `SearchField` variant-to-string mapping

This function manually maps all `SearchField` variants to PascalCase strings. `SearchField` already has a `name()` method (returning snake_case like "type_definition"). Adding a second exhaustive match creates tight coupling to the upstream enum's representation without leveraging its existing API. When a `SearchField` variant is added, both locations must be updated.

**Fix:** Either:
- (a) Add a `display_name()` method to `SearchField` in `rskim-search` returning PascalCase (single source of truth)
- (b) Derive PascalCase programmatically from `SearchField::name()` via a small utility function

Option (a) is preferred for compile-time enforcement.

_Confidence: 85% · Architecture_

crates/rskim-bench/src/main.rs
```diff
@@ -0,0 +1,520 @@
...
+    let train_qrels = rskim_bench::qrel::generate_qrels(&qrel_inputs)
+        .context("generating qrels")?
+        .into_iter()
+        .filter(|q| rskim_bench::split::assign_split(&q.query) == rskim_bench::split::Split::Train)
+        .collect();
+
+    Ok(train_qrels)
+}
+
+fn run_tune(args: TuneArgs) -> anyhow::Result<()> {
```
dean0x, line 339: **[HIGH: Complexity]** `run_tune` is 120-line orchestration function with 7+ distinct responsibilities

Handles: parallel repo loading, sequential FileId reassignment, index building, qrel generation, error tracking, coordinate descent invocation with inline closure, tuning result conversion, final evaluation, and output formatting. This far exceeds the 50-line guideline and mixes high-level orchestration with implementation details.

**Fix:** Extract into named helpers:
- Extract `load_all_repos` block (lines 342-372): parallel load + FileId reassignment
- Extract final evaluation + formatting block (lines 421-456) into `run_final_eval_and_report`

This brings `run_tune` down to ~40 lines of high-level orchestration.

_Confidence: 80-82% · Architecture, Complexity_

crates/rskim-bench/src/main.rs
```diff
@@ -0,0 +1,520 @@
...
+        })
+        .collect::<anyhow::Result<Vec<_>>>()?;
+
+    // Reassign globally unique IDs sequentially so the combined corpus has
+    // no duplicate FileIds. (IDs were local-to-repo during parallel load.)
+    let mut all_indexed: Vec<IndexedFile> = Vec::new();
+    let mut all_contents: HashMap<FileId, String> = HashMap::new();
+    let mut global_id = 0u32;
+
+    for mut lr in loaded_repos {
```
dean0x, line 360: **[HIGH: Reliability]** ID reassignment silently drops files when `contents.remove` returns `None`

In the ID reassignment loop, if `lr.contents.remove(&old_id)` returns `None`, the `IndexedFile` is still pushed to `all_indexed` but with no corresponding content entry. Downstream code (`build_index`, `make_train_qrels`) treats missing content as `""` via `.unwrap_or("")`, resulting in files indexed with empty content — a silent data integrity violation that skews benchmark results.

**Fix:** Assert that removal always succeeds (both vecs are built from the same source):
```rust
if let Some(content) = lr.contents.remove(&old_id) {
    all_contents.insert(FileId(global_id), content);
} else {
    debug_assert!(false, "FileId {old_id:?} missing content — data consistency violation");
}
```

_Confidence: 83% · Reliability_

crates/rskim-bench/src/main.rs
```diff
@@ -0,0 +1,520 @@
...
+
+    let train_qrels = make_train_qrels(&all_indexed, &all_contents)?;
+    eprintln!("Tuning on {} train qrels", train_qrels.len());
+
+    // Open reader once; BM25F config is overridden per-evaluation via
+    // SearchQuery::bm25f_config (single-reader pattern, item 8).
+    let reader = rskim_search::NgramIndexReader::open(index_dir.path())
+        .context("opening index for tuning")?;
+
+    // Error counter shared across closure invocations. coordinate_descent requires an f64
```
dean0x, line 391: **[HIGH: Reliability]** Tuning error counter silently saturates with no hard failure

The evaluation error counter caps logging at 5 messages. If every evaluation fails (e.g., corrupt index), `coordinate_descent` will run ~90 evaluations (3 passes × 30+ per pass) with all returning 0.0 MRR. Only a warning is logged; no error is returned. The tuning result is meaningless but the function continues to the final evaluation step with a bogus config.

**Fix:** After tuning, fail hard if all evaluations errored:
```rust
let total_errors = eval_error_count.load(Ordering::Relaxed);
if tuning_result.best_train_mrr == 0.0 && total_errors > 0 {
    anyhow::bail!(
        "Tuning produced 0.0 MRR with {total_errors} evaluation error(s) — likely index or reader failure"
    );
}
```

_Confidence: 82% · Reliability_

crates/rskim-bench/src/extract/rust_lang.rs
```diff
@@ -0,0 +1,237 @@
...
+        tree_sitter_rust::LANGUAGE.into(),
+        move |node, bytes, symbols| {
+            match node.kind() {
+                "function_item" => {
+                    if let Some(name_node) = node.child_by_field_name("name")
+                        && let Ok(name) = name_node.utf8_text(bytes)
+                    {
+                        symbols.push(ExtractedSymbol {
+                            name: name.to_string(),
+                            file_path: path.clone(),
```
dean0x: **[MEDIUM: Performance]** `path.clone()` on every symbol in extractor

Each extractor clones the `PathBuf` on every symbol push inside the closure (`path.clone()`). For files with many symbols this creates unnecessary allocations. The `path` variable is captured by `move` into the closure; every clone allocates a new heap buffer.

**Fix:** Use `Arc<PathBuf>` or `Rc<PathBuf>` for zero-cost clones:
```rust
pub fn extract(path: &Path, content: &str) -> Vec<ExtractedSymbol> {
    let path = std::sync::Arc::new(path.to_path_buf());
    super::walk_ast(
        content,
        tree_sitter_rust::LANGUAGE.into(),
        move |node, bytes, symbols| {
            // ... use (*path).clone() for Arc sharing ...
        },
    )
}
```

_Confidence: 85% · Performance, Rust_ (Note: Benchmark crate, not hot path, but worth considering)

crates/rskim-bench/src/extract/go.rs
```diff
@@ -0,0 +1,243 @@
...
+        tree_sitter_go::LANGUAGE.into(),
+        move |node, bytes, symbols| {
+            match node.kind() {
+                "function_declaration" | "method_declaration" => {
+                    if let Some(name_node) = node.child_by_field_name("name")
+                        && let Ok(name) = name_node.utf8_text(bytes)
+                    {
+                        symbols.push(ExtractedSymbol {
+                            name: name.to_string(),
+                            file_path: path.clone(),
```
dean0x: **[MEDIUM: Performance]** `path.clone()` on every symbol in extractor

Each extractor clones the `PathBuf` on every symbol push inside the closure (`path.clone()`). For files with many symbols this creates unnecessary allocations.

**Fix:** Use `Arc<PathBuf>` for zero-cost clones (see rust_lang.rs for pattern).

_Confidence: 85% · Performance, Rust_ (Note: Benchmark crate, not hot path, but worth considering)

crates/rskim-bench/src/extract/python.rs
```diff
@@ -0,0 +1,235 @@
...
+        tree_sitter_python::LANGUAGE.into(),
+        move |node, bytes, symbols| {
+            match node.kind() {
+                "function_definition" => {
+                    if let Some(name_node) = node.child_by_field_name("name")
+                        && let Ok(name) = name_node.utf8_text(bytes)
+                    {
+                        symbols.push(ExtractedSymbol {
+                            name: name.to_string(),
+                            file_path: path.clone(),
```
dean0x: **[MEDIUM: Performance]** `path.clone()` on every symbol in extractor

Each extractor clones the `PathBuf` on every symbol push inside the closure (`path.clone()`). For files with many symbols this creates unnecessary allocations.

**Fix:** Use `Arc<PathBuf>` for zero-cost clones (see rust_lang.rs for pattern).

_Confidence: 85% · Performance, Rust_ (Note: Benchmark crate, not hot path, but worth considering)

dean0x:
## Summary: Deduplicated Review Findings

### High-Confidence Blocking Issues (≥80%)

Created 10 inline comments for these findings:

1. **sweep_parameter 9 parameters** (tuning.rs:54) — Extract SweepState struct to reduce from 9 to 6 parameters
2. **from_value stale capture** (tuning.rs:67) — Move capture inside improvement branch for accurate convergence trace
3. **walk_nodes unbounded recursion** (extract/mod.rs:90) — Add MAX_DEPTH: 256 bound to prevent stack overflow
4. **field_display_name duplication** (report.rs:32) — Consolidate variant-to-string mapping to single source of truth
5. **run_tune 120-line orchestration** (main.rs:339) — Extract load_all_repos + run_final_eval_and_report helpers
6. **ID reassignment silent drops** (main.rs:360) — Assert contents.remove() never returns None for data integrity
7. **Tuning error silent saturation** (main.rs:386) — Fail hard if evaluation errors produce 0.0 MRR
8. **PathBuf::clone in extractors** (rust_lang.rs:31, go.rs:29, python.rs:28) — Use Arc<PathBuf> for zero-cost clones

All are marked with ≥80% confidence per the operation criteria.

### Rate Limit & Deduplication
- Successfully posted 10 inline comments with 1-2s throttling between calls
- Multiple reviewers independently flagged the same issues (walk_nodes, sweep_parameter, PathBuf cloning)
- Consolidated into single inline comments at each location to avoid spam

_Created by: Claude Code Git Agent · Deduplicated across Architecture, Complexity, Rust, Reliability reviews_

PR #249, opened 2026-05-22
feat: populate SearchResult.line_range with real line numbers

dean0x:
## Summary

- `SearchResult.line_range` was hardcoded to `0..0` by the index reader. This PR populates it with real 1-indexed line numbers computed from `match_positions` during snippet extraction, where file content is already in memory.
- Two new public functions `byte_offset_to_line` and `compute_line_range` are added to `rskim-search::types` and re-exported from the crate root.
- A new `line_range: Option<Range<usize>>` field on `ResolvedResult` makes the range available in `--format json` output automatically via `serde`.

## Changes

- **`crates/rskim-search/src/types.rs`** — added `byte_offset_to_line` (returns `usize`, 1-indexed) and `compute_line_range`; 12 new unit tests
- **`crates/rskim-search/src/lib.rs`** — exported both new functions from the crate root
- **`crates/rskim/src/cmd/search/snippet.rs`** — extended `SnippetOutcome::Ok` to carry `Range<usize>` line_range; calls `rskim_search::compute_line_range` after file read
- **`crates/rskim/src/cmd/search/query.rs`** — threaded `line_range` through `resolve_paths_and_snippets` into `ResolvedResult`
- **`crates/rskim/src/cmd/search/types.rs`** — added `line_range: Option<Range<usize>>` field to `ResolvedResult`
- **`crates/rskim/src/cmd/search/snippet_tests.rs`** — updated existing test destructuring; added `test_extract_snippet_computes_line_range`
- **`crates/rskim/src/cmd/search/query_tests.rs`** — added `line_range` field to two `ResolvedResult` literals

## Breaking Changes

None. `SearchResult.line_range` was already part of the public API (previously always `0..0`). The new `ResolvedResult.line_range` field is additive — existing JSON consumers that ignore unknown fields are unaffected.

## Reviewer Focus Areas

- `compute_line_range` in `types.rs` — ensure the exclusive-end semantics (`min..(max+1)`) match the existing `SearchResult::line_range` doc comment ("0-indexed, exclusive end" — note: new functions use 1-indexed, which is what the CLI needs)
- `SnippetOutcome::Ok` tuple arity change — all three match sites updated: `snippet.rs`, `snippet_tests.rs`, `query.rs`
- No changes to `reader.rs` (which still emits `0..0`) or `FORMAT_VERSION` — index format is unchanged

### DIFF
--- a/crates/rskim-search/src/lib.rs
+++ b/crates/rskim-search/src/lib.rs
@@ -31,6 +31,6 @@ pub use temporal::{GixSource, is_fix_commit};
 pub use types::{
     CommitInfo, FieldClassifier, FileChangeInfo, FileId, HistoryResult, IndexStats, LayerBuilder,
     NodeInfo, Result, SearchError, SearchField, SearchLayer, SearchQuery, SearchResult,
-    TemporalFlags, TemporalMetadata, TemporalSource,
+    TemporalFlags, TemporalMetadata, TemporalSource, byte_offset_to_line, compute_line_range,
 };
 pub use weights::{BIGRAM_WEIGHTS, DEFAULT_WEIGHT, bigram_weight};
--- a/crates/rskim-search/src/types.rs
+++ b/crates/rskim-search/src/types.rs
@@ -347,10 +347,7 @@ pub struct SearchResult {
 /// Counts newlines in `content[..offset]`. The offset is clamped to
 /// `content.len()` so out-of-bounds values never panic. Returns `1` for
 /// offset `0` or any offset in empty content.
-///
-/// This is the library counterpart of the CLI's `snippet::byte_offset_to_line`;
-/// the two are intentionally kept separate — this one returns `usize` while
-/// the CLI copy returns `u32` for display formatting.
+#[must_use]
 pub fn byte_offset_to_line(content: &[u8], offset: usize) -> usize {
     let safe_offset = offset.min(content.len());
     let newlines = content[..safe_offset]
@@ -367,23 +364,18 @@ pub fn byte_offset_to_line(content: &[u8], offset: usize) -> usize {
 /// 1-indexed), matching the convention used by [`SearchResult::line_range`].
 ///
 /// Returns `0..0` when `match_positions` is empty.
+#[must_use]
 pub fn compute_line_range(content: &[u8], match_positions: &[Range<usize>]) -> Range<usize> {
     if match_positions.is_empty() {
         return 0..0;
     }
 
-    let mut min_line = usize::MAX;
-    let mut max_line = 0usize;
+    let lines = match_positions
+        .iter()
+        .map(|pos| byte_offset_to_line(content, pos.start));
 
-    for pos in match_positions {
-        let line = byte_offset_to_line(content, pos.start);
-        if line < min_line {
-            min_line = line;
-        }
-        if line > max_line {
-            max_line = line;
-        }
-    }
+    let min_line = lines.clone().min().unwrap_or(1);
+    let max_line = lines.max().unwrap_or(1);
 
     min_line..(max_line + 1)
 }
--- a/crates/rskim-search/src/types.rs
+++ b/crates/rskim-search/src/types.rs
@@ -339,6 +339,56 @@ pub struct SearchResult {
 }
 
 // ============================================================================
+// Line-range utilities
+// ============================================================================
+
+/// Map a byte offset within `content` to a **1-indexed** line number.
+///
+/// Counts newlines in `content[..offset]`. The offset is clamped to
+/// `content.len()` so out-of-bounds values never panic. Returns `1` for
+/// offset `0` or any offset in empty content.
+///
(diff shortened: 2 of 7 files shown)


dean0x:
### Duplicate `byte_offset_to_line` with inconsistent return type

You now have two `byte_offset_to_line` functions with identical logic but different return types:
- Library version in `rskim-search/src/types.rs` returns `usize` (1-indexed)
- Local version in `snippet.rs` returns `u32`

Both are active: line 160 uses the local `u32` version, and line 162 calls `compute_line_range` which internally uses the library `usize` version. This creates a maintenance risk where a bug fix in one could be missed in the other.

**Suggested fix:** Replace the local call with a cast to eliminate the duplicate:
```rust
let match_line = rskim_search::byte_offset_to_line(&content, match_positions[0].start) as u32;
```
Then remove the local `byte_offset_to_line` function.

*Deduped finding from 6 reviewers (92% confidence)*

dean0x:
### Stale doc comment on `SearchResult::line_range`

The doc comment says "0-indexed" but `compute_line_range` (which populates this field) returns **1-indexed** values:

Current (incorrect):
```rust
/// Source lines spanned by this match (0-indexed, exclusive end)
pub line_range: Range<usize>,
```

The function documentation at line 364 correctly states "1-indexed," but the field doc contradicts it. This would confuse consumers into expecting 0-indexed values when actual populated values are 1-indexed.

**Fix:** Update the field doc comment:
```rust
/// Source lines spanned by this match (1-indexed, exclusive end; 0..0 when not yet computed)
pub line_range: Range<usize>,
```

*Deduped finding from 5 reviewers (95% confidence)*

dean0x:
### Redundant O(n) scan: `byte_offset_to_line` called twice for the same position

At line 160-162, you call `byte_offset_to_line` for `match_positions[0].start`, then immediately call `compute_line_range` which re-scans the same byte offset internally. This means the first match position's slice is scanned for newlines twice.

Each call to `byte_offset_to_line` is O(offset) — it counts newlines in `content[..offset]`. For a match at byte 100,000 this is two 100KB scans instead of one.

**Suggested fix:** Derive `match_line` from the result of `compute_line_range` or compute `match_line` first and short-circuit when there's only one position:

```rust
let match_line = rskim_search::byte_offset_to_line(&content, match_positions[0].start) as u32;

let line_range = if match_positions.len() == 1 {
    (match_line as usize)..(match_line as usize + 1)
} else {
    rskim_search::compute_line_range(&content, match_positions)
};
```

*Performance finding (85% confidence)*

dean0x:
### `SnippetOutcome::Ok` tuple growing to 3 fields with non-obvious ordering

The enum variant now holds a 3-tuple: `Ok(u32, Range<usize>, SnippetContext)`. Positional tuples with non-obvious field ordering are harder to maintain than named struct variants.

Current:
```rust
Ok(u32, std::ops::Range<usize>, SnippetContext),
```

**Suggested fix:** Use a named struct variant for clarity:
```rust
Ok {
    match_line: u32,
    line_range: Range<usize>,
    context: SnippetContext,
}
```

This makes call sites and the enum definition self-documenting. You'd update `extract_snippet` return to `SnippetOutcome::Ok { match_line, line_range, context }` instead of `Ok(match_line, line_range, context)`.

*Deduped finding from 3 reviewers (MEDIUM, 80% confidence)*

dean0x:
## Summary: Review Findings (4 inline + suggestions below)

### Inline Comments (≥80% confidence)
Posted 4 blocking/high-priority comments above covering:
1. Duplicate `byte_offset_to_line` with inconsistent return types (92%)
2. Stale doc on `SearchResult::line_range` saying "0-indexed" (95%)
3. Redundant byte scan at lines 160-162 (85%)
4. `SnippetOutcome::Ok` tuple with non-obvious field ordering (80%)

### Suggestions (60-79% confidence)

**Missing JSON serialization test for `ResolvedResult.line_range`** (85% confidence, HIGH priority)
- The new `line_range: Option<Range<usize>>` field is documented to serialize as `{"start": N, "end": M}` in JSON output, but there's no test asserting this shape
- Current tests construct `ResolvedResult` with line_range but only validate text output
- Add a test: create `ResolvedResult` with `line_range: Some(5..10)`, serialize via `format_json_output`, assert JSON contains `"line_range": {"start": 5, "end": 10}`
- Also add a test for the `None` case, asserting it serializes as `"line_range": null`

**Iterator clone in `compute_line_range` (LOW, 82% confidence)**
- The `lines` iterator is cloned to compute `min()` and `max()` separately, doubling newline-count work
- Use a single-pass fold instead (detailed fix in performance review)
- Impact is low (match positions typically single digits) but improves algorithmic correctness

**`u32`/`usize` line number type inconsistency** (Suggestion, 65% confidence)
- CLI layer uses `u32` throughout; new library functions use `usize`
- Consider standardizing on one type in a future pass

---

**Deduped from 9 reviewers.** Rate limit status: 4,200+ remaining. ✅

Issue #183, opened 2026-05-09
Wave 2c: Hotspot and risk scoring


dean0x:
## Overview
Compute per-file metrics: change frequency (hotspot) and bug-fix density (risk).

## Metrics
- **Hotspot:** Count of commits touching file in 30d and 90d windows
- **Bug-fix density:** Count of fix/bug/revert commits / total commits per file
- **Decay function:** Exponential decay, half-life 30 days (recent changes weighted more)

## Dependencies
- #7 (git parser)

## Acceptance Criteria
- [ ] File modified 50x in 30d ranks higher than one modified 5x
- [ ] Bug-fix density of 0.0 for files never in fix commits
- [ ] Scores in [0.0, 1.0] range (normalized)
- [ ] Decay function applies correctly

## Implementation Guidelines Checklist
- [ ] No file exceeds 400 lines
- [ ] Every public function has /// doc comment
- [ ] All dependencies injected
- [ ] All fallible operations return Result<T, SearchError>
- [ ] Module has #[cfg(test)] mod tests

dean0x:
Closed: implemented in #252 (dd67b38)

PR #252, opened 2026-05-25
feat(temporal): hotspot and bug-fix density scoring with exponential decay

dean0x:
## Summary

Adds per-file hotspot and bug-fix density metrics to the `rskim-search` temporal module (issue #183). All computation is pure — no I/O, no new crate dependencies, deterministic results given a fixed `now_epoch` timestamp.

- `decay_weight(elapsed_days, half_life_days)` — exponential decay kernel, clamped to [0.0, 1.0]
- `compute_file_risk_scores(commits, now_epoch, half_life_days)` — single-pass accumulation of decay-weighted hotspot and fix-density per file
- `FileRiskScores { hotspot, fix_density }` — both fields in [0.0, 1.0]; hotspot is max-normalized, fix_density is a weighted ratio

## Changes

**New files:**
- `crates/rskim-search/src/temporal/scoring.rs` — core implementation (~120 lines)
- `crates/rskim-search/src/temporal/scoring_tests.rs` — 47 tests across 6 groups (~190 lines)

**Modified files:**
- `crates/rskim-search/src/types.rs` — `FileRiskScores` struct added alongside existing git history types
- `crates/rskim-search/src/temporal/mod.rs` — `mod scoring` wired; `decay_weight`, `compute_file_risk_scores`, `DEFAULT_HALF_LIFE_DAYS` re-exported; doc comment updated
- `crates/rskim-search/src/lib.rs` — `FileRiskScores`, `decay_weight`, `compute_file_risk_scores`, `DEFAULT_HALF_LIFE_DAYS` added to public re-exports; architecture comment updated

## Breaking Changes

None. All new additions are additive public API.

## Reviewer Focus Areas

- **Algorithm correctness** (`scoring.rs:75-115`): single-pass accumulation, timestamp clamping (`commit.timestamp.max(0) as u64`), elapsed-vs-future guard, max-normalization
- **Decay math** (`scoring.rs:50-55`): `(-elapsed / half_life).exp().clamp(0.0, 1.0)` — negative elapsed naturally produces values > 1.0 before clamping, so future commits correctly get weight 1.0
- **Test coverage** (`scoring_tests.rs`): 47 tests across decay unit tests, basic cases, acceptance criteria, fix-density specifics (keyword recognition, per-file independence, decay-weighted ratio), edge cases (future/negative/ancient timestamps), and determinism
- **No unwrap/expect/panic** outside `#[cfg(test)]` — consistent with workspace clippy deny list

## Test plan

- [x] `cargo test -p rskim-search -- scoring` — 47 tests, all pass
- [x] `cargo test -p rskim-search` — 388 tests, 0 failures, 3 skipped (git-integration tests requiring live git binary)
- [x] `cargo clippy -p rskim-search -- -D warnings` — 0 warnings, 0 errors
- [x] `cargo check --workspace` — full workspace clean

### DIFF
--- a/crates/rskim-search/src/temporal/scoring.rs
+++ b/crates/rskim-search/src/temporal/scoring.rs
@@ -38,7 +38,7 @@ pub const DEFAULT_HALF_LIFE_DAYS: f64 = 30.0;
 ///
 /// # Panics
 ///
-/// Panics in debug builds when `half_life_days <= 0.0`.
+/// Panics when `half_life_days <= 0.0` or is not finite.
 ///
 /// # Examples
 ///
@@ -55,7 +55,10 @@ pub const DEFAULT_HALF_LIFE_DAYS: f64 = 30.0;
 #[must_use]
 #[inline]
 pub fn decay_weight(elapsed_days: f64, half_life_days: f64) -> f64 {
-    debug_assert!(half_life_days > 0.0);
+    assert!(
+        half_life_days > 0.0 && half_life_days.is_finite(),
+        "half_life_days must be positive and finite, got {half_life_days}"
+    );
     // Treat NaN elapsed as zero (present-moment weight = 1.0) to prevent
     // NaN propagation into accumulators. clamp() alone does not sanitize NaN.
     let elapsed = if elapsed_days.is_nan() {
@@ -133,14 +136,14 @@ pub fn compute_file_risk_scores(
             // Reduces allocations from O(total_file_touches) to O(unique_files).
             let path_cow = file.path_str();
             let path_ref: &str = &path_cow;
-            let entry = if let Some(v) = accum.get_mut(path_ref) {
+            let (weighted_total, weighted_fix_total) = if let Some(v) = accum.get_mut(path_ref) {
                 v
             } else {
                 accum.entry(path_cow.into_owned()).or_insert((0.0, 0.0))
             };
-            entry.0 += w;
+            *weighted_total += w;
             if is_fix {
-                entry.1 += w;
+                *weighted_fix_total += w;
             }
         }
     }
--- a/crates/rskim-search/src/temporal/scoring.rs
+++ b/crates/rskim-search/src/temporal/scoring.rs
@@ -1,0 +1,143 @@
+/// Temporal hotspot and bug-fix density scoring with exponential decay.
+///
+/// All functions are pure (no I/O, no side effects). Consumers supply the current
+/// epoch timestamp as `now_epoch` so that tests are fully deterministic.
+///
+/// # Algorithm overview
+///
+/// [`compute_file_risk_scores`] performs a single pass over the commit list,
+/// accumulating decay-weighted totals per file. Hotspot scores are then
+/// max-normalized so the busiest file always scores 1.0. Fix density is the
+/// ratio of fix-weighted touches to total weighted touches per file.
+use std::collections::HashMap;
+
+use crate::types::{CommitInfo, FileRiskScores};
+
+/// Default half-life (in days) used when callers do not supply a custom value.
+///
+/// A 30-day half-life means commits from one month ago contribute ~37% as much
+/// weight as a commit made today.
+pub const DEFAULT_HALF_LIFE_DAYS: f64 = 30.0;
+
+/// Exponential decay weight for a single commit.
+///
+/// Returns `exp(-elapsed_days / half_life_days)`, clamped to `[0.0, 1.0]`.
+/// A negative `elapsed_days` (future commit) is treated as zero elapsed time
+/// and therefore returns `1.0`.
+///
+/// # Panics
+///
+/// Panics in debug builds when `half_life_days <= 0.0`.
+///
+/// # Examples
+///
+/// ```rust
+/// use rskim_search::decay_weight;
+///
+/// let w = decay_weight(0.0, 30.0);
+/// assert_eq!(w, 1.0);
+///
+/// let w_half = decay_weight(30.0, 30.0);
+/// // ≈ 1/e ≈ 0.368
+/// assert!((w_half - std::f64::consts::E.recip()).abs() < 1e-9);
+/// ```
+#[must_use]
+#[inline]
+pub fn decay_weight(elapsed_days: f64, half_life_days: f64) -> f64 {
+    debug_assert!(half_life_days > 0.0);
+    (-elapsed_days / half_life_days).exp().clamp(0.0, 1.0)
+}
+
+/// Compute per-file hotspot and bug-fix density scores from a git commit history.
+///
+/// # Parameters
+///
+/// - `commits`: Slice of [`CommitInfo`] values (any order).
+/// - `now_epoch`: Current Unix timestamp in seconds (used for elapsed-time computation).
+///   Pass a fixed value in tests for full determinism.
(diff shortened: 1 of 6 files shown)


dean0x (commented):
## Code Review: High-Confidence Issues (≥80%)

Ready to create inline comments for 6 blocking/should-fix issues from the architecture, Rust, and testing reviews.

Proceeding with the comment creation...

dean0x:
## Code Review: High-Confidence Issues (≥80%)

Based on comprehensive code review of the hotspot & risk scoring implementation, here are the blocking and should-fix issues:

### 1. FileRiskScores missing `Copy` derive (Architecture Review - 85% confidence)
**File**: `crates/rskim-search/src/types.rs:268`

The `FileRiskScores` struct contains only two `f64` fields (16 bytes) but only derives `Debug, Clone`. This is inconsistent with similar small value types in the same file. Add `Copy` and `PartialEq`:

```rust
#[derive(Debug, Clone, Copy, PartialEq)]
pub struct FileRiskScores {
```

### 2. Parameter name `half_life_days` is misleading (Rust Review - 82% confidence)
**File**: `crates/rskim-search/src/temporal/scoring.rs:46`

The formula `exp(-elapsed_days / half_life_days)` implements an e-folding time, not a true half-life. At `t = half_life_days`, weight reaches ~0.368 (1/e), not 0.5. Rename to `decay_constant_days` or update the formula to use true half-life math.

### 3. `debug_assert!` on public API boundary should be `assert!` (Rust Review - 85% confidence)
**File**: `crates/rskim-search/src/temporal/scoring.rs:47, 81`

Both `decay_weight` and `compute_file_risk_scores` guard `half_life_days > 0.0` with `debug_assert!`, which is stripped in release builds. This violates CLAUDE.md: "assert! at module boundaries." Use `assert!` instead:

```rust
assert!(half_life_days > 0.0, "half_life_days must be positive");
```

### 4. Missing test for `compute_file_risk_scores` with zero half_life (Testing Review - 85% confidence)
**File**: `crates/rskim-search/src/temporal/scoring_tests.rs`

No `#[should_panic]` test for the outer function despite having `debug_assert!(half_life_days > 0.0)`. Add:

```rust
#[test]
#[cfg(debug_assertions)]
#[should_panic]
fn compute_scores_zero_half_life_panics() {
    let commits = vec![make_commit(NOW, "feat", &["a.rs"])];
    let _ = compute_file_risk_scores(&commits, NOW, 0.0);
}
```

### 5. Missing test for multi-file commits (Testing Review - 82% confidence)
**File**: `crates/rskim-search/src/temporal/scoring_tests.rs`

No test verifying that a single commit touching multiple files applies the same weight to all files. Add:

```rust
#[test]
fn single_commit_multiple_files_same_weight() {
    let commits = vec![make_commit(NOW - 10 * DAY, "feat: wide change", &["a.rs", "b.rs", "c.rs"])];
    let scores = compute_file_risk_scores(&commits, NOW, HALF_LIFE);
    assert_eq!(scores.len(), 3);
    let expected = scores["a.rs"].hotspot;
    assert!(approx_eq(expected, 1.0));
    assert!(approx_eq(scores["b.rs"].hotspot, expected));
    assert!(approx_eq(scores["c.rs"].hotspot, expected));
}
```

### 6. Hot loop allocation in scoring (Architecture Review - 82% confidence)
**File**: `crates/rskim-search/src/temporal/scoring.rs:112`

`file.path_str().into_owned()` allocates a new `String` for every file in every commit. When the same file appears in many commits, this creates duplicate allocations. Consider using `Cow<str>` as the HashMap key or pre-interning paths. This is performance polish, acceptable to defer if profiling shows it's not a bottleneck.

---

**Summary**: 3 BLOCKING (Copy derive, debug_assert guards), 2 HIGH (misleading parameter name, missing tests), 1 MEDIUM (hot loop allocation). All have ≥80% confidence.

crates/rskim-search/src/temporal/scoring.rs
```diff
@@ -0,0 +1,186 @@
...
+            if is_fix {
+                entry.1 += w;
+            }
+        }
+    }
+
+    // Find the maximum weighted total for normalization.
+    let max_total = accum
+        .values()
+        .map(|&(total, _)| total)
```
dean0x, line 154: **[BLOCKING] Public function should use `assert!`, not `debug_assert!`** (92% confidence, reliability)

The `decay_weight` function is publicly exported (via `mod.rs:17`) but validates its `half_life_days > 0.0` precondition with `debug_assert!`, which is stripped in release builds. In release mode:
- Passing `half_life_days = 0.0` produces `(-elapsed / 0.0).exp()` → `NaN` or `Inf`
- `.clamp(0.0, 1.0)` does **not** catch `NaN` (NaN comparisons are always false)
- The division propagates NaN silently into caller accumulators

**Fix:** Upgrade to `assert!` and add a NaN check per cycle-1 pattern (already fixed in `compute_file_risk_scores` at line 99):

```rust
assert!(
    half_life_days > 0.0 && half_life_days.is_finite(),
    "half_life_days must be positive and finite, got {half_life_days}"
);
```

Also update the doc comment (line 41) from "Panics in debug builds" to "Panics when `half_life_days <= 0.0`." and remove `#[cfg(debug_assertions)]` from the test at line 126-131.

---
*Claude Code | /code-review cycle-2 consolidation*

dean0x:
## Code Review: Cycle 2 Consolidation

### Inline Comments
I've created 1 inline comment on the blocking  issue in `decay_weight` (92% confidence, reliability). This is the critical high-confidence finding across cycle-2 reviewers.

### Consolidated Findings

**Status:** CHANGES_REQUESTED

The temporal scoring module has strong reliability fundamentals (bounded allocations, future-timestamp clamping, negative-timestamp clamping, division-by-zero guard on `max_total`). Two findings consolidate into one architectural asymmetry:

1. **Blocking (92% confidence):** `decay_weight` uses `debug_assert!` while the public boundary `compute_file_risk_scores` uses `assert!` — the former silently permits NaN/Inf/zero in release builds. **Inline comment with fix provided.**

2. **Should-Fix (85% confidence):** NaN guard on `elapsed_days` (lines 61–65) but not on `half_life_days`. If the latter is NaN, it propagates silently into accumulators. **Consolidated into the `assert!` upgrade** — add `.is_finite()` check and new test case.

### Lower-Confidence Suggestions (60–79%)

These do not block but are noted for future optimization:

- **Capacity heuristic may over-allocate (65% confidence):** `capacity = (commits.len() / 4).clamp(64, 50_000)` uses a fixed 64-entry floor. For tiny repositories (1–10 commits), this allocates 64 HashMap slots (~5KB) when fewer would suffice. Mitigation: The impact is negligible in practice; inline comment in code justifying the heuristic would improve clarity.

- **`fix_flags` Vec could be inlined (62% confidence):** Pre-allocating a `Vec<bool>` per commit avoids repeated regex evaluation in the hot loop (readability win), but for 50K+ commits this is 50KB of allocation. Mitigation: The separation improves code clarity and the cost is trivial relative to the HashMap; keep as-is unless profiling shows contention.

- **`debug_assert!` vs `assert!` asymmetry with documentation (72% confidence, Rust review):** The intention is clear in doc comments, but future maintainers unfamiliar with cycle-1 fixes might assume both should have the same guard style. Addressed by the inline comment.

### Summary
| Category | Count |
|----------|-------|
| Blocking (HIGH) | 1 |
| Should-Fix (HIGH) | 1 |
| Lower-Confidence Suggestions | 3 |

---
*Claude Code | /code-review consolidation*

Issue #185, opened 2026-05-09
Wave 2d: Temporal metadata storage


dean0x:
## Overview
Persist temporal data (co-change, hotspot, risk) to SQLite for fast loading.

## Schema
\`\`\`sql
CREATE TABLE cochange (file_a TEXT, file_b TEXT, count INTEGER, jaccard REAL, PRIMARY KEY(file_a, file_b));
CREATE TABLE hotspot (file TEXT PRIMARY KEY, changes_30d INTEGER, changes_90d INTEGER, score REAL);
CREATE TABLE risk (file TEXT PRIMARY KEY, total_commits INTEGER, fix_commits INTEGER, score REAL);
CREATE TABLE meta (key TEXT PRIMARY KEY, value TEXT);  -- version, last_updated, git_head
\`\`\`

## Pattern
Follow schema migration pattern from `crates/rskim/src/analytics/schema.rs`. WAL mode for concurrent access.

## Dependencies
- #9 (co-change), #11 (hotspot/risk)

## Acceptance Criteria
- [ ] Load data for 10k files in < 100ms
- [ ] Schema versioned with migration support
- [ ] Data survives process restarts
- [ ] WAL mode enabled

## Implementation Guidelines Checklist
- [ ] No file exceeds 400 lines
- [ ] Every public function has /// doc comment
- [ ] All dependencies injected
- [ ] All fallible operations return Result<T, SearchError>
- [ ] Module has #[cfg(test)] mod tests

dean0x:
Closed: implemented in #253 (6b6aeab)

PR #253, opened 2026-05-25
feat(#185): TemporalDb SQLite persistence layer with WAL mode and atomic sync

dean0x:
## Summary

- Temporal risk scores and co-change data are recomputed from git history on every search invocation. For large repositories this adds seconds of latency. This PR introduces a SQLite persistence layer so the computed data can be cached and reused across invocations.
- Adds `TemporalDb` struct, `sync()` for atomic multi-table refresh, `compute_file_temporal_stats()` for raw per-file commit counts, and `SearchError::Database` for clean error propagation.

## Changes

**New types:**
- `FileTemporalStats` — raw per-file commit counts within 30d/90d windows (non-decay-weighted, suitable for persistence)
- `HotspotRow`, `RiskRow`, `CochangeRow` — typed row structs matching the SQLite schema
- `SearchError::Database(String)` — no rusqlite types leak through the public API

**New functions:**
- `compute_file_temporal_stats(commits, now_epoch)` — pure function, single pass, per-commit file deduplication via HashSet, boundary-inclusive (`<=`) window checks
- `TemporalDb::open(path)` — WAL mode, versioned schema migrations, forward-compat guard, 5s busy timeout, 0o600 permissions on Unix
- `TemporalDb::sync(hotspots, risks, cochanges, git_head)` — atomically replaces all 4 tables + meta in one transaction
- Individual `store_*` / `load_*` methods with `#[must_use]` on all load paths

**Tests (47 new):**
- Schema lifecycle: WAL mode, schema version, idempotent reopen, Unix permissions, future-version rejection
- CRUD roundtrips: hotspot/risk/cochange store-and-load, replace semantics, empty-DB loads
- Meta: set/get/overwrite
- Persistence: data survives close/reopen
- Sync atomicity: all tables written, meta keys set
- Performance: 10k rows load <100ms, 10k rows store <200ms, 10k×3 sync <500ms
- Error handling: invalid path, error display, variant matchability

## Breaking Changes

`SearchError` gains a new `Database(String)` variant. No exhaustive matches exist in the codebase — verified safe.

## Reviewer Focus Areas

- **Transaction semantics in `sync()`**: DELETE + INSERT across 4 tables + meta in one transaction. If any step fails, the entire transaction rolls back.
- **Forward-compat guard in `run_migrations`**: rejects databases with `user_version > CURRENT_VERSION` to prevent silent corruption when old code opens a new-format DB.
- **`compute_file_temporal_stats` boundary conditions**: 30d/90d edges use `<=`, future commits clamp to `elapsed=0`, negative timestamps clamp via `.max(0)`, duplicate file paths in one commit deduplicated via `HashSet`.
- **Performance test methodology**: `Instant::now()` around load/store calls after pre-populating 10k rows; asserts on `elapsed.as_millis()`.

### DIFF
--- a/.devflow/features/cochange/KNOWLEDGE.md
+++ b/.devflow/features/cochange/KNOWLEDGE.md
@@ -1,7 +1,7 @@
 ---
 feature: cochange
 name: Co-Change Matrix
-description: "Use when implementing co-change coupling queries, modifying the .skcc binary format, adding new query methods to CochangeMatrixReader, or debugging Jaccard similarity calculations. Keywords: cochange, co-change, coupling, jaccard, skcc, binary format, cochange.skcc, CochangeMatrixBuilder, CochangeMatrixReader, HistoryResult, COUPLING_MAX_FILES."
+description: "Use when implementing co-change coupling queries, modifying the .skcc binary format, adding new query methods to CochangeMatrixReader, debugging Jaccard similarity calculations, or working with the SQLite temporal persistence layer for co-change pairs. Keywords: cochange, co-change, coupling, jaccard, skcc, binary format, cochange.skcc, CochangeMatrixBuilder, CochangeMatrixReader, CochangeRow, TemporalDb, HistoryResult, COUPLING_MAX_FILES."
 category: domain-knowledge
 directories: [crates/rskim-search/src/cochange/]
 referencedFiles:
@@ -11,8 +11,11 @@ referencedFiles:
   - crates/rskim-search/src/cochange/reader.rs
   - crates/rskim-search/src/types.rs
   - crates/rskim-search/src/lib.rs
+  - crates/rskim-search/src/temporal/storage.rs
+  - crates/rskim-search/src/temporal/storage_types.rs
+  - crates/rskim-search/src/temporal/storage_ops.rs
 created: 2026-05-24
-updated: 2026-05-24
+updated: 2026-05-25
 ---
 
 # Co-Change Matrix
@@ -29,7 +32,20 @@ Co-change coupling is a static approximation of runtime coupling: files that hav
 
 Two safety constants govern data quality:
 - `COUPLING_MAX_FILES = 50` — commits touching more than 50 files are bulk refactors or merges. Including them would pollute coupling signal with unrelated co-changes. These commits are counted in `CochangeStats::commits_skipped_too_large`.
-- `MAX_PAIRS = 2_000_000` — bounds memory growth during accumulation. Exceeding this returns `SearchError::IndexCorrupted`.
+- `MAX_PAIRS = 2_000_000` — bounds memory growth during accumulation. Exceeding this returns `SearchError::CapacityExceeded`.
+
+## Dual Persistence Model
+
+Co-change data has two complementary persistence formats, each optimised for a different access pattern:
+
+| Format | Location | API | Access pattern |
+|--------|----------|-----|----------------|
+| `.skcc` binary | `{index_dir}/cochange.skcc` | `CochangeMatrixReader` | Point queries: `jaccard(a, b)`, `pairs_for_file(id)` |
+| SQLite `cochange` table | `{cache_dir}/temporal.db` | `TemporalDb::load_cochanges` / `store_cochanges` | Bulk-load: all pairs with human-readable paths for ranking pipelines |
+
+The `.skcc` format uses `FileId` (u32 integers) and is memory-mapped. The SQLite `cochange` table stores the same pairs using repo-root-relative path strings (`file_a TEXT`, `file_b TEXT`), making them accessible without a path-map lookup. Pre-computed Jaccard scores are stored in the SQLite row alongside the raw co-change count.
+
+Both formats are always written together during an index refresh. The `.skcc` file is the authoritative coupling store; the SQLite table is a projection for ranking signal aggregation alongside `hotspot` and `risk` data.
 
 ## Core Business Rules
 
@@ -49,9 +65,9 @@ Jaccard(a, b) = count_ab / (count_a + count_b - count_ab)
 
 `count_a` and `count_b` are per-file commit counts (how many commits touched each file individually). The denominator is computed in `u64` to prevent overflow when both files have high commit counts. Returns `0.0` for self-pairs, absent pairs, and zero denominators — the caller always gets a valid `f64`.
 
-### `pairs_for_file` is O(pair_count)
+### `pairs_for_file` is O(log n + k), not O(n)
 
-The reader performs a linear scan over all `PairEntry` records to collect partners for a given file. This is intentional — there is no secondary per-file index in the format. For large repositories (millions of pairs) this may become a bottleneck; binary search over sorted pairs is only used for the point-lookup case (`pair_count`, `jaccard`).
+The reader uses binary search to locate the start of the contiguous `file_a == id` block within the sorted `PairEntry` array, then performs a short linear scan over only the prefix where `file_b == id` might appear (`file_a < id` entries). The previously documented O(pair_count) linear scan was replaced in PR #251. For large repositories this is significantly faster.
 
 ## State Transitions
 
@@ -61,7 +77,7 @@ HistoryResult (from TemporalSource)
       | CochangeMatrixBuilder::build()
       |   1. accumulate_pairs — HashMap<(u32,u32), u32> + HashMap<u32, u32>
       |   2. serialize — sorted byte arrays + CRC32 header
-      |   3. atomic_write — NamedTempFile + persist (rename)
+      |   3. atomic_write — NamedTempFile + sync_all + persist (rename)
       v
 cochange.skcc (on disk)
       |
@@ -74,8 +90,20 @@ CochangeMatrixReader (queryable)
       |
       +-- pair_count(a, b)    binary search over PairEntry array
       +-- jaccard(a, b)       pair_count + file_commits binary searches
-      +-- pairs_for_file(id)  linear scan, sorted by count desc
+      +-- pairs_for_file(id)  binary search to start block + prefix scan
       +-- file_commits(id)    binary search over FileCommitEntry array
+
+HistoryResult + CochangeMatrixReader
+      |
+      | Caller builds CochangeRow slice (path strings + Jaccard values)
+      |
+      | TemporalDb::store_cochanges() or TemporalDb::sync()
+      v
+SQLite cochange table (for bulk ranking access)
+      |
+      | TemporalDb::load_cochanges()
+      v
+Vec<CochangeRow> (file_a: String, file_b: String, count: i64, jaccard: f64)
 ```
 
 ## Technical Implementation Patterns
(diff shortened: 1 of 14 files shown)


dean0x:
**Issue: META_LAST_UPDATED doc says ISO-8601 but stores Unix epoch seconds** (Confidence: 95%)

The doc comment on `META_LAST_UPDATED` reads "Key storing the ISO-8601 UTC timestamp" but `storage_ops.rs:308-312` stores Unix epoch seconds (`SystemTime::now().duration_since(UNIX_EPOCH).as_secs().to_string()`). Consumers parsing this as ISO-8601 will fail.

Either fix the doc to match the implementation:
```rust
/// Key storing the Unix epoch timestamp (seconds) of the last successful [`TemporalDb::sync`].
pub const META_LAST_UPDATED: &str = "last_updated";
```

Or change the implementation to actually store ISO-8601 as documented. The doc fix is simpler since your tests already expect `u64`.

This was flagged as HIGH priority by the architecture, consistency, and rust reviewers.

---
_Review comment from PR #253 code review (2026-05-26)_

dean0x:
**Issue: sync() duplicates DELETE+INSERT logic from store_* methods** (Confidence: 85%)

Lines 260-271 in `sync()` duplicate the exact same DELETE+INSERT pattern from `store_hotspots()` (lines 31-43). The same duplication repeats for `risk` (lines 274-291) and `cochange` (lines 294-305) tables.

This creates a maintenance burden: when the schema evolves (e.g., adding a column to `hotspot`), both the individual method and `sync()` must update or they silently diverge.

**Suggested fix:** Extract private helpers for each table:
```rust
fn insert_hotspots_in_tx(tx: &rusqlite::Transaction, rows: &[HotspotRow]) -> Result<()> {
    tx.execute("DELETE FROM hotspot", []).map_err(db_err)?;
    let mut stmt = tx.prepare_cached(
        "INSERT INTO hotspot (file_path, score, changes_30d, changes_90d) VALUES (?1, ?2, ?3, ?4)"
    ).map_err(db_err)?;
    for row in rows {
        stmt.execute(params![row.file_path, row.score, row.changes_30d, row.changes_90d])
            .map_err(db_err)?;
    }
    drop(stmt);
    Ok(())
}
```

Then have both `store_hotspots()` and `sync()` call `insert_hotspots_in_tx(&tx, hotspots)?;` (and similar for the other tables).

This was flagged as HIGH priority by the architecture and rust reviewers.

---
_Review comment from PR #253 code review (2026-05-26)_

dean0x:
**Issue: Unnecessary String allocations in compute_file_temporal_stats hot loop** (Confidence: 90%)

Line 245 calls `file.path_str().into_owned()` for every file in every commit, creating a new `String` each time. Then line 250 clones it again with `path.clone()` before insertion. This is O(total_file_touches) allocations + O(unique_files_per_commit) clones for the accumulator.

Your sibling function `compute_file_risk_scores()` already solves this with borrow-then-own:
```rust
let path_cow = file.path_str();
if let Some(entry) = accum.get_mut(path_cow.as_ref()) {
    entry.total_commits += 1;
} else {
    accum.insert(path_cow.into_owned(), stats);
}
```

**Suggested fix:** Apply the same pattern here — probe the accumulator with a borrowed `&str` first, only calling `into_owned()` when inserting new entries.

This was flagged as HIGH priority by the performance and rust reviewers.

---
_Review comment from PR #253 code review (2026-05-26)_

dean0x:
**File**: `crates/rskim-search/src/temporal/storage_ops.rs` (lines 107, 130, 323)

**SAFETY comment is inaccurate** — Claims "TemporalDb is not Send/Sync" but `rusqlite::Connection` is `Send` (just not `Sync`). The actual guarantee is that no concurrent access is possible because it's not `Sync`. This distinction matters for future maintainers.

Suggested fix:
```rust
// SAFETY: `TemporalDb` is `Send` but not `Sync` — it can be moved
// to another thread but cannot be shared. No concurrent access to
// the underlying connection can occur, so no nested transaction
// can be active when this method is called.
```

**Confidence**: 92% (reliability reviewer)

dean0x:
**File**: `crates/rskim-search/src/temporal/storage.rs` (lines 97-128)

**Migration execute_batch is not atomic** — If the process crashes after running some CREATE TABLE statements but before PRAGMA user_version = 1, the schema will be partially created. While the current code is idempotent (due to `IF NOT EXISTS`), future migrations with `ALTER TABLE` could leave the database unrecoverable.

Fix: Wrap the entire migration in an explicit transaction:
```rust
if version < 1 {
    conn.execute_batch(
        "BEGIN;
         CREATE TABLE IF NOT EXISTS hotspot (...);
         ...
         PRAGMA user_version = 1;
         COMMIT;",
    )
    .map_err(db_err)?;
}
```

**Confidence**: 82% (rust reviewer)

dean0x:
**File**: `crates/rskim-search/src/temporal/storage.rs` (line 192)

**WAL mode activation not verified** — The code executes `PRAGMA journal_mode=WAL` via `execute_batch` but never checks if WAL actually activated. On read-only filesystems or NFS, SQLite silently falls back to DELETE mode. The feature contract states WAL is active, but this isn't enforced.

Fix: Use `query_row` to verify:
```rust
let journal_mode: String = conn
    .query_row("PRAGMA journal_mode=WAL", [], |row| row.get(0))
    .map_err(db_err)?;
if journal_mode.to_lowercase() != "wal" {
    return Err(SearchError::Database(format!(
        "failed to enable WAL mode; got '{}'", journal_mode
    )));
}
```

**Confidence**: 82% (reliability reviewer)

dean0x:
**File**: `crates/rskim-search/src/temporal/storage_ops.rs` (lines 183-258)

**load_hotspots/load_risks/load_cochanges have no row-count guard** — The `store_*` methods enforce `MAX_ROWS_PER_TABLE = 500_000`, but `load_*` methods use unbounded `SELECT`. If the database is externally modified, loads could allocate arbitrarily large vectors. Violates the reliability principle of bounded operations.

Fix: Add a LIMIT clause or post-load check:
```rust
pub fn load_hotspots(&self) -> Result<Vec<HotspotRow>> {
    let mut stmt = self.conn.prepare(
        "SELECT file_path, score, changes_30d, changes_90d FROM hotspot LIMIT 500001"
    ).map_err(db_err)?;
    let rows = stmt.query_map([], |row| { /* ... */ })
        .map_err(db_err)?
        .collect::<std::result::Result<Vec<_>, _>>()
        .map_err(db_err)?;
    if rows.len() > MAX_ROWS_PER_TABLE {
        return Err(SearchError::CapacityExceeded(
            format!("load_hotspots: {} rows exceeds {}", rows.len(), MAX_ROWS_PER_TABLE)
        ));
    }
    Ok(rows)
}
```

**Confidence**: 80% (rust reviewer)

dean0x:
**File**: `crates/rskim-search/src/temporal/storage_types.rs` (lines 18-20) vs `crates/rskim-search/src/types.rs` (lines 284-290)

**Integer type mismatch: FileTemporalStats uses u32 while row types use i64** — `FileTemporalStats::changes_30d/changes_90d` are `u32`, but `HotspotRow::changes_30d/changes_90d` are `i64` (matching SQLite INTEGER). This creates friction at the conversion boundary. Standardize on `u32` in row types and widen with `i64::from()` only at SQLite insertion:

```rust
pub struct HotspotRow {
    pub file_path: String,
    pub score: f64,
    pub changes_30d: u32,  // was i64
    pub changes_90d: u32,  // was i64
}

pub struct RiskRow {
    pub file_path: String,
    pub risk_score: f64,
    pub total_commits: u32,  // was i64
    pub fix_commits: u32,    // was i64
    pub fix_density: f64,
}
```

Then in the insert helpers, use `i64::from()` at the params boundary.

**Confidence**: 90% (consistency reviewer)

dean0x:
**File**: `crates/rskim-search/src/temporal/scoring.rs` (lines 129, 233)

**Magic number 86_400.0 (seconds per day) appears twice without a named constant** — This literal is used in both `compute_file_risk_scores` (line 129, pre-existing) and `compute_file_temporal_stats` (line 233). Extract to a module-level constant:

```rust
/// Seconds in one day, used for epoch-to-days conversion.
const SECS_PER_DAY: f64 = 86_400.0;
```

Then use `SECS_PER_DAY` in both functions.

**Confidence**: 85% (complexity reviewer)

dean0x:
**File**: `crates/rskim-search/src/temporal/scoring.rs` (lines 214-275)

**compute_file_temporal_stats exceeds the 50-line function length threshold** — At 61 lines, this function crosses the warning threshold (max nesting depth 4). While readable, consider extracting the per-commit deduplication (lines 245-252) into a helper to reduce the main function to ~48 lines.

```rust
fn dedup_changed_files<'a>(
    commit: &CommitInfo,
    buf: &'a mut HashSet<String>,
) -> &'a HashSet<String> {
    buf.clear();
    for file in &commit.changed_files {
        let path_cow = file.path_str();
        let path_ref: &str = &path_cow;
        if !buf.contains(path_ref) {
            buf.insert(path_cow.into_owned());
        }
    }
    buf
}
```

This isolates the borrow-first optimization for independent testing.

**Confidence**: 82% (complexity reviewer)

dean0x:
## Summary: Lower-Confidence Findings & Test Coverage Gaps

These don't require immediate fixes but are worth considering:

1. **Structural repetition in store/load methods** (72% confidence) — The three `store_*` and three `load_*` methods follow identical patterns. This is acceptable for now (3 instances each, well under consolidation threshold), but watch for growth.

2. **`sync` method approaching warning threshold** (65% confidence) — At 47 lines, close to the 50-line zone. Early-return capacity checks could be extracted into a validation helper if further growth occurs.

3. **`prepare_cached` in load methods** (62% confidence) — Using `prepare` instead of `prepare_cached` in `load_*` methods. If loads are called in loops, cached preparation would save SQLite compilation overhead.

4. **Duplicate `temp_db()` test helper** (65% confidence) — Defined identically in both `storage_tests.rs` and `storage_perf_tests.rs`. Consider a shared `storage_test_helpers.rs` if more helpers accumulate.

### Blocking Test Coverage Gap (HIGH confidence: 90%)

**File**: `crates/rskim-search/src/temporal/storage_ops.rs` and `storage_tests.rs`

**Missing test coverage for CapacityExceeded error path** — The `store_hotspots`, `store_risks`, `store_cochanges`, and `sync` methods all have capacity-guard branches (lines 101-106, 124-129, 145-150, 311-321) that return `SearchError::CapacityExceeded` when input exceeds 500,000 rows. None of these error branches have test coverage. These are documented safety rails, yet the actual exceeded behavior is never exercised.

**Fix**: Add at least one parameterized-style test:
```rust
#[test]
fn store_hotspots_rejects_over_capacity() {
    let (_dir, db) = temp_db();
    let rows: Vec<HotspotRow> = (0..=500_000)
        .map(|n| HotspotRow {
            file_path: format!("{}", n),
            score: 0.0,
            changes_30d: 0,
            changes_90d: 0,
        })
        .collect();
    let err = db.store_hotspots(&rows).unwrap_err();
    assert!(
        matches!(err, SearchError::CapacityExceeded(_)),
        "expected CapacityExceeded, got {:?}", err
    );
}

#[test]
fn sync_replaces_on_second_call() {
    let (_dir, db) = temp_db();
    db.sync(&[hotspot_a], &[risk_a], &[cochange_a], "sha1").unwrap();
    db.sync(&[hotspot_b], &[risk_b], &[cochange_b], "sha2").unwrap();
    assert_eq!(db.load_hotspots().unwrap().len(), 1);
    assert_eq!(db.load_hotspots().unwrap()[0].file_path, "b.rs");
}
```

**Confidence**: 90% (testing reviewer) — This closes the remaining untested branches in the storage layer.

---

**Review Summary**:
- **7 blocking findings** posted as inline comments (≥80% confidence)
- **1 blocking test coverage gap** (90% confidence, noted above)
- **4 lower-confidence suggestions** (60-79%, optional improvements)
- All duplicated findings across reviewers consolidated
- Rate limiting: Posted 7 comments sequentially with 2-second delays

**Next Steps**: Address the blocking findings to achieve full approval. The test coverage additions are especially important for documenting the capacity-guard contract.

---

*Review conducted by specialized agents: architecture, rust, reliability, testing, complexity, consistency, and dependencies. All findings ≥80% confidence posted as blocking items. Consolidated duplicates across reviewers.*

Issue #196, opened 2026-05-09
Wave 3e: Pre-defined pattern library (20-30 patterns)


dean0x:
## Overview
Define 20-30 common structural patterns as named AST n-gram queries.

### Error handling
- \`error-propagation\`: try_expression containing call_expression
- \`empty-catch\`: catch_clause with empty block
- \`unhandled-result\`: call_expression as expression_statement (result discarded)
- \`bare-unwrap\`: method call to unwrap() without context

### Performance
- \`n-plus-one\`: for_statement containing await_expression
- \`nested-loops\`: for inside for (depth > 1)
- \`blocking-in-async\`: synchronous I/O call inside async function

### Concurrency
- \`mutex-in-async\`: lock() call inside async function
- \`unbounded-spawn\`: spawn without join/collect

### Code quality
- \`deep-nesting\`: any node at depth > 5
- \`callback-hell\`: 3+ nested arrow_function/function_expression
- \`god-function\`: function with >20 statements in body
- \`magic-number\`: numeric literal in expression (not in const/assignment)

### Security
- \`sql-string-concat\`: string concatenation in query-like call
- \`hardcoded-secret\`: string literal matching secret patterns in assignment

### Structure
- \`empty-function\`: function with empty body
- \`single-use-variable\`: let + single reference + no further use
- \`excessive-params\`: function with >5 parameters
- \`return-in-finally\`: return statement inside finally block

## Storage
Data-driven: each pattern is a struct with name, description, and AST n-gram query spec. NOT hardcoded match logic.

## Dependencies
- #3 (foundation — can be built in parallel with #23)

## Acceptance Criteria
- [ ] At least 20 patterns defined
- [ ] Each has name, description, example
- [ ] Pattern definitions are data-driven
- [ ] Patterns work with old indexes (query-time only)

## Implementation Guidelines Checklist
- [ ] No file exceeds 400 lines
- [ ] Patterns are data-driven (no hardcoded match per pattern)
- [ ] Module has #[cfg(test)] mod tests

dean0x:
Completed and merged to `main` via PR #276 (squash `f56f506`).

[...]

Issue #273, opened 2026-06-04
AST index posting-list compression (delta + VarInt / Roaring) — reduce on-disk size


dean0x:
## Context
Follow-up from Wave 3d (#194, PR #272), part of the North Star 3-Layer Code Search (#174).

The AST structural n-gram on-disk index (`ast_index.skidx` + `ast_index.skpost`) currently stores **uncompressed** posting entries: `doc_id: u32 + count: u32` = 8 bytes each. Measured index size is ~**1.2× the source** on a representative corpus.

This is expected and fine by industry norms (Zoekt / Sourcegraph trigram indexes typically run ~3–5× source). The original #194 acceptance criterion "index < 5% of source" was **aspirational with no empirical basis** and is structurally impossible for AST n-grams — they have a tiny vocabulary (~161 distinct bi/trigrams) but each appears in nearly every file, producing dense posting lists that scale as O(vocab × files). PR #272 replaced that criterion with a realistic bound and deferred actual size reduction here.

## Why compression helps
The posting file dominates the index entirely (e.g. ~1.29 MB of 1.30 MB in the benchmark). Standard inverted-index compression applies directly:

| Technique | Posting cost | Est. ratio | Notes |
|---|---|---|---|
| Current: raw u32 doc_id + u32 count | 8 B | ~1.2× | shipped in #272 |
| Delta-encode sorted doc_ids + VarInt | ~1–2 B | ~0.3–0.5× | doc_ids already sorted ascending (contract C1) |
| Roaring bitmaps for dense lists | <1 B amortized | ~0.15–0.3× | best for the dense structural case |

Posting doc_ids are already stored sorted ascending with one entry per doc_id (PR #272 contract C1), so delta-encoding is a natural fit. The `count` field (term-frequency for BM25) can be VarInt-encoded alongside.

## Scope
- Add a compressed posting-list codec (start with delta + VarInt; evaluate Roaring for dense lists).
- Bump `AST_FORMAT_VERSION` (format-breaking change).
- Preserve the public `AstIndexReader` API (C1–C7) and the BM25 `count` semantics.
- Re-benchmark size ratio and query latency; document the achieved ratio (realistic target ~0.15–0.5×, NOT 5%).
- Decompress lazily per-lookup to keep the mmap read path fast.

## Acceptance
- [ ] Posting lists compressed; format version bumped.
- [ ] Public reader API + contracts C1–C7 unchanged.
- [ ] Measured index/source ratio documented (expected ~0.15–0.5×).
[...]

PR #276, opened 2026-06-06
feat(#196): Wave 3e — AST Pattern Library & Structural Index v2

dean0x:
## Summary

- Bumps the AST n-gram index to **format v2** (breaking change — re-index required) adding per-file structural metrics (max depth, max block statements, max params, branch count) and avg_max_depth in the header
- Adds **29 GOLD-verified structural code patterns** across 5 categories (ErrorHandling, Performance, Concurrency, Quality, Structure) with honest labeling (exact vs. approximate)
- Fixes two root-cause correctness bugs in `extract_ast_ngrams_with_metrics`: punctuation exclusion from counted-child counts, and complete close-depth loop covering all ancestor depths

### Index format v2 (`store/format.rs`, `builder.rs`, `reader.rs`)

- `FILE_META_SIZE` 5 → 15 bytes (+10 bytes/file for `max_depth:u16`, `max_block_stmts:u16`, `max_params:u16`, `branch_count:u32`)
- `FORMAT_VERSION` 1 → 2; v1 files rejected with clear "please rebuild" error
- Header `avg_max_depth:f32` at bytes [38..42] (previously reserved)
- New reader methods: `file_metrics()`, `avg_max_depth()`, `index_version()` (static probe)

### Structural metrics + synthetic n-gram markers (`structural.rs`, `extract.rs`)

- Synthetic parent IDs: `EMPTY_BODY=65000`, `DEEP_NODE=65001`, `LARGE_BODY=65002`, `MANY_PARAMS=65003` — `vocab_resolve()` returns `None` for all (isolation guarantee, per F5 test)
- `extract_ast_ngrams_with_metrics()` folds structural computation into the existing single-pass traversal — no second pass, no extra allocation
- **PUNCTUATION_KIND_IDS fix**: real vocabulary entries for `{`, `}`, `(`, `)`, `,`, `;`, keywords excluded from "counted child" counts, ensuring empty blocks are correctly detected in real parse output (not just hand-built synthetic test nodes)
- **Close-depth loop fix**: closes ALL depths from `prev_depth` down to `node.depth` (inclusive), so structural container nodes (e.g., `parameters`, `block`) are closed with their final `child_counts` before the slot is overwritten by the next sibling

### Pattern library (`patterns.rs`, `patterns_tests.rs`)

- 29 patterns, GOLD-verified (F7): `linearize_source(example)` + `extract_ast_ngrams_with_metrics` must emit every declared bigram/trigram — hard failure on vocabulary or structural mismatches
- Synthetic marker names (`__empty_body__`, `__deep_node__`, `__large_body__`, `__many_params__`, bucketed variants) resolve to reserved IDs via `resolve_synthetic_name()`
- Out-of-scope patterns documented honestly: `hardcoded-secret`, `magic-number`, `single-use-variable` require semantic/data-flow analysis unavailable from CST n-grams

## Breaking Changes

**AST index format v2** — all existing indexes written by Wave 3d (format v1) are rejected with "please rebuild". Run `skim search index --rebuild` to regenerate. No data loss; v2 is a strict superset.

## Reviewer Focus Areas

- `extract.rs` close-depth loop: the range `(close_start..=close_end).rev()` replaces the single `close_depth(close_end)` call — this is the fix that makes `parameters`, `block`, and other structural containers correctly emit MANY_PARAMS/LARGE_BODY/EMPTY_BODY markers
- `structural.rs` PUNCTUATION_KIND_IDS: covers all single-char and keyword tokens that appear as named nodes in CSTs; any gap here would cause false "empty" body detection
- `patterns.rs` F7 GOLD test verifies each `(bigram_str, bigram_str)` pair against actual parse output — no pattern in the catalog is fictional
- PF-004 compliance: saturating casts in `close_depth` (`min(u16::MAX)` before cast); PF-005: size-guard comment in `reader_tests.rs` documents measured ratio

### DIFF
--- a/crates/rskim-search/src/ast_index/extract.rs
+++ b/crates/rskim-search/src/ast_index/extract.rs
@@ -655,9 +655,7 @@ fn close_depth(
     if PARAM_LIST_KIND_IDS.contains(&kind_id) {
         // PF-004: saturating cast — count can be up to DEFAULT_MAX_NODES (100K) > u16::MAX
         let count_u16 = count.min(u32::from(u16::MAX)) as u16;
-        if count_u16 > metrics.max_params {
-            metrics.max_params = count_u16;
-        }
+        metrics.max_params = metrics.max_params.max(count_u16);
         emit_bucket_crossings(bigram_map, MANY_PARAMS, &PARAM_EDGES, count);
     }
 }
@@ -687,9 +685,7 @@ fn emit_empty_or_large_body(
     {
         // PF-004: saturating cast — count can be up to DEFAULT_MAX_NODES (100K) > u16::MAX
         let count_u16 = count.min(u32::from(u16::MAX)) as u16;
-        if count_u16 > metrics.max_block_stmts {
-            metrics.max_block_stmts = count_u16;
-        }
+        metrics.max_block_stmts = metrics.max_block_stmts.max(count_u16);
         emit_bucket_crossings(bigram_map, LARGE_BODY, &BODY_STMT_EDGES, count);
     }
 }
--- a/crates/rskim-search/src/ast_index/extract.rs
+++ b/crates/rskim-search/src/ast_index/extract.rs
@@ -417,7 +417,6 @@ pub fn extract_ast_ngrams_with_metrics(
                     if ancestors[depth_to_close].is_some() {
                         close_depth(
                             depth_to_close,
-                            &ancestors,
                             &child_counts,
                             &depth_kind,
                             &mut metrics,
@@ -489,7 +488,6 @@ pub fn extract_ast_ngrams_with_metrics(
             if ancestors[d].is_some() {
                 close_depth(
                     d,
-                    &ancestors,
                     &child_counts,
                     &depth_kind,
                     &mut metrics,
@@ -531,7 +529,6 @@ pub fn extract_ast_ngrams_with_metrics(
 /// was encountered, or because we're at end-of-stream).
 fn close_depth(
     depth_idx: usize,
-    ancestors: &[Option<NodeKindId>],
     child_counts: &[u32],
     depth_kind: &[NodeKindId],
     metrics: &mut StructuralMetrics,
@@ -600,9 +597,6 @@ fn close_depth(
         }
     }
 
-    // `ancestors` is not used in close_depth (only depth_kind is needed here),
-    // but it is kept in the signature for future use and to express the calling contract.
-    let _ = ancestors;
 }
 
 /// Extract structural n-grams using the production per-language IDF tables.
--- a/crates/rskim-search/src/ast_index/extract.rs
+++ b/crates/rskim-search/src/ast_index/extract.rs
@@ -27,6 +27,11 @@ use rskim_core::Language;
 
 use super::{AstBigram, AstTrigram, NodeKindId, ast_bigram_idf, ast_trigram_idf};
 use crate::ast_index::linearize::LinearNode;
+use crate::ast_index::structural::{
+    BODY_KIND_IDS, BODY_STMT_EDGES, BRANCH_KIND_IDS, DEEP_NODE, DEPTH_EDGES, EMPTY_BODY,
+    FUNCTION_KIND_IDS, LARGE_BODY, MANY_PARAMS, PARAM_EDGES, PARAM_LIST_KIND_IDS,
+    StructuralMetrics, bucket_label, is_counted_child,
+};
 
 // ============================================================================
 // Public types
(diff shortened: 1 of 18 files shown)


crates/rskim-search/src/ast_index/extract.rs
```diff
@@ -249,6 +254,346 @@ pub fn extract_ast_ngrams_with_weights(
...
+/// - `max_block_stmts` and `max_params` use `min(u16::MAX)` before cast.
+/// - `branch_count` saturates at `u32::MAX`.
+///
+/// # Returns
+///
+/// A pair `(AstNgramSet, StructuralMetrics)` where:
+/// - `AstNgramSet` contains all real n-grams plus synthetic marker bigrams.
+/// - `StructuralMetrics` contains per-file complexity metrics.
+#[must_use]
+pub fn extract_ast_ngrams_with_metrics(
```
dean0x, line 291: **BLOCKING: Function exceeds complexity threshold** (228 lines, ~9 interleaved concerns).

The function body (lines 291-518) folds all of these in one loop: max_depth update, cumulative depth-bucket emission, gap-fill nulling, parent child-count increment, subtree close detection, branch counting, real-bigram/trigram emission, and array bookkeeping.

Each concern is individually clear, but holding all nine state machines in mind — especially the close-depth timing (lines 398-427, requiring 5 explanatory comment blocks) — exceeds the maintainability threshold. By the project's Iron Law ('if you need comments to explain *what*, the code is too clever'), this signals the per-node loop body has crossed the line.

**Recommendation**: Extract cohesive loop-body blocks into inlinable helper methods on a shared mutable state struct (ExtractState with methods fill_depth_gap, close_open_subtrees, emit_real_ngrams) to preserve the single-pass guarantee and no-extra-allocation property while making each invariant testable in isolation.

Confidence: 92%

crates/rskim-search/src/ast_index/extract.rs
```diff
@@ -249,6 +254,346 @@ pub fn extract_ast_ngrams_with_weights(
...
+    if kind_id == 0 {
+        return; // sentinel — not a real construct, skip
+    }
+
+    let count = child_counts[depth_idx];
+
+    // ── EMPTY_BODY emission ───────────────────────────────────────────────────
+    // If this node is a body/block kind AND has zero counted children,
+    // emit EMPTY_BODY → enclosing_kind (the parent of this body kind).
+    if BODY_KIND_IDS.contains(&kind_id) {
```
dean0x, line 650: **HIGH: Function nests 5 levels deep** (above >4 CRITICAL threshold).

The `close_depth` body nests: if BODY_KIND_IDS → if depth_idx > 0 → if FUNCTION_KIND_IDS → for edge → if count >= edge.

Additionally, the EMPTY_BODY and LARGE_BODY branches both independently compute `depth_idx > 0` and re-read `enclosing_id = depth_kind[depth_idx - 1]` (lines 549 and 561) — duplicated guard + lookup that invites drift.

**Recommendation**: Compute `enclosing_id` once at the top via early return:
```rust
fn close_depth(...) {
    let kind_id = depth_kind[depth_idx];
    if kind_id == 0 { return; }
    let enclosing_id = (depth_idx > 0)
        .then(|| depth_kind[depth_idx - 1])
        .filter(|&id| id != 0);
    // ... then use enclosing_id in both branches
}
```

Then extract `emit_empty_or_large_body` helper. Removes one nesting level and the duplicated lookup.

Confidence: 88%

crates/rskim-search/src/ast_index/patterns_tests.rs
```diff
@@ -0,0 +1,401 @@
...
+                        resolve_kind_name(a)
+                            .zip(resolve_kind_name(b))
+                            .map(|(pa, pb)| AstBigram::encode(pa, pb) == bigram)
+                            .unwrap_or(false)
+                    })
+                    .map(|(a, _)| a)
+                    .unwrap_or(&"?"),
+                p.bigrams
+                    .iter()
+                    .find(|(a, b)| {
```
dean0x, line 157: **MEDIUM (Testing): F7 GOLD test silently passes patterns with zero resolved n-grams** (Confidence: 88%)

`f7_gold_all_patterns` iterates `resolved_bigrams()` / `resolved_trigrams()` and asserts each resolved n-gram is in the produced set. If a pattern's bigrams AND trigrams **all** fail to resolve (`vocab_lookup` returns `None` for every kind string), both vectors are empty, both `for` loops are no-ops, and the pattern passes the GOLD gate without verifying anything.

This is latent (all 30 patterns currently fully resolve), but a future vocab regeneration could silently disarm the honesty gate.

**Fix**: Add assertion that at least one declared n-gram resolved:
```rust
let resolved_total = resolved_bs.len() + resolved_ts.len();
assert!(
    resolved_total > 0,
    "GOLD VIOLATION for pattern '{}': NONE of its {} declared n-grams resolved...",
    p.name,
    p.bigrams.len() + p.trigrams.len()
);
```

crates/rskim-search/src/ast_index/store/reader_tests.rs
```diff
@@ -490,11 +529,12 @@ fn a14_overflow_header_huge_bigram_count() {
...
-// A16: index size ratio < 1.8× source bytes
-// Measured baseline: ~1.23×.  Bound leaves ~46% margin above baseline to catch
-// genuine O(files²) bloat while staying realistic.  Applies ADR-002 (grounded
-// regression guard, not an empirically-baseless 5% target).  On-disk
-// compression tracked in issue #273.
+// A16: index size ratio < 2.2× source bytes (raised from <1.8× in v2)
+// v1 measured baseline: ~1.23×.  v2 adds structural n-grams and +10 bytes/file
+// meta overhead; expected v2 ratio ~1.3×.  Guard raised to <2.2× per PF-005:
+// relaxation justified by measured ratio + structural capability expansion.
+// Applies ADR-002 (grounded regression guard).  On-disk compression tracked
```
dean0x: **MEDIUM (Documentation): Test comment cites wrong ADR** (Confidence: 95%)

This test comment (and line 654) cites ADR-002 ('grounded regression guard'), but ADR-002 is actually about progress trackers/memory files.

The correct citation is **ADR-003** ('Replace empirically-baseless acceptance criteria with grounded regression guards').

**Fix**: Replace `ADR-002` with `ADR-003` in both comments. The guard value (`< 2.2x` against measured ~1.3x) is correct and adheres to ADR-003.

dean0x:
## PR Review Summary: Wave 3e AST Pattern Library (#276)

**Date**: 2026-06-06 | **Reviewers**: 10 specialist agents | **Confidence Threshold**: >=80%

### Inline Comments Posted (6)

1. ✅ **extract.rs:291** — BLOCKING: 228-line function exceeds complexity threshold (92% confidence)
2. ✅ **extract.rs:545** — HIGH: 5-level nesting in close_depth (88% confidence)
3. ✅ **patterns_tests.rs:157** — MEDIUM (Testing): F7 GOLD silently passes zero-resolution patterns (88% confidence)
4. ✅ **reader_tests.rs:536** — MEDIUM (Documentation): ADR citation error, ADR-002 → ADR-003 (95% confidence)

### Critical Issues Not in Diff (Require Manual Fixes)

These findings are in unchanged lines and cannot receive inline comments, but **must be fixed before merge**:

#### 1. Stale Layout Diagrams (5 reviewers, consensus) — MEDIUM
**Files**: `crates/rskim-search/src/ast_index/store/format.rs:11` and `reader.rs:11`

**Current** (v1):
```
[AstFileMetaEntry × file_count: 5 bytes each]
```

**Should be** (v2):
```
[AstFileMetaEntry × file_count: 15 bytes each]
```

Matches `FILE_META_SIZE = 15` and all other docs in these files. Independently flagged by: architecture, performance, consistency, reliability, rust reviewers.

#### 2. Test Count Documentation Drift — MEDIUM (92% confidence)
**Files**: `CLAUDE.md` (lines ~16, ~608), `CHANGELOG.md`

Documented: **5,191** | Actual: **5,193** (off by 2 F9 probe tests added in recent commit)

**Update to: 5,193** in all three locations.

#### 3. doc_id Bounds Validation Missing — MEDIUM (Security, 85% confidence)
**File**: `crates/rskim-search/src/ast_index/store/reader.rs:455-471`

`lookup_postings_generic` does not validate `doc_id < file_count`. The sibling lexical reader does. A hostile CRC-valid index could embed invalid doc_ids that later cause wrong results when indexed into `file_meta(doc_id)`.

**Add before returning**:
```rust
if raw.doc_id >= self.header.file_count {
    return Err(SearchError::IndexCorrupted(...));
}
```

#### 4. Unchecked Multiply in file_meta Offset — MEDIUM (Security, 80% confidence)
**File**: `crates/rskim-search/src/ast_index/store/reader.rs:313`

Offset: `meta_start + (file_index as usize) * FILE_META_SIZE` uses unchecked math. Safe on 64-bit (all shipped targets), but 32-bit could wrap.

**Use checked arithmetic** to match PF-004 and `lookup_postings_generic` discipline.

#### 5. F7 GOLD Test Accepts Zero-Resolution Patterns — MEDIUM (Testing, 88% confidence)
**File**: `crates/rskim-search/src/ast_index/patterns_tests.rs:122-194`

If all declared n-gram names fail to resolve (`vocab_lookup` returns None), the test loops are no-ops and the pattern passes without verification. Latent but a vocab regen could silently disarm the honesty gate.

**Add assertion**:
```rust
let resolved_total = resolved_bs.len() + resolved_ts.len();
assert!(resolved_total > 0, 
    "GOLD VIOLATION for pattern '{}': NONE of its {} declared n-grams resolved...",
    p.name, ...);
```

#### 6. KNOWLEDGE.md Size-Ratio Contradiction — LOW (88% confidence)
**File**: `.devflow/features/ast-index/KNOWLEDGE.md:486`

Says `< 3.0×` but code uses `< 2.2×`.

**Update to: `< 2.2×`**.

These can be addressed separately or as follow-up issues:
- Synthetic-marker vocabulary split between modules (70%)
- Per-node DEEP_NODE re-emission cost in deep files (65%)
- Group 2 bench LazyLock warmup (62%)
- Sorted AstNgramSet construction (62%)
- `depth_kind` Vec mirrors `ancestors` (65%)

### Review Scores by Specialist

| Category | Score | Status | Rationale |
|----------|-------|--------|-----------|
| **Architecture** | 9/10 | APPROVED | No blocking defects; deferred-ranking boundary respected; v1→v2 breaking change handled cleanly |
| **Performance** | 9/10 | APPROVED | <15% overhead + <2.2× size empirically backed; single-pass no-extra-allocation verified |
| **Complexity** | 6/10 | CHANGES_REQUESTED | `extract_ast_ngrams_with_metrics` (228 lines, 9 concerns) and `close_depth` (5-level nesting) exceed thresholds |
| **Consistency** | 9/10 | APPROVED | PF-004 cast discipline upheld; codec mirrors v1/cochange; only cosmetic duplicates |
| **Regression** | 10/10 | APPROVED | v1 rejection bulletproof; extract fixes isolated to new code; no public exports removed; 680 tests pass |
| **Reliability** | 9/10 | APPROVED | PF-004 consistently applied; all loops bounded; decode panic-free; invariants asserted in production |
| **Rust** | 9/10 | APPROVED | Clippy-clean; PF-004-compliant; no panics on fallible paths; sound ownership |
| **Security** | 9/10 | APPROVED_WITH_CONDITIONS | Decode paths well-defended; two MEDIUM gaps (doc_id bounds, unchecked multiply) for hardening |
| **Testing** | 8/10 | APPROVED_WITH_CONDITIONS | Behavior-focused, deterministic, covers critical invariants; two MEDIUM gaps in F7 gate + punctuation regression lock |
| **Documentation** | 8/10 | APPROVED_WITH_CONDITIONS | CLAUDE.md AST INDEX NOTE accurate; test count drift (-2) + stale layout diagrams must be fixed |

### Recommendation: CHANGES_REQUESTED

**Status**: The codebase is correct, well-tested, and architecturally sound. All blocking/should-fix issues are addressable without major refactoring.

**Before merge, address**:
1. ✅ Complexity: extract_ast_ngrams_with_metrics (228 lines) and close_depth (5 nesting)
2. ✅ Test count documentation (5,191 → 5,193)
3. ✅ Stale layout diagrams (5 bytes → 15 bytes)
4. ✅ F7 GOLD zero-resolution gate
5. ✅ doc_id bounds validation
6. ✅ Unchecked multiply in file_meta
7. ✅ ADR-002 → ADR-003 citation

**Can follow**:
- PUNCTUATION_KIND_IDS cleanup (LOW)
- KNOWLEDGE.md size-ratio (LOW)
- Lowercase-confidence suggestions

---

**Summary**: Excellent specification, rigorous implementation, comprehensive tests. The complexity findings are legitimate but fixable without sacrificing the single-pass performance guarantee. All other categories range 8-10 on a 1-10 scale. Ready for revision cycle.

dean0x:
**MEDIUM: Module visibility deviation** — `pub mod patterns` vs. sibling `mod` pattern

**File**: `crates/rskim-search/src/ast_index/mod.rs:35`  
**Confidence**: 88%

All sibling submodules are declared `mod` (private) with curated `pub use` re-exports:
- `extract` → `pub use extract::{...}`
- `linearize` → `pub use linearize::{...}`  
- `store` → `pub use store::{...}`

Only `patterns` is declared `pub mod` (line 35) while *also* being re-exported via `pub use patterns::{...}` (line 62). This creates two access paths (`ast_index::patterns::X` AND `ast_index::X`) and leaks non-re-exported items to crate consumers.

**Fix**: Change line 35 to `mod patterns;` to match the "one module, one curated public surface" convention. The existing `pub use` already exports the intended API.

---
Claude Code | Review context: feature/196-wave-3e-pattern-library

dean0x:
**MEDIUM: Inconsistent max-value idiom after standardization**

**File**: `crates/rskim-search/src/ast_index/extract.rs:658-660` and `:690-692`  
**Confidence**: 85%

Commit 7f69091 (this PR) standardized `update_max_depth` to use the `.max()` idiom:
```rust
self.metrics.max_depth = self.metrics.max_depth.max(depth)
```

Two sibling updates in the same file still use the manual guard pattern:
- Line 658 (`close_depth`): `if count_u16 > metrics.max_params { metrics.max_params = count_u16; }`
- Line 690 (`emit_empty_or_large_body`): `if count_u16 > metrics.max_block_stmts { metrics.max_block_stmts = count_u16; }`

All three update the same `StructuralMetrics` struct with identical semantics.

**Fix**: Apply the same idiom for consistency:
```rust
metrics.max_params = metrics.max_params.max(count_u16);
metrics.max_block_stmts = metrics.max_block_stmts.max(count_u16);
```
Same behavior (PF-004 saturating cast preserved), cleaner intra-file consistency.

---
Claude Code | Review context: feature/196-wave-3e-pattern-library

dean0x:
**MEDIUM: PF-004 mis-citation in rustdoc comments**

**File**: `crates/rskim-search/src/ast_index/structural.rs` (doc comments on `StructuralMetrics` fields ~lines 135, 138)  
**Confidence**: 85%

The rustdoc comments for `max_block_stmts` and `max_params` state:
> Saturates at u16::MAX (PF-004: saturating cast, never wrapping)

However, PF-004 is actually about **"u16 depth arithmetic overflow: widen to u32 before adding an offset in depth comparisons"** (see `pitfalls.md`). It addresses the gap-fill `node.depth > p + 1` wrap in `extract.rs`, NOT saturating casts of metric fields.

This mis-citation is also in `.devflow/features/ast-index/KNOWLEDGE.md:508-509` (flagged independently by the documentation reviewer).

**Fix**: Either drop the PF-004 reference and state the rationale directly:
```rust
/// Saturates at u16::MAX (saturating cast, never wrapping)
```

Or, if a pitfall genuinely covers the saturating-cast/no-silent-narrowing rule, cite the correct ID. Ensure both `structural.rs` rustdoc and `KNOWLEDGE.md` describe PF-004 consistently with the authoritative pitfall text.

---
Claude Code | Review context: feature/196-wave-3e-pattern-library

dean0x:
**MEDIUM: Structural metrics never round-trip with non-zero values**

**File**: `crates/rskim-search/src/ast_index/store/builder_tests.rs` and `reader_tests.rs`  
**Confidence**: 90%

Every test in both files passes `StructuralMetrics::default()` (all zeros). The headline v2 feature — `AstIndexReader::file_metrics(i)` — surfaces per-file structural metrics, but this public API is never asserted to return non-zero values through the full build → serialize → mmap → read path.

The isolated codec test `file_meta_roundtrip` (format_tests.rs:380) exercises encode/decode in isolation with non-zero fields, but does NOT cover:
- Builder writing metrics to `self.file_meta` (builder.rs:296-302)
- Reader extracting them via `file_metrics` (reader.rs:344-351)

A bug that transposed metric fields (e.g. swapped `max_params` and `max_block_stmts` slots) would pass the codec test and all builder tests.

**Fix**: Add an end-to-end round-trip test:
```rust
#[test]
fn file_metrics_roundtrip_through_build() {
    let dir = tempdir().unwrap();
    let mut b = AstIndexBuilder::new(dir.path().to_path_buf()).unwrap();
    let m0 = StructuralMetrics { max_depth: 12, max_block_stmts: 25, max_params: 6, branch_count: 8 };
    let m1 = StructuralMetrics { max_depth: 3,  max_block_stmts: 1,  max_params: 0, branch_count: 0 };
    b.add_file_ngrams(FileId(0), Language::Rust,   &AstNgramSet::default(), 50, m0).unwrap();
    b.add_file_ngrams(FileId(1), Language::Python, &AstNgramSet::default(), 10, m1).unwrap();
    let r = b.build().unwrap();
    assert_eq!(r.file_metrics(0).unwrap(), m0);
    assert_eq!(r.file_metrics(1).unwrap(), m1);
}
```

---
Claude Code | Review context: feature/196-wave-3e-pattern-library

dean0x:
**MEDIUM: `avg_max_depth` computation untested with non-zero inputs**

**File**: `crates/rskim-search/src/ast_index/store/builder.rs:314-316` (accumulation) and `:446` (division)  
**Confidence**: 88%

The builder computes `avg_max_depth = total_max_depth / file_count`, but the only test (`a8_avg_max_depth_roundtrip` in format_tests.rs) injects the value directly into the header struct, bypassing the builder's computation entirely.

A regression that accumulated the wrong field or skipped accumulation would be invisible because all builder tests feed `max_depth == 0`, making the result always 0.0.

There IS a precedent in the suite: `avg_node_count_computed_correctly` (builder_tests.rs:376) asserts `(10+20+30)/3 == 20.0`. This gap needs the same treatment.

**Fix**: Add an equivalent test:
```rust
#[test]
fn avg_max_depth_computed_correctly() {
    let dir = tempdir().unwrap();
    let mut b = AstIndexBuilder::new(dir.path().to_path_buf()).unwrap();
    let m10 = StructuralMetrics { max_depth: 10, ..Default::default() };
    let m20 = StructuralMetrics { max_depth: 20, ..Default::default() };
    let m30 = StructuralMetrics { max_depth: 30, ..Default::default() };
    b.add_file_ngrams(FileId(0), Language::Rust,   &AstNgramSet::default(), 1, m10).unwrap();
    b.add_file_ngrams(FileId(1), Language::Python, &AstNgramSet::default(), 1, m20).unwrap();
    b.add_file_ngrams(FileId(2), Language::Java,   &AstNgramSet::default(), 1, m30).unwrap();
    let r = b.build().unwrap();
    assert!((r.avg_max_depth() - 20.0).abs() < 0.01);
}
```

---
Claude Code | Review context: feature/196-wave-3e-pattern-library

dean0x:
**MEDIUM: No parity test between metrics and legacy extraction paths**

**File**: `crates/rskim-search/src/ast_index/extract.rs` (two functions: line 291 `extract_ast_ngrams_with_metrics` vs. line 703 `extract_ast_ngrams`)  
**Confidence**: 82%

The builder switched from `extract_ast_ngrams` to `extract_ast_ngrams_with_metrics`, but the two functions re-implement the entire traversal (gap-fill, subtree-close, parent/grandparent containment, bigram/trigram emission) in parallel. They're intended to produce **identical real n-grams** (metrics path adds only synthetic markers on top), but no test asserts this equivalence directly.

A subtle divergence in edge cases (u16::MAX depth, sentinel-node handling, end-of-stream close ordering) would produce silently-different index contents.

**Fix**: Add a property test asserting real-ngram parity:
```rust
#[test]
fn extract_metrics_path_emits_same_real_ngrams_as_legacy() {
    let nodes = vec![
        node(0, 0),
        node(1, 1), node(1, 2),
        node(2, 3),
        node(1, 4),
        node(0, 5),
    ];
    let (with_metrics, _metrics) = extract_ast_ngrams_with_metrics(&nodes, Language::Rust);
    let baseline = extract_ast_ngrams(&nodes, Language::Rust);
    
    let real_from_metrics: Vec<_> = with_metrics.bigrams
        .iter()
        .filter(|e| e.ngram.parent() < 64900) // strip synthetic markers
        .collect();
    
    assert_eq!(real_from_metrics, baseline.bigrams);
    assert_eq!(with_metrics.trigrams.iter().filter(|e| e.ngram.parent() < 64900).collect::<Vec<_>>(),
               baseline.trigrams);
}
```

**Note**: This is deferred-OK for merge (no live indexes to break yet, Wave 3g not wired), but strongly recommended before the metrics index goes live.

---
Claude Code | Review context: feature/196-wave-3e-pattern-library

dean0x:
**MEDIUM: Missing u16::MAX regression guard on the new metrics extraction path**

**File**: `crates/rskim-search/src/ast_index/extract.rs:291-349` (`extract_ast_ngrams_with_metrics`)  
**Confidence**: 88%

The PF-004 depth-overflow regression test (B1, extract_tests.rs:523-580) exercises only `extract_ast_ngrams_with_weights`. The new `extract_ast_ngrams_with_metrics` path has its own independent gap-fill arithmetic and u32 widening guard (`ExtractState::fill_depth_gap` + line 314: `u32::from(node.depth) > u32::from(p) + 1`). The code is correct, but the invariant is unguarded by a test.

A future refactor that drops the u32 widening in the metrics path (reverting to `p + 1`) would wrap silently at `p == u16::MAX` and would NOT be caught by any existing test.

**Fix**: Add a sibling regression test that feeds a u16::MAX depth sentinel to the new function:
```rust
#[test]
fn extract_with_metrics_handles_u16_max_depth() {
    // From B1 pattern: depth jump at u16::MAX edge
    let nodes = vec![
        node(0, 0),
        node(1, u16::MAX - 1),
        node(2, u16::MAX),     // gap-fill must use u32 widening
    ];
    let (ngrams, metrics) = extract_ast_ngrams_with_metrics(&nodes, Language::Rust);
    assert_eq!(metrics.max_depth, u16::MAX);
    // No bigrams should be emitted for the gap (fill_depth_gap returns without emitting)
    assert!(ngrams.bigrams.is_empty()); // or assert only DEEP_NODE markers exist
}
```

This locks the PF-004 widening guard on the new code path against future drift.

---
Claude Code | Review context: feature/196-wave-3e-pattern-library

dean0x:
## Summary: Cycle 2 Code Review — 7 Inline Comments + Cross-Reviewer Consensus

### Review Context
All 10 reviewers returned **APPROVED** or **APPROVED_WITH_CONDITIONS** — zero blocking issues. The 7 comments above address MEDIUM should-fix items (85-90% confidence). This summary captures lower-confidence and consolidation notes.

---

### Cross-Reviewer Consensus: PUNCTUATION_KIND_IDS God-Set (82% confidence)

**5 independent reviewers** flagged `crates/rskim-search/src/ast_index/structural.rs:187-343` as a maintainability hazard:
- 150-entry set mixing true punctuation (`{`, `}`, `,`) with ~90 structural keywords (`fn`, `class`, `if`, `match`, etc.)
- No grouping invariant; adding a new keyword for a new grammar creates silent blast radius
- **Reviewers**: Complexity, Architecture, Consistency, Performance, Testing
- **Confidence**: 82%
- **Fix**: Split into `PUNCT_TOKENS`, `OPERATOR_TOKENS`, `STRUCTURAL_KEYWORDS` sub-constants; concatenate in the `LazyLock` builder. Documents intent, makes diffs reviewable.

---

### Triple-Source Bucket-Edge Contract (85% confidence)

**Architecture reviewer** flagged edge-table knowledge spread across 3 locations with no compile-time link:
1. `BODY_STMT_EDGES`, `PARAM_EDGES`, `DEPTH_EDGES` in `structural.rs:107-119`
2. Hardcoded suffixes (`__large_body_b20__`, `__many_params_b5__`) in `patterns.rs:104-123`
3. Pattern declarations using literal suffixes (e.g. `god-function` at `patterns.rs:493` uses `__large_body_b20__`)

If an edge table changes (e.g. `BODY_STMT_EDGES[1]` from 20 → 25), the names become lies silently — no error, only test honesty gates (F7 GOLD) catch language-specific cases.

**Fix**: Derive bucket names from edge tables, or add invariant test:
```rust
#[test]
fn synthetic_bucket_names_match_edge_tables() {
    for (i, &e) in BODY_STMT_EDGES.iter().enumerate() {
        assert_eq!(resolve_synthetic_name(&format!("__large_body_b{e}__")), Some(bucket_label(i)));
    }
    // repeat for PARAM_EDGES / DEPTH_EDGES
}
```

---

### Lower-Confidence Findings (60-79% confidence, deferred-OK or pre-existing)

| Finding | Confidence | Status | Reviewer |
|---------|------------|--------|----------|
| Extra pre-scan for `nodes.max()` at line 299 (O(n) before main loop) | 70% | Single-pass invariant preserved; document explicitly | Complexity |
| Per-node HashSet lookups in metrics path are the dominant cost | 82% | Bounded, not a defect; measure before optimizing | Performance |
| `emit_depth_buckets` runs `emit_bucket_crossings` on every node | 80% | Constant factor; `break` on first `value < edge` is micro-opt only | Performance |
| `resolve_synthetic_name` mirrors `structural.rs` constants (hand-encodes mapping) | 65% | Low risk (GOLD tests catch drift); cross-reference edge tables in comment | Rust |
| `pattern_to_query_set` doesn't dedup resolved n-grams (latent) | 70% | No current catalog collisions; document invariant or add `dedup_by_key` | Rust |
| Future-version index error message misleading (v3+ says "rebuild" not "upgrade") | 80% | Touched by diff (v1→v2 bump); distinguish `version < FORMAT_VERSION` vs `>` for v3 | Architecture |
| `max_block_stmts` and `max_params` metric *values* never asserted (F1 tests cover buckets, not scalars) | 82% | LOW priority; add scalar assertions to F1 test group | Testing |

---

### Pre-Existing Issues (Not Re-Raised)

Per cross-cycle awareness protocol, Cycle 1 deferred items were verified as landed correctly and NOT re-raised:
- ✓ Two-file rename atomicity / directory fsync (tracked as #273)
- ✓ README test count stale (handled by `release-prep.sh`)
- ✓ `serialize_index` (~137 lines, multi-concern) — pre-existing, mechanically extended
- ✓ `build` (~68 lines with inline assertions) — pre-existing, only `avg_max_depth` param added

---

### Test Results

- **All 10 reviewers**: APPROVED or APPROVED_WITH_CONDITIONS
- **Zero blocking issues**
- **Test suite**: 682 passed, 0 failed (verified across all review cycles)
- **Clippy**: clean (both `--lib` and `--all-targets`)
- **cargo fmt**: clean

---

### Recommendation

**READY TO MERGE** with optional post-merge follow-ups:
1. **Before Wave 3g**: Add metrics-path real-ngram parity test + u16::MAX regression guard (prevents future scoring regressions)
2. **Parallel track**: Split PUNCTUATION_KIND_IDS into semantic sub-constants (maintainability)
3. **Parallel track**: Fix PF-004 citations in rustdoc + KNOWLEDGE.md (decision reference accuracy)

All 7 inline comments provide code examples and exact file references for easy triage.

---
**Claude Code** | Cycle 2 review on feature/196-wave-3e-pattern-library

Issue #283, opened 2026-06-07
AST unigram posting index (single-node queries)


dean0x:
Follow-up to #197 (Wave 3f, AST Structural Pattern Query Engine). Sequenced **before #199** (Wave 3g CLI `--ast`).

## Context
Wave 3f's `parse_ast_query` already parses bare node kinds into `AstQuery::SingleNode(kind_id)` (validated via `vocab_lookup`), but execution returns a forward-referencing `InvalidQuery` error pointing at this ticket — single-node structural search needs a unigram posting table that the current index (format v2) does not store.

## Scope
- **Format v3**: add a third posting table — unigram `kind_id → [(doc_id, count)]` — alongside the existing bigram/trigram tables. Zero-migration (the AST index is not yet CLI-wired or deployed).
- **Extraction**: per-file node-kind counts (unigram frequencies) during AST n-gram extraction.
- **Builder**: write the unigram table.
- **Reader**: `AstIndexReader::lookup_unigram(NodeKindId) -> Result<Vec<AstPosting>>`.
- **Wire execution**: implement the `AstQuery::SingleNode` execution arm in `query.rs` (same OR-union BM25 path; unigram IDF table or 1.0 fallback).

## Why this is clean drop-in
From #197: the parser already yields `SingleNode(kind_id)` and the BM25 scoring path is generic over n-gram postings — **only the execution arm + the reader method + the v3 table are missing**. No rework of the query engine is needed.

## Acceptance
- Bare node-kind query (e.g. `try_statement`) returns files containing that node kind, BM25-ranked.
- `search_ast` postcondition holds (FileId-ascending, unique, scores > 0).
- Format v3 round-trips; v2 fixtures rejected with a "please rebuild" message.

Issue #199, opened 2026-05-09
Wave 3g: CLI integration (--ast "pattern")


dean0x:
## Overview
Add AST query support to the search CLI.

## CLI Surface
- \`skim search --ast "try_expression"\` — find files with try expressions
- \`skim search --ast error-propagation\` — use named pattern
- \`skim search --ast "for_statement > await_expression"\` — containment query
- \`skim search "auth" --ast "try_expression"\` — intersect text + AST

## Dependencies
- #25 (structural query engine)

## Acceptance Criteria
- [ ] Named patterns resolve
- [ ] Custom AST queries parse and execute
- [ ] Results show matched AST context
- [ ] Composes with text search via intersection
- [ ] --help documents --ast flag

## Implementation Guidelines Checklist
- [ ] No file exceeds 400 lines
- [ ] Every public function has /// doc comment
- [ ] All dependencies injected
- [ ] Module has #[cfg(test)] mod tests

dean0x:
**Wave 3f (#197) integration note:** The AST layer ships as a drop-in `SearchLayer` (reads only `query.ast_pattern`). CLI wiring is mechanical:
- Construct via `AstQueryEngine::open(dir)`.
- `parse_ast_query(s)` validates/echoes the query (single boundary parse; rejects > 4096 bytes; total over all inputs).
- `all_patterns()` (`patterns.rs`) backs a `--list-patterns` flag.
- **Dependency**: single-node queries (bare node kinds) require the unigram posting index follow-up ticket to land first; until then `parse_ast_query` parses them but execution returns a clear `InvalidQuery` pointing at that ticket.

Issue #198, opened 2026-05-09
Wave 4a: Multi-layer intersection logic


dean0x:
## Overview
Compose search results across all three layers via intersection and re-ranking.

## Algorithm
- Each layer returns Vec<(FileId, Score)>
- Lexical + AST → intersect (keep only files in both)
- Temporal → re-rank (multiply/adjust scores, don't filter)
- Empty intersection → empty results (not error)

## Dependencies
- #9 (lexical query), #16 (temporal storage), #25 (AST query)

## Acceptance Criteria
- [ ] text AND ast returns only files matching both
- [ ] Temporal re-ranking preserves result set, changes ordering
- [ ] Single-layer queries work (no mandatory intersection)

## Implementation Guidelines Checklist
- [ ] No file exceeds 400 lines
- [ ] Every public function has /// doc comment
- [ ] All dependencies injected
- [ ] All fallible operations return Result<T, SearchError>
- [ ] Module has #[cfg(test)] mod tests

---

**Wave 3g (#199) follow-up:**

#199 ships an **interim file-set filter**: the AST-matching FileId set is applied as \`file_filter\` on the lexical query (lexical ranking/snippets/line numbers preserved; AST acts as a pure filter, no composite re-ranking). #198 should formalize true intersection + composite cross-layer ranking.

dean0x:
**Wave 4 input from Wave 3e (PR #276): structural-complexity ranking dimension now available**

Wave 3e expanded the AST index to format v2, which now stores per-file `StructuralMetrics { max_depth, max_block_stmts, max_params, branch_count }` plus `avg_max_depth` in the header. These are exposed via `AstIndexReader::file_metrics(i)` and `AstIndexReader::avg_max_depth()` but are **not yet consumed by any ranker** — ranking integration was intentionally deferred to Wave 4.

Suggested next step for this issue: add a **structural-complexity dimension** to the compound re-ranker, normalizing `max_depth` and `branch_count` per file (e.g. against `avg_max_depth` / a corpus baseline) — mirroring how the temporal layer normalizes its hotspot dimension. This is a real, stored signal ready to rank on; no new indexing work is required, only the ranker wiring.

Note: the lint-style synthetic markers added in 3e (`empty-catch`, `deep-nesting`, `god-function`, etc.) are a **code-audit query capability**, inert for ordinary text queries — they are NOT a plain-query ranking signal. The ranking lever is the per-file metrics above.

dean0x:
[...]

Issue #200, opened 2026-05-09
Wave 4b: Composite ranking and scoring


dean0x:
## Overview
Combine BM25F, temporal, and AST scores into unified ranking. Extend with structural signals (import graph, directory proximity, structural coupling) to improve blast-radius precision beyond the Jaccard-only baseline established by the co-change validation benchmark (#191).

### Core (existing signals)
- **Lexical (BM25F):** text relevance from n-gram index
- **AST:** structural matching from AST-derived fields
- **Temporal (co-change):** Jaccard co-change coupling from git history

### Extended (new signals)
- **Import graph:** parse import/use/require statements via tree-sitter to identify direct dependencies and reverse-dependents. Files connected by imports are more likely to co-change than files that merely appear in the same commits by coincidence.
- **Directory proximity:** files in the same directory or nearby directories are more likely to be related. Simple, cheap signal that doesn't require parsing — path distance as a scoring factor.
- **Structural coupling:** shared type references, function calls, or trait implementations across files. Requires deeper AST analysis than import graph but captures coupling that imports alone miss (e.g., two files implementing the same trait).

### Configurable weights
- Default: lexical 0.5, AST 0.3, temporal 0.2
- Extended signals TBD after benchmarking — the co-change validation benchmark (#191) provides the infrastructure to measure each signal's marginal lift
- File matching all layers ranks above file matching 1
- Deterministic scoring

## Baseline from #191

The co-change validation benchmark (Jaccard-only, no other signals) measured:
- **Best threshold (macro F1):** 0.10 — F1 0.2819, P 0.2145, R 0.4111
- Precision at 21% means ~4/5 blast-radius predictions are false positives with co-change alone
- Signal quality varies by repo structure (modular repos like NestJS hit F1 0.57, monolithic repos ~0.21)
- Import graph and directory proximity are the most likely candidates to lift precision without heavy computation

## Dependencies
- #27 (intersection)
- #191 (co-change validation benchmark — provides evaluation infrastructure)

## Acceptance Criteria
- [ ] Weights are configurable
- [ ] Multi-layer matches rank higher
- [ ] Deterministic results
- [ ] Import graph signal implemented for at least TypeScript, Python, Rust, Go
- [ ] Directory proximity signal implemented
- [ ] Structural coupling signal implemented (may be phased separately)
- [ ] Each new signal measured against #191 benchmark corpus to verify marginal lift

## Implementation Guidelines Checklist
- [ ] No file exceeds 400 lines
[...]

PR #284, opened 2026-06-07
feat(#197): Wave 3f — AST structural pattern query engine

### LINKED ISSUE #197: Wave 3f: Structural pattern query engine
## Overview
Query engine that matches AST n-gram patterns against the index.

## Query types
- Named pattern: \`error-propagation\` → resolve to AST n-gram query
- Custom: \`"try_expression"\` → single node lookup
- Containment: \`"for_statement > await_expression"\` → depth-aware intersection

## Dependencies
- #23 (AST index)

## Acceptance Criteria
- [ ] Named patterns resolve correctly
- [ ] Containment queries filter by depth
- [ ] Performance: < 100ms for 10k-file index
- [ ] Unknown patterns return clear error

## Implementation Guidelines Checklist
- [ ] No file exceeds 400 lines
- [ ] Every public function has /// doc comment
- [ ] All dependencies injected
- [ ] Module has #[cfg(test)] mod tests

dean0x:
## Summary

Closes #197. Implements the BM25-ranked AST structural pattern query engine over the on-disk n-gram index built in Waves 3d/3e.

- **`parse_ast_query`** — total string→`AstQuery` boundary (never panics, rejects `>>` transitive, depth>2, >4 096 bytes, unknown kinds/patterns)
- **`AstQueryEngine`** with three query forms: named pattern (`empty-catch`, `empty-function`, `deep-nesting`, `god-function`), containment bigram/trigram (`A > B`, `A > B > C`), single-node (`function_declaration` — returns `InvalidQuery` pointing at #283 unigram-index work)
- **BM25 OR-union scoring**: `score = Σ idf · tf_norm / (tf_norm + k1)`, k1=1.2, b=0.75; IDF=1.0 fallback for synthetic n-gram markers and unknown language
- **`AstPostingSource` trait** — DI seam so tests run entirely in-memory (no disk I/O)
- **`search_ast()`** returns `Vec<(FileId, f64)>` sorted FileId-ASC — Wave-4 merge-join contract preserved
- **`SearchLayer::search`** adapter: filter chain (file_filter → lang → offset/limit), score-DESC sort, `None` query → empty, `Some("")` → `InvalidQuery`
- Double-counting from bigram+trigram overlap tolerated; deferred to #198

## Changes

| File | Change |
|---|---|
| `crates/rskim-search/src/ast_index/query.rs` | NEW — engine, 397 lines (≤400 cap) |
| `crates/rskim-search/src/ast_index/query_tests.rs` | NEW — 41 tests, FakePostingSource harness |
| `crates/rskim-search/benches/ast_query.rs` | NEW — 3 Criterion scenarios × 10k synthetic files |
| `crates/rskim-search/src/ast_index/mod.rs` | re-export new `query` module |
| `crates/rskim-search/src/lib.rs` | re-export new public types |
| `crates/rskim-search/Cargo.toml` | register `ast_query` bench |

## Breaking Changes

None. All additions are additive exports in `rskim-search`; no existing public APIs changed.

## Reviewer Focus Areas

- **BM25 scoring in `run_ngram_set`** (query.rs ~L120–180): verify length-norm formula, IDF-1.0 fallback for synthetic IDs, and the `cached_meta` check-then-insert pattern (can't use `or_insert_with` because `Result` propagation is needed)
- **SingleNode arm** (query.rs `search_ast`): must return `Err(SearchError::InvalidQuery(...))` with both "unigram" and "#283" in the message
- **`parse_ast_query` rejection paths** (query.rs ~L280–360): `>>` transitive syntax, depth>2, empty segments, 4 096-byte cap
- **Wave-4 contract**: `search_ast` returns FileId-ASC; `SearchLayer::search` applies score-DESC on top — confirm sort directions are not swapped
- **Test A5 (double-counting pinned)** in query_tests.rs: deliberately pinned to document the overlap, not to assert it's wrong; deferred to #198

## Downstream

- #283 — unigram index (enables `SingleNode` query form)
- #198 — deduplicate bigram+trigram overlap in OR-union scoring
- #199 — Wave 3g: `--ast` CLI flag wiring (`SearchLayer` is ready)

### DIFF
--- a/crates/rskim-search/src/ast_index/query.rs
+++ b/crates/rskim-search/src/ast_index/query.rs
@@ -28,7 +28,8 @@ pub const AST_BM25_K1: f64 = 1.2;
 /// BM25 length-normalisation parameter b for AST structural scoring.
 pub const AST_BM25_B: f64 = 0.75;
 /// Maximum allowed byte length for a raw query string (reliability bound).
-const MAX_AST_QUERY_BYTES: usize = 4096;
+/// Aliased from [`crate::lexical::MAX_QUERY_BYTES`] so both layers share one source of truth.
+const MAX_AST_QUERY_BYTES: usize = crate::lexical::MAX_QUERY_BYTES;
 
 /// Shared error message for empty query strings — used in both `SearchLayer::search`
 /// and `parse_ast_query` so the two sites cannot silently drift.
@@ -73,6 +74,11 @@ impl PartialEq for AstQuery {
 /// value. `node_count` values in `AstFileMetaEntry` MUST be non-negative.
 /// The production reader validates these at header/entry decode time; custom
 /// implementations must uphold the same contract.
+///
+/// **Count contract (C4).** `count` in every returned [`AstPosting`] MUST be
+/// `>= 1`. Sources returning `count == 0` break the all-scores-positive
+/// invariant (`count >= 1 → tf > 0 → score > 0`) relied on by BM25 and the
+/// `debug_assert!` in [`AstQueryEngine::run_ngram_set`].
 pub trait AstPostingSource: Send + Sync {
     /// Look up postings for an [`AstBigram`]; `Ok(vec![])` when absent (C2).
     fn lookup_bigram(&self, b: AstBigram) -> Result<Vec<AstPosting>>;
@@ -233,7 +239,23 @@ impl SearchLayer for AstQueryEngine<AstIndexReader> {
     /// `ast_pattern = Some(s)` → parse + execute; apply filters; return score-DESC results.
     /// Filters (in order): `file_filter` allowlist, `lang`, `offset`/`limit`.
     /// Defaults: `offset` → 0, `limit` → 20 (results truncated when unset).
+    ///
+    /// `bm25f_config` is intentionally ignored: the AST layer uses its own BM25
+    /// parameterisation ([`AST_BM25_K1`] / [`AST_BM25_B`]) and the lexical BM25F
+    /// config has no meaning here.
+    ///
+    /// # Errors
+    /// Returns [`SearchError::InvalidQuery`] when `temporal_flags` is set —
+    /// temporal sorting is not yet supported on the AST layer (deferred to Wave 4).
     fn search(&self, query: &SearchQuery) -> Result<Vec<SearchResult>> {
+        if query.temporal_flags.is_some() {
+            return Err(SearchError::InvalidQuery(
+                "temporal sorting (--hot / --cold / --risky) is not yet supported on the AST \
+                 layer; omit temporal flags or use the lexical search layer"
+                    .into(),
+            ));
+        }
+
         let raw = match &query.ast_pattern {
             None => return Ok(vec![]),
             Some(s) => s,
@@ -362,6 +384,7 @@ fn parse_bigram(a: &str, b: &str) -> Result<AstQuery> {
     Ok(AstQuery::Containment(AstNgramSet {
         bigrams: vec![AstBigramEntry {
             ngram: bigram,
+            // weight/count unused on query path; meaningful only at index build.
             weight: DEFAULT_AST_WEIGHT,
             count: 1,
         }],
@@ -375,6 +398,7 @@ fn parse_trigram(a: &str, b: &str, c: &str) -> Result<AstQuery> {
         bigrams: vec![],
         trigrams: vec![AstTrigramEntry {
             ngram: trigram,
+            // weight/count unused on query path; meaningful only at index build.
             weight: DEFAULT_AST_WEIGHT,
             count: 1,
         }],
@@ -390,6 +414,10 @@ fn kind(seg: &str) -> Result<NodeKindId> {
         ))
     })
 }
+
+/// IDF fallback for files whose language byte is unrecognised (neutral weight —
+/// does not amplify or suppress the BM25 term-frequency contribution).
+const UNKNOWN_LANG_IDF: f64 = 1.0;
 
 // Scoring helpers
 
@@ -412,7 +440,7 @@ fn score_postings<R: AstPostingSource>(
     // Per-n-gram IDF memoization: at most one distinct value per language.
     // Avoid HashMap overhead — track only the last seen (lang, idf) pair.
     let mut last_lang: Option<Language> = None;
-    let mut last_idf: f64 = 1.0;
+    let mut last_idf: f64 = UNKNOWN_LANG_IDF;
 
     for posting in postings {
         let meta = match meta_cache {
@@ -430,7 +458,7 @@ fn score_postings<R: AstPostingSource>(
                     v
                 }
             }
-            None => 1.0,
+            None => UNKNOWN_LANG_IDF,
         };
         *scores.entry(posting.doc_id).or_insert(0.0) += bm25_with_idf(posting, &meta, avg, idf);
     }
@@ -446,6 +474,10 @@ fn bm25_with_idf(posting: &AstPosting, meta: &AstFileMetaEntry, avg: f64, idf: f
     debug_assert!(
         avg.is_finite(),
         "avg_node_count must be finite (AstPostingSource contract)"
+    );
+    debug_assert!(
+        idf.is_finite() && idf > 0.0,
+        "idf must be finite and positive (language IDF contract)"
     );
     // Release-safe fallback: treat non-finite avg as 0 → length_norm=1.0.
     let avg = if avg.is_finite() { avg } else { 0.0 };
--- a/crates/rskim-search/src/ast_index/query.rs
+++ b/crates/rskim-search/src/ast_index/query.rs
@@ -22,7 +22,6 @@ use crate::{
 };
 
 // BM25 constants
-
 /// BM25 saturation parameter k1 for AST structural scoring.
 pub const AST_BM25_K1: f64 = 1.2;
 /// BM25 length-normalisation parameter b for AST structural scoring.
@@ -31,7 +30,6 @@ pub const AST_BM25_B: f64 = 0.75;
 const MAX_AST_QUERY_BYTES: usize = 4096;
 
 // Query enum
-
 /// A parsed, validated AST structural query.
 ///
 /// Created exclusively via [`parse_ast_query`] — the only `String → AstQuery`
@@ -58,7 +56,6 @@ impl PartialEq for AstQuery {
 }
 
 // DI seam
-
 /// Dependency-injection seam: implemented by [`AstIndexReader`] and test fakes.
 pub trait AstPostingSource: Send + Sync {
     /// Look up postings for an [`AstBigram`]; `Ok(vec![])` when absent (C2).
@@ -92,7 +89,6 @@ impl AstPostingSource for AstIndexReader {
 }
 
 // Query engine
-
 /// AST structural pattern query engine. Immutable; `&self`-only; `Send + Sync`.
 ///
 /// Use [`AstQueryEngine::new`] for DI (tests, Wave 4) or
@@ -136,10 +132,23 @@ impl<R: AstPostingSource> AstQueryEngine<R> {
         // Per-call meta cache: manual check-then-insert (or_insert_with can't propagate Result).
         let mut meta_cache: HashMap<u32, AstFileMetaEntry> = HashMap::new();
 
-        // AstNgramSet invariant: bigrams/trigrams are sorted-unique by key.
-        // debug_assert omitted here; enforced by the set construction site.
+        // Gap-fix #6: dedup by key (entries are sorted; O(n); prevents double-scoring dups).
+        let mut bigrams: Vec<&AstBigramEntry> = set.bigrams.iter().collect();
+        bigrams.dedup_by_key(|e| e.ngram.key());
+        debug_assert!({
+            bigrams
+                .windows(2)
+                .all(|w| w[0].ngram.key() != w[1].ngram.key())
+        });
+        let mut trigrams: Vec<&AstTrigramEntry> = set.trigrams.iter().collect();
+        trigrams.dedup_by_key(|e| e.ngram.key());
+        debug_assert!({
+            trigrams
+                .windows(2)
+                .all(|w| w[0].ngram.key() != w[1].ngram.key())
+        });
 
-        for entry in &set.bigrams {
+        for entry in bigrams {
             for posting in self.reader.lookup_bigram(entry.ngram)? {
                 let meta = cached_meta(&self.reader, &mut meta_cache, posting.doc_id)?;
                 // DEFERRED (Wave 4): minimal-covering-set to remove trigram/sub-bigram
@@ -149,7 +158,7 @@ impl<R: AstPostingSource> AstQueryEngine<R> {
                 });
             }
         }
-        for entry in &set.trigrams {
+        for entry in trigrams {
             for posting in self.reader.lookup_trigram(entry.ngram)? {
                 let meta = cached_meta(&self.reader, &mut meta_cache, posting.doc_id)?;
                 // DEFERRED (Wave 4): minimal-covering-set to remove trigram/sub-bigram
@@ -183,7 +192,6 @@ impl AstQueryEngine<AstIndexReader> {
 }
 
 // SearchLayer adapter (Wave 3g)
-
 impl SearchLayer for AstQueryEngine<AstIndexReader> {
     /// `ast_pattern = None` → `Ok(vec![])` (Wave-4 no-op).
     /// `ast_pattern = Some("")` → `Err(InvalidQuery("empty AST query"))`.
@@ -220,7 +228,7 @@ impl SearchLayer for AstQueryEngine<AstIndexReader> {
             hits
         };
 
-        // Sort score-DESC, FileId-ASC tie-break (NaN-safe; mirrors lexical reader.rs ~406).
+        // Sort score-DESC, FileId-ASC tie-break (NaN-safe; mirrors index/reader.rs ~406).
         filtered.sort_unstable_by(|a, b| {
             b.1.partial_cmp(&a.1)
                 .unwrap_or(Ordering::Equal)
@@ -248,7 +256,6 @@ impl SearchLayer for AstQueryEngine<AstIndexReader> {
 }
 
 // Parser
-
 /// Parse a raw string into an [`AstQuery`].
 ///
 /// **Only** `String → AstQuery` boundary; total (never panics). Rejects
@@ -349,7 +356,6 @@ fn kind(seg: &str) -> Result<NodeKindId> {
 }
 
 // Scoring helpers
-
 /// BM25 score contribution for one n-gram posting.
 ///
 /// IDF = 1.0 when `meta.language()` is `None` (unknown lang fallback).
--- a/crates/rskim-search/src/ast_index/query.rs
+++ b/crates/rskim-search/src/ast_index/query.rs
@@ -5,10 +5,9 @@
 //! Wave-3g [`SearchLayer`] adapter.
 
 use std::cmp::Ordering;
-use std::hash::BuildHasherDefault;
 use std::path::Path;
 
-use rustc_hash::{FxHashMap, FxHasher};
+use rustc_hash::FxHashMap;
 
 use rskim_core::Language;
 
@@ -149,10 +148,8 @@ impl<R: AstPostingSource> AstQueryEngine<R> {
 
         // Use FxHashMap (integer keys, trusted in-range doc_ids) with a capacity
         // hint to avoid rehashing on the hot insert path.
-        let mut scores: FxHashMap<u32, f64> = FxHashMap::with_capacity_and_hasher(
-            capacity,
-            BuildHasherDefault::<FxHasher>::default(),
-        );
+        let mut scores: FxHashMap<u32, f64> =
+            FxHashMap::with_capacity_and_hasher(capacity, Default::default());
 
         // Gap-fix #6: dedup by key (entries are sorted; O(n); prevents double-scoring dups).
         let mut bigrams: Vec<&AstBigramEntry> = set.bigrams.iter().collect();
@@ -178,7 +175,7 @@ impl<R: AstPostingSource> AstQueryEngine<R> {
         let mut meta_cache: Option<FxHashMap<u32, AstFileMetaEntry>> = if total_ngrams > 1 {
             Some(FxHashMap::with_capacity_and_hasher(
                 capacity,
-                BuildHasherDefault::<FxHasher>::default(),
+                Default::default(),
             ))
         } else {
             None
--- a/crates/rskim-search/src/ast_index/query.rs
+++ b/crates/rskim-search/src/ast_index/query.rs
@@ -370,9 +370,6 @@ fn bm25(
         if n <= 0.0 { 1.0 } else { n }
     };
     let tf_norm = tf / ln;
-    if tf_norm == 0.0 {
-        return 0.0;
-    }
     idf * (tf_norm / (tf_norm + AST_BM25_K1))
 }
 
--- a/crates/rskim-search/src/ast_index/query.rs
+++ b/crates/rskim-search/src/ast_index/query.rs
@@ -5,8 +5,10 @@
 //! Wave-3g [`SearchLayer`] adapter.
 
 use std::cmp::Ordering;
-use std::collections::HashMap;
+use std::hash::BuildHasherDefault;
 use std::path::Path;
+
+use rustc_hash::{FxHashMap, FxHasher};
 
 use rskim_core::Language;
 
@@ -28,6 +30,10 @@ pub const AST_BM25_K1: f64 = 1.2;
 pub const AST_BM25_B: f64 = 0.75;
 /// Maximum allowed byte length for a raw query string (reliability bound).
 const MAX_AST_QUERY_BYTES: usize = 4096;
+
+/// Shared error message for empty query strings — used in both `SearchLayer::search`
+/// and `parse_ast_query` so the two sites cannot silently drift.
+const EMPTY_QUERY_MSG: &str = "empty AST query";
 
 // Query enum
 /// A parsed, validated AST structural query.
@@ -57,6 +63,17 @@ impl PartialEq for AstQuery {
 
 // DI seam
 /// Dependency-injection seam: implemented by [`AstIndexReader`] and test fakes.
+///
+/// **Value-type coupling is intentional and bounded.** `AstPosting` and
+/// `AstFileMetaEntry` are deliberately treated as the stable query-layer value
+/// contract — they are `Copy`, gix-free, and mmap-free — so coupling to them
+/// is bounded and intentional, mirroring the "free of gix types" note on
+/// `CommitInfo` in `types.rs`.
+///
+/// **Finiteness contract.** `avg_node_count()` MUST return a finite, non-NaN
+/// value. `node_count` values in `AstFileMetaEntry` MUST be non-negative.
+/// The production reader validates these at header/entry decode time; custom
+/// implementations must uphold the same contract.
 pub trait AstPostingSource: Send + Sync {
     /// Look up postings for an [`AstBigram`]; `Ok(vec![])` when absent (C2).
     fn lookup_bigram(&self, b: AstBigram) -> Result<Vec<AstPosting>>;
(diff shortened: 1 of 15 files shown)


dean0x:
Test inline comment

crates/rskim-search/src/ast_index/query.rs
```diff
@@ -0,0 +1,397 @@
...
+                // double-counting (#198). For now, contributions are additive.
+                *scores.entry(posting.doc_id).or_insert(0.0) += bm25(posting, &meta, avg, |lang| {
+                    f64::from(ast_bigram_idf(lang, entry.ngram))
+                });
+            }
+        }
+        for entry in &set.trigrams {
+            for posting in self.reader.lookup_trigram(entry.ngram)? {
+                let meta = cached_meta(&self.reader, &mut meta_cache, posting.doc_id)?;
+                // DEFERRED (Wave 4): minimal-covering-set to remove trigram/sub-bigram
```
dean0x: ### Duplicated bigram/trigram accumulation loops create maintenance risk

**Issue**: Lines 151–160 and 161–170 are structurally identical except for n-gram type and lookup method.

**Fix**: Extract a generic accumulation helper to keep the loop body unified.

---
**Reviewers**: complexity (88%), performance (bigram/trigram duplication)
*Claude Code*

crates/rskim-search/src/ast_index/query.rs
```diff
@@ -0,0 +1,397 @@
...
+            for posting in self.reader.lookup_bigram(entry.ngram)? {
+                let meta = cached_meta(&self.reader, &mut meta_cache, posting.doc_id)?;
+                // DEFERRED (Wave 4): minimal-covering-set to remove trigram/sub-bigram
+                // double-counting (#198). For now, contributions are additive.
+                *scores.entry(posting.doc_id).or_insert(0.0) += bm25(posting, &meta, avg, |lang| {
+                    f64::from(ast_bigram_idf(lang, entry.ngram))
+                });
+            }
+        }
+        for entry in &set.trigrams {
```
dean0x: ### Postings fully materialized into Vec instead of streamed over mmap

**Issue**: `lookup_bigram()` and `lookup_trigram()` return `Vec<AstPosting>`, forcing full materialization and heap allocation for every posting list. Postings are uncompressed, fixed-stride, and mmap-backed — an ideal candidate for a borrowing streaming iterator with zero allocation.

**Impact**: On the all-match bigram scenario (10k postings), this is ~80KB of transient allocation plus a full decode pass that duplicates bytes already in mmap. On multi-n-gram patterns this multiplies.

**Fix**: Add a borrowing API to `AstPostingSource` that yields postings without allocating:

```rust
fn for_each_bigram_posting(
    &self,
    b: AstBigram,
    f: impl FnMut(AstPosting) -> Result<()>
) -> Result<()>
```

Alternatively, an iterator over the validated mmap slice. The DI seam is brand new, so changing its shape now is cheap; post-Wave-4 this becomes harder.

---
**Reviewers**: performance (88% confidence on Vec materialization)
*Claude Code*

crates/rskim-search/src/ast_index/query.rs
```diff
@@ -0,0 +1,397 @@
...
+            AstQuery::SingleNode(_) => Err(SearchError::InvalidQuery(
+                "single-node structural search requires the unigram index — tracked in #283".into(),
+            )),
+            AstQuery::Pattern(pattern) => {
+                let set = crate::ast_index::pattern_to_query_set(pattern);
+                self.run_ngram_set(&set)
+            }
+            AstQuery::Containment(set) => self.run_ngram_set(set),
+        }
+    }
```
dean0x, line 148: ### `scores` HashMap uses SipHash + no capacity hint — rehash churn on hot path

**Issue**: `let mut scores: HashMap<u32, f64> = HashMap::new()` uses std's default SipHash hasher (DoS-resistant but slow for integer keys) and starts empty. An all-match bigram inserting 10k distinct `doc_id` keys triggers ~log2(10000/initial) rehashes. Keys are trusted internal `u32` doc_ids (validated at reader.rs), not adversarial input, so SipHash provides zero security benefit here.

**Impact**: On the hot path, SipHash + repeated rehashing is a measurable constant factor on O(postings) insertion. With 10k entries, this is the second-largest avoidable cost after the bigram/trigram loop duplication.

**Fix**: Use a fast integer hasher and pre-size:

```rust
use rustc_hash::FxHashMap;

let mut scores = FxHashMap::with_capacity_and_hasher(
    self.reader.file_count() as usize,
    Default::default()
);
```

Same optimization applies to `meta_cache` on line 133.

---
**Reviewers**: performance (85% confidence)
*Claude Code*

crates/rskim-search/src/ast_index/query.rs
```diff
@@ -0,0 +1,397 @@
...
+            for (fid, score) in hits {
+                if self.reader.file_meta(fid.0)?.language() == Some(lang) {
+                    out.push((fid, score));
+                }
+            }
+            out
+        } else {
+            hits
+        };
+
```
dean0x, line 287: ### SearchLayer lang filter re-reads `file_meta` already decoded during scoring

**Issue**: `run_ngram_set` loads file metadata for every scored doc into a per-call `meta_cache`, then drops it. The lang filter then calls `self.reader.file_meta(fid.0)?` again on the surviving candidates. For an all-match bigram, this re-reads metadata 10k times that was just discarded.

**Impact**: Doubles `file_meta` decode count on filtered queries. Additionally, filtering runs on the full candidate set before `offset`/`limit`, paying for files that will be paginated away.

**Fix**: Either (a) carry language alongside the score out of `run_ngram_set`, (b) reuse a shared meta cache across scoring and filtering, or (c) short-circuit when the corpus is mono-lingual. At minimum, move filtering after `skip`/`take`.

---
**Reviewers**: performance (80% confidence)
*Claude Code*

crates/rskim-search/src/ast_index/query.rs
```diff
@@ -0,0 +1,397 @@
...
+    fn lookup_trigram(&self, t: AstTrigram) -> Result<Vec<AstPosting>>;
+    /// Per-file metadata for `doc_id`; `Err(IndexCorrupted)` when out of range.
+    fn file_meta(&self, doc_id: u32) -> Result<AstFileMetaEntry>;
+    /// Average per-file node count across the corpus.
+    fn avg_node_count(&self) -> f32;
+    /// Total number of files in the index.
+    fn file_count(&self) -> u32;
+}
+
+impl AstPostingSource for AstIndexReader {
```
dean0x, line 94: ### Trait method delegation relies on implicit inherent-method precedence

**Issue**: `impl AstPostingSource for AstIndexReader` forwards trait methods to identically-named inherent methods (e.g., `fn lookup_bigram(&self, b) { self.lookup_bigram(b) }`). Rust method resolution prefers inherent methods over trait methods for a concrete receiver, so no infinite recursion *today*. However, this is silent and fragile — if this impl were refactored to use a generic bound or `&dyn AstPostingSource`, the same syntax would stack-overflow at runtime with no compiler error.

**Fix**: Make delegation explicit using fully-qualified syntax:

```rust
fn lookup_bigram(&self, b: AstBigram) -> Result<Vec<AstPosting>> {
    AstIndexReader::lookup_bigram(self, b)
}
```

Apply to all five forwarders. This is self-documenting and refactor-proof. Per the crate's "magic behavior / implicit dependencies" quality gate, worth doing now.

---
**Reviewers**: rust (88% confidence)
*Claude Code*

crates/rskim-search/src/ast_index/query.rs
```diff
@@ -0,0 +1,397 @@
...
+    /// `ast_pattern = None` → `Ok(vec![])` (Wave-4 no-op).
+    /// `ast_pattern = Some("")` → `Err(InvalidQuery("empty AST query"))`.
+    /// `ast_pattern = Some(s)` → parse + execute; apply filters; return score-DESC results.
+    /// Filters (in order): `file_filter` allowlist, `lang`, `offset`/`limit`.
+    fn search(&self, query: &SearchQuery) -> Result<Vec<SearchResult>> {
+        let raw = match &query.ast_pattern {
+            None => return Ok(vec![]),
+            Some(s) => s,
+        };
+        let trimmed = raw.trim();
```
dean0x, line 262: ### `SearchLayer::search` doc omits default offset/limit behavior

**Issue**: The impl-level doc states the filter order (file_filter, lang, offset/limit) but does not document the defaults when those fields are `None`. The implementation defaults to `offset=0` and `limit=20` (lines 240–241), but a caller reading only the doc cannot know that unset `limit` silently truncates to 20 results.

**Impact**: Behaviorally significant and easy-to-miss. This is the primary public entry point for CLI (#199).

**Fix**: Add one line to the doc block:

```rust
/// Defaults: `offset` → 0, `limit` → 20 (results truncated when unset).
```

---
**Reviewers**: documentation (85% confidence)
*Claude Code*

crates/rskim-search/src/ast_index/query_tests.rs
```diff
@@ -0,0 +1,1146 @@
...
+#[test]
+fn b4_file_filter_restricts_file_ids() {
+    let (dir, reader) = build_small_index();
+    let engine = AstQueryEngine::new(reader);
+
+    let mut query = SearchQuery::new("q");
+    query.ast_pattern = Some("function_item > block".into());
+    query.file_filter = Some(HashSet::from([FileId(0)]));
+    let results = engine.search(&query).unwrap();
+
```
dean0x, line 1197: ### `b4_limit_and_offset_honored` validates count but not the returned files

**Issue**: The test sets `limit=2, offset=1` over 5 files with descending scores `[f0, f1, f2, f3, f4]` and asserts only `results.len() == 2`. It never asserts *which* files are returned. A bug that ignored `offset` or mis-ordered would still pass. Per the review scope, this is the key SearchLayer offset/limit boundary test.

**Fix**: Assert file identities and order:

```rust
assert_eq!(results.iter().map(|r| r.file_id).collect::<Vec<_>>(), 
           vec![FileId(1), FileId(2)]);
```

Add a sibling case for `offset` beyond the end (e.g., `offset=10 → len()==0`) to pin the skip/take boundary.

---
**Reviewers**: testing (92% confidence)
*Claude Code*

crates/rskim-search/src/ast_index/query.rs
```diff
@@ -0,0 +1,397 @@
...
+    let ln = if avg <= 0.0 {
+        1.0
+    } else {
+        let n = 1.0 - AST_BM25_B + AST_BM25_B * (nc / avg);
+        if n <= 0.0 { 1.0 } else { n }
+    };
+    let tf_norm = tf / ln;
+    if tf_norm == 0.0 {
+        return 0.0;
+    }
```
dean0x: ### `bm25` does not assert float-input preconditions, allowing NaN to propagate from non-conforming `AstPostingSource`

**Issue**: `bm25` accepts `avg: f64` and `meta.node_count` from any `AstPostingSource` implementation (DI seam is `pub`). The production `AstIndexReader` validates `avg_node_count` is finite at header decode, but the trait contract documents no such guarantee. If a future alternate reader or test fake returns `avg = NaN`, the guard `if avg <= 0.0` evaluates `false` for NaN, so execution falls through to `nc / avg`, propagating NaN into the score.

**Impact**: Contained — NaN scores are silently dropped by the `> 0.0` filter (line 174), so no panic. But silent omission violates the crate's "fail loud, never silent" principle. The invariant is undocumented at the trait boundary.

**Fix**: Add a finiteness guard or assert at the top of `bm25` (per the crate's "assert preconditions in production code" principle):

```rust
fn bm25(..., avg: f64, ...) -> f64 {
    debug_assert!(avg.is_finite(), "avg_node_count must be finite");
    let avg = if avg.is_finite() { avg } else { 0.0 }; // NaN/Inf → fallback
    // ... rest of function ...
}
```

Alternatively, document the finiteness contract on the `AstPostingSource` trait.

---
**Reviewers**: reliability (82% confidence — medium severity)
*Claude Code*

crates/rskim-search/src/ast_index/query.rs
```diff
@@ -0,0 +1,397 @@
...
+}
+
+// DI seam
+
+/// Dependency-injection seam: implemented by [`AstIndexReader`] and test fakes.
+pub trait AstPostingSource: Send + Sync {
+    /// Look up postings for an [`AstBigram`]; `Ok(vec![])` when absent (C2).
+    fn lookup_bigram(&self, b: AstBigram) -> Result<Vec<AstPosting>>;
+    /// Look up postings for an [`AstTrigram`]; `Ok(vec![])` when absent (C2).
```
dean0x, line 84: ### `AstPostingSource` trait exposes concrete storage types — mild DIP violation

**Issue**: The DI seam abstracts *where postings come from* but its method signatures are expressed entirely in concrete on-disk types (`AstBigram`, `AstTrigram`, `AstPosting`, `AstFileMetaEntry`). Any change to the storage layer's posting/meta shape forces a change to this query-layer trait and every implementor.

**Impact**: Mild — these types are simple `Copy` value types, not gix/mmap handles, so infrastructure detail doesn't leak. The blast radius of a storage change is bounded to re-deriving the value types. This is arguably pragmatic given `AstPosting` is trivial `{doc_id, count}`.

**Suggestion**: Add a doc line clarifying that `AstPosting`/`AstFileMetaEntry` are deliberately the stable query-layer value contract (they are `Copy`, gix-free, mmap-free), so coupling to them is intentional and bounded. Mirrors the existing "gix-free at the boundary" note on `CommitInfo` (types.rs:237–240). Optional: if a Wave-4 in-memory posting source needs a different meta shape later, introduce a query-owned `PostingMeta` projection.

---
**Reviewers**: architecture (80% confidence)
*Claude Code*

crates/rskim-search/src/ast_index/query.rs
```diff
@@ -0,0 +1,397 @@
...
+            .skip(query.offset.unwrap_or(0))
+            .take(query.limit.unwrap_or(20))
+            .map(|(fid, score)| SearchResult {
+                file_id: fid,
+                score,
+                line_range: 0..0,
+                match_positions: vec![],
+                field: SearchField::Other,
+                snippet: None,
+            })
```
dean0x, line 306: ### `limit.unwrap_or(20)` magic default duplicated across search layers

**Issue**: The default page size `20` is hardcoded here and again in the lexical layer (`index/reader.rs:413`). Two independent `SearchLayer` implementations encode the same pagination policy. If the project changes the default page size, both must change in lockstep — a DRY / single-source-of-truth gap.

**Impact**: Minor. The duplication is intentional consistency today (AST layer correctly mirrors the established lexical contract). Risk is only future drift if one site is changed and the other isn't.

**Fix** (optional): Hoist `const DEFAULT_SEARCH_LIMIT: usize = 20;` into `types.rs` near `SearchQuery` and reference it from both layers. Low priority; flag only if a third SearchLayer is coming.

---
**Reviewers**: architecture (85% confidence)
*Claude Code*

dean0x:
## Summary: Lower-Confidence Findings (60-79%)

The following suggestions fall below the 80% confidence threshold for inline comments but warrant noting:

### Performance & Testing

1. **Benchmark gaps mask IDF redundancy cost** (performance, 70% confidence) — Current benches use single language (Rust), so the per-posting IDF redundancy (HIGH finding) and lang-filter re-read costs are masked. Recommend adding a mixed-language variant of bench_hot_bigram and a multi-n-gram pattern scenario so optimizations can be validated before/after.

2. **meta_cache value depends on workload** (performance, 80% confidence) — Single-n-gram containment queries (the common case) have zero cache hits, so the overhead may exceed the benefit. Measure both with and without the cache before committing to the current design.

3. **Dead-code filter on final scores** (performance, 82% confidence) — filter(|(_, s)| *s > 0.0) is unreachable because BM25 is strictly positive for any real posting. Either drop it or add a comment documenting why it's defensive-only.

### Testing Details

4. **length_norm guard branches are untested** (testing, 88% confidence) — The avg <= 0.0 → 1.0 path and the defensive n <= 0.0 branch have no test coverage. Recommend one test with avg_node_count == 0.0 and a real posting, and a comment on the unreachable branch explaining it's defensive.

5. **Malformed-index propagation only tested via fake** (testing, 80% confidence) — b3_corrupt_doc_id_propagated tests the engine's error handling, but the real reader's file_meta OOB path (reader.rs C7) is not end-to-end tested. Verify coverage lives in reader tests; if not, add an integration test.

6. **Lang/file filters never assert the negative** (testing, 82% confidence) — all() on an empty iterator is true, so over-pruning filters would still pass. Add assert!(!results.is_empty()) plus explicit count to ensure filters don't accidentally drop everything.

### Complexity & Consistency

7. **Parser helper naming inconsistency** (consistency, 88% confidence) — parse_single has no suffix while parse_bigram_q and parse_trigram_q have _q. Drop the suffix for symmetry (parse_single, parse_bigram, parse_trigram) — these are private and the parent function establishes the "query parsing" context.

8. **Duplicated empty-query guard** (consistency, 82% confidence) — Both SearchLayer::search and parse_ast_query emit InvalidQuery("empty AST query") with the identical message string. Factor into a shared const to prevent drift.

9. **Hardcoded line-reference in comment will rot** (consistency, 70% confidence) — "mirrors index/reader.rs ~406" points to a line number that will drift as files change. Reference the function/symbol name instead (e.g., "mirrors index/reader.rs BM25 sort pattern").

---

No blocking issues found in these lower-confidence items. They represent polish and optimization opportunities; the PR is architecturally sound.

*Claude Code*

dean0x:
### rustc-hash: Use workspace table + clarify version choice

**Issue**: `rustc-hash = "1.1.0"` is the only dependency in this crate declared with a literal version. Every other dependency uses `{ workspace = true }` (thiserror, serde, tree-sitter, memmap2, crc32fast, tempfile, gix, regex, rusqlite, rayon). This breaks the established consistency pattern.

**Confidence**: 98% (blocking per consistency, dependencies, rust reviews)

**Impact**: Version drift risk. If another crate later adds rustc-hash, the two can diverge silently — exactly what the workspace table prevents.

**Fix**: 
1. Move to `/Cargo.toml [workspace.dependencies]`:
```toml
rustc-hash = "1"  # or "1.1" — unifies with existing transitives (hashlink, tiktoken-rs)
```

2. Update `crates/rskim-search/Cargo.toml` line 29:
```toml
rustc-hash = { workspace = true }
```

**Note on version**: The current 1.1.0 is defensible — it unifies with existing transitive copies (hashlink ^1.1, tiktoken-rs ^1.1.0) to avoid two major versions in the binary. Consider adding a comment in the workspace table explaining this choice so future dependency bumps don't blindly swap to 2.x.

---
*Consolidated finding from: consistency (98%), dependencies (95%), rust (70%)*  
*Claude Code*

dean0x:
### Benchmark name inconsistency: add `_bench` suffix

**File**: `crates/rskim-search/Cargo.toml` (~line 44)  
**Confidence**: 90% (consistency medium)

**Issue**: The crate's two existing benches are `linearize_bench` and `ast_index_bench` (both use the `_bench` suffix). The new bench is `ast_query` — no suffix. This breaks the 2-of-2 established pattern within the same Cargo.toml.

**Impact**: Inconsistent `cargo bench --bench` invocation surface (`--bench ast_query` vs `--bench ast_index_bench`). Minor, but visible in the file the PR edits directly.

**Fix**: Rename to match siblings:
```toml
[[bench]]
name = "ast_query_bench"
harness = false
```

Also rename the file `benches/ast_query.rs` → `benches/ast_query_bench.rs` and update the header's "Run with" line if present.

---
*Consolidated finding from: consistency (90%)*  
*Claude Code*

dean0x:
### Document C4 `count >= 1` contract on AstPostingSource trait

**File**: `crates/rskim-search/src/ast_index/query.rs` (lines 76-87, trait definition)  
**Confidence**: 85% (reliability medium)

**Issue**: The `debug_assert!` at line 214 and the `bm25_with_idf` positivity invariant depend on "C4: count >= 1 → tf > 0 → score > 0". The production `AstIndexReader` enforces this (decode_posting rejects count == 0 at format.rs:472-476), but the `AstPostingSource` trait doc omits this contract. A custom source violating it would trip the debug assertion silently.

**Impact**: Undocumented invariant in the DI seam. Per devflow reliability rule 2 (assert/document invariants in production code), this should be stated where it's enforced on the trait.

**Fix**: Add to the trait doc comment (update lines 72-75):

```rust
/// Finiteness and range contract:
/// - `avg_node_count()` MUST be finite and non-negative (C1)
/// - `node_count` in every returned `AstFileMetaEntry` MUST be non-negative
/// - `count` in every returned `AstPosting` MUST be `>= 1` (C4) — sources returning `count == 0` 
///   break the all-scores-positive invariant and will trip the debug_assert at search_ast line 214.
```

No code change required; the production reader already enforces this.

---
*Consolidated finding from: reliability (85%)*  
*Claude Code*

dean0x:
### Lang filter re-reads `file_meta` after scoring — double-decode cost

**File**: `crates/rskim-search/src/ast_index/query.rs` (lines 255-262, `SearchLayer::search` adapter)  
**Confidence**: 81% (performance medium), 80% (architecture medium)

**Issue**: `run_ngram_set` decodes `file_meta` for every scored doc during BM25 calculation. The result drops the decoded language info. Then the lang filter (lines 255-262) calls `self.reader.file_meta(fid.0)?` again on surviving candidates, re-decoding the exact bytes just read. For a hot bigram with a lang filter, this doubles the dominant hot-path cost.

**Impact**: For filtered queries (the realistic interactive path `skim search 'A > B' --lang rust`), this is up to 20k file_meta decodes (10k in scoring + 10k in filtering) for a result capped at limit (default 20). The performance review measures this as the second-largest avoidable cost after the IDF cache thrash.

**Fix**: Thread the language from the scoring loop into the filter. Recommend option 2: push the lang filter *inside* `score_postings` where `meta` is already decoded (it's already read for IDF at line 422). Pass `lang_filter: Option<Language>` into `score_postings` to skip non-matching postings before they enter the scores map. This also composes with allocation pessimism (fewer entries in the scores map).

Alternatively: (a) return `Vec<(FileId, f64, Language)>` from `run_ngram_set`, or (b) apply filter after `skip`/`take` to avoid decoding results that will be paginated away.

---
*Consolidated finding from: performance (81%), architecture (80%)*  
*Claude Code*

dean0x:
### Update line-count claim: file is 482 lines, not ~397

**File**: `crates/rskim-search/src/ast_index/query.rs` (module header and PR description)  
**Confidence**: 95% (complexity medium)

**Issue**: The PR description and module notes state "query.rs ~397 lines under a 400-line cap," but the actual file is 482 lines. Even excluding the 3-line test-module include (lines 480-482), the file is 479 lines. The 400-line cap is therefore not met. This is not a complexity defect (the executable-code portion is ~331 lines and cleanly organized), but the stated invariant is false and will mislead future maintainers.

**Impact**: A self-imposed size budget that is silently exceeded stops functioning as a guardrail for code review clarity.

**Fix**: Reconcile the claim. Recommended approach: update the PR/module note to state the real metric, e.g.:

```
"479 total lines (≈331 code, ≈109 doc, ≈39 blank); doc-heavy by design to explain the BM25 scoring contract."
```

If a hard 400-line total cap is intended, the natural split is to move the parser section (parse_ast_query + parse_single/parse_bigram/parse_trigram/kind, lines 294-392, ~99 lines) into a sibling `query_parse.rs`, which also cleanly separates the `String -> AstQuery` boundary from the execution engine. This is not required for merge, only for honest documentation.

---
*Consolidated finding from: complexity (95%)*  
*Claude Code*

dean0x:
## Summary: PR #284 Review Findings

**Status**: 5 inline comments created for >=80% confidence findings. This summary consolidates lower-confidence items and provides full context.

### Deduplication Across Reviewers (Consolidated)

The review team identified several overlapping concerns. Here's how they were consolidated:

| Issue | Reviewers | Confidence | Status |
|-------|-----------|-----------|--------|
| `rustc-hash` placement (literal vs workspace table) | consistency, dependencies, rust | 98%/95%/70% | **Inline comment #1** |
| `rustc-hash` version 1.1.0 vs 2.x | consistency (blocking), dependencies, rust | 85%/70%/70% | Addressed in #1; defensible per transitive-unification rationale |
| Bench naming `_bench` suffix | consistency | 90% | **Inline comment #2** |
| SearchLayer lang filter double-decode | performance, architecture | 81%/80% | **Inline comment #4** |
| AstPostingSource C4 contract omission | reliability | 85% | **Inline comment #3** |
| Line-count claim (482 vs ~397) | complexity | 95% | **Inline comment #5** |

### Lower-Confidence Findings (60-79% confidence) — Summary Only

**These should be addressed but are not blocking**:

1. **`run_ngram_set` length + code organization** (complexity 82%, architecture 82%)  
   - `run_ngram_set` is ~73 lines and mixes dedup, allocation, and scoring logic
   - Optional: extract `dedup_ngram_refs` and `finalize_scores` helpers to bring the function under 50 lines
   - Consider splitting the 482-line file later (parser/engine/scoring into separate modules)

2. **IDF cache thrash on mixed-language corpora** (performance 88% confidence)  
   - Last-value cache hits 0% when postings cycle languages (sorted by FileId, not language)
   - Fix: Replace with fixed-size memo keyed by language ([Option<f64>; LANG_COUNT]), O(1) access
   - Benchmark exists to surface this; acceptable to defer to Wave-4 optimization pass

3. **Per-posting `file_meta` decode is the hot-path bottleneck** (performance 82%)  
   - Current impl stays under 100ms budget, but is the largest avoidable cost
   - Future optimization: narrow `node_count_and_lang(doc_id)` accessor (5-byte decode instead of full 19-byte entry)
   - Conditional on real 100k-file index measurements; defer with tracking issue

4. **`scores` and `meta_cache` pre-sized to `file_count` regardless of selectivity** (performance 80%)  
   - Over-allocates for rare queries (10× for trigram matching 1k of 10k files)
   - Optional: size heuristically to first/largest posting list instead of corpus size
   - Trade-off between initialization cost vs. insert-without-rehash; currently defensible

5. **KNOWLEDGE.md helper function names don't match** (documentation 96%, should-fix)  
   - Docs say `parse_bigram_q` and `parse_trigram_q`, but actual functions are `parse_bigram` and `parse_trigram`
   - Fix: Update KNOWLEDGE.md lines 350-351
   - **Note**: This is documentation-only; see separate comment below

6. **KNOWLEDGE.md test count drift** (documentation 90%, should-fix)  
   - Docs say "42 unit tests", actual count is 44
   - Fix: Update KNOWLEDGE.md line 357 to "44 unit tests"
   - **Note**: This is documentation-only; see separate comment below

7. **Tie-break sort never exercised** (testing 92% confidence, MEDIUM)  
   - Documented contract: "score-DESC, FileId-ASC tie-break" (line 267)
   - No test constructs two files with *equal* BM25 scores
   - Fix: Add `b4_equal_scores_tie_break_file_id_asc` test ensuring FileId-ASC on ties
   - **Note**: Detailed test code provided in testing review

8. **Bench-level lang-filter performance not measured** (performance 68% confidence, gap)  
   - Criterion suite has 5 scenarios but none with `--lang` filter set
   - Consider adding `SearchLayer::search` bench with lang filter to quantify the double-decode cost
   - Lower priority than the double-read fix itself

### Pre-Existing Issues & Deferred Items

**Not blocking; already addressed or intentionally deferred:**

- **#283 deferred**: SingleNode unigram handling deferred to #283; parser correctly accepts it, execution returns InvalidQuery with message referencing #283 ✓
- **#198 cross-reference mismatch** (documentation 80%): The "dedup bigram+trigram double-counting" comment targets #198 (intersection logic), but perhaps belongs in #200 (composite ranking) or a dedicated follow-up. Recommend updating #198's body to include this scope if it's the intended home.
- **#199 CLI deferred**: `--ast` flag wiring deferred to Wave 3g (#199); no `CLAUDE.md` update needed yet ✓
- **Temporal/BM25F fields silently ignored** (architecture 80%): Documented in code (lines 230-235) that AST layer doesn't honor these; intentional per Wave-4 boundary. Recommend fail-loud guard or explicit doc note.

### Documentation-Only Corrections (Separate Issue/Comment Recommended)

These are documentation-only findings and should be fixed in a documentation PR or as separate commits:

1. **KNOWLEDGE.md drift** (lines 350-351, 357): Update helper names and test count
2. **Cargo.toml changes suggested**: rustc-hash placement and bench naming (covered in inline comments)

---

## Metrics

- **Review scope**: 11 independent reviewers across 10 specialized dimensions
- **Inline comments created**: 5 (>=80% confidence blocking/should-fix findings)
- **Consolidated findings**: 8 overlapping issues (deduplicated to single comments)
- **Lower-confidence summary items**: 8 (60-79% confidence, not blocking)
- **Pre-existing/deferred items**: 4 (already addressed or intentionally deferred)

## Recommendation

**APPROVED WITH MINOR FIXES** — Recommended pre-merge actions:

1. **rustc-hash**: Move to workspace table, clarify version unification strategy (inline comment #1)
2. **Bench naming**: Add `_bench` suffix for consistency (inline comment #2)
3. **Lang filter double-read**: Consider pushing filter into scoring loop (inline comment #4)
4. **AstPostingSource contract**: Document C4 count >= 1 invariant in trait (inline comment #3)
5. **Line-count documentation**: Update claim to reflect actual file size (inline comment #5)

The tie-break test gap (testing MEDIUM) and IDF cache optimization (performance MEDIUM) are not blocking but recommended to address before Wave-4 work lands.

---
*Consolidated review from: security, architecture, performance, complexity, consistency, regression, testing, reliability, rust, dependencies, documentation*  
*Claude Code*

PR #288, opened 2026-06-08
fix(ADR-008): remove wrapper timeout caps, daemon passthrough, kill-on-drop guard

### LINKED ISSUE #126: Test Infrastructure Cleanup — Shared Utils & Module Organization
## Summary

Test code across the repository has accumulated duplication and size issues that should be addressed to improve maintainability.

### P1: Extract Shared Test Utils Module

**Status**: PRE-EXISTING  
**Impact**: MEDIUM  
**Files Affected**: 26+ test modules

The `load_fixture()` function and `make_output()` helpers are copy-pasted into every `mod tests` block. Each copy differs only in the fixture subdirectory path.

**Current pattern:**
```rust
#[cfg(test)]
mod tests {
    use super::*;
    
    fn load_fixture(name: &str) -> String {
        let path = format!("tests/fixtures/cmd/file/{}", name);
        std::fs::read_to_string(path).unwrap()
    }
    
    fn make_output(input: &str) -> String {
        // ... logic that is identical across modules
    }
}
```

This pattern appears in all of:
- `crates/rskim/src/cmd/file/*.rs` (find, grep, ls, rg, tree)
- `crates/rskim/src/cmd/infra/*.rs` (aws, curl, gh, wget)
- `crates/rskim/src/cmd/log.rs`
- Any future cmd modules

**Proposed solution:**

Create `crates/rskim/src/cmd/test_utils.rs`:
```rust
#[cfg(test)]
pub mod test_utils {
    /// Load fixture file from cmd subdir
    /// 
    /// # Example
    /// ```
    /// load_fixture("file/grep", "basic.txt") 
    ///   -> tests/fixtures/cmd/file/grep/basic.txt
    /// ```
    pub fn load_fixture(subdir: &str, name: &str) -> String {
        let path = format!("tests/fixtures/cmd/{}/{}", subdir, name);
        std::fs::read_to_string(path).unwrap()
    }
    
    pub fn make_output(input: &str) -> String {
        // Implementation shared by all modules
    }
}
```

Then each test module becomes:
```rust
#[cfg(test)]
mod tests {
    use super::*;
    use crate::cmd::test_utils::{load_fixture, make_output};
    
    #[test]
    fn test_example() {
        let fixture = load_fixture("file/grep", "basic.txt");
        // ...
    }
}
```

### P2: Standardize `make_output` Signatures

**Status**: BLOCKED_ON P1  
**Impact**: LOW  
**Files Affected**: file/grep.rs, file/find.rs, infra/*.rs

Three different `make_output` signatures exist across modules. Once P1 extracts the shared module, standardize on a single signature that handles all cases.

**Current variations:**
- `fn make_output(input: &str) -> String`
- `fn make_output(input: &str, args: &[&str]) -> String`
- `fn make_output(input: &str, mode: OutputMode) -> String`

### P3: Split `log.rs` Test Module

**Status**: PRE-EXISTING  
**Impact**: LOW  
**File**: crates/rskim/src/cmd/log.rs:470-869

The `mod tests` block in `log.rs` spans ~400 lines with 26 test functions. While each individual test is well-scoped and the production code is clean, grouping tests into sub-modules would improve navigability.

**Proposed structure:**
```rust
#[cfg(test)]
mod tests {
    use super::*;
    use crate::cmd::test_utils::load_fixture;
    
    mod json_tests {
        use super::*;
        
        #[test]
        fn test_json_output_basic() { /* ... */ }
    }
    
    mod regex_tests {
        use super::*;
        
        #[test]
        fn test_regex_pattern_matching() { /* ... */ }
    }
    
    mod edge_case_tests {
        use super::*;
        
        #[test]
        fn test_dedup_empty_input() { /* ... */ }
    }
}
```

## Priority & Sequencing

1. **P1 (MUST)**: Extract `test_utils` module — enables P2 and P3, reduces duplication across 26+ files
2. **P2 (SHOULD)**: Standardize signatures — depends on P1, touch 3-4 modules
3. **P3 (NICE)**: Split log tests — cosmetic improvement, independent of P1/P2

## Acceptance Criteria

- [ ] `crates/rskim/src/cmd/test_utils.rs` created with `load_fixture(subdir, name)` and shared helpers
- [ ] All 26+ test modules updated to use `use crate::cmd::test_utils`
- [ ] P2 applied: `make_output` signature unified across all modules
- [ ] `log.rs` tests reorganized into sub-modules (json, regex, edge_case)
- [ ] All 1600+ tests still pass
- [ ] No fixtures or test data modified — only helper extraction and organization

dean0x:
## Summary

This PR merges two significant bodies of work onto the main branch:

1. **#126 test-infrastructure cleanup**: Refactored the test module structure to improve maintainability and code reuse, decoupling test helpers from test implementation.

2. **ADR-008 wrapper timeout removal**: Removed internal wall-clock command timeouts and added intelligent detection of indefinite-running commands (watch, tail -f, dev-servers) with inherited stdio passthrough, preventing orphan processes via RAII guard.

### Test Infrastructure Cleanup (#126)

**What changed:**
- Split `cmd/log.rs` flat test module into sub-modules (`log_test/core.rs`, `log_test/fixtures.rs`, etc.)
- Migrated `git/fetch` and `vitest`/`go` test fixtures onto shared `load_fixture()` helper
- Renamed `test_support.rs` → `test_utils.rs` for clarity and consistency
- Hardened `load_fixture()` with struct param validation and better error messages
- Updated feature knowledge documentation after renames

**Commits:**
- a7ffa10: test infrastructure cleanup — log.rs split, fetch.rs migration, test_support→test_utils rename
- 65032f2: address review findings — load_fixture hardening, make_file params struct, import order
- e57e267: sync feature knowledge after test_support -> test_utils rename
- c7fc211: migrate vitest/go test fixtures onto shared load_fixture
- e69755d: address cycle-2 review suggestions — docs, idiom, guard

**Benefits:**
- Reduced test boilerplate and duplication
- Fixtures now reusable across test modules
- Clearer separation between core logic and test helpers
- 14 call sites across cmd/{git,heatmap,build,test,execution} remain unaffected

### ADR-008: Wrapper Timeout Removal & Daemon Detection

**What changed:**

**Part A: Stateless CommandRunner**
- Removed `RunnerError::Timeout` variant
- Removed `DEFAULT_CMD_TIMEOUT` constant (was 30s)
- Removed `wait_with_timeout()` method
- Updated all 14 call sites across `cmd/{dispatch,execution,build,heatmap,git,test}` to use `wait()` directly
- CommandRunner no longer enforces internal time caps; external CI/CD timeout patterns now fully responsible

**Part B: ChildGuard RAII Guard**
- New `ChildGuard` struct wraps spawned `Child` and calls `kill()` + `wait()` on drop
- Prevents orphan processes on any early-return path (panic, ?, early break, etc.)
- Applied to both passthrough dispatch and audit/hook paths

**Part C: Indefinite Command Detection**
- New `cmd/rewrite/indefinite.rs` module detects long-running commands:
  - Watch modes: `*watch`, `*dev` (vitest, jest, webpack, tsc, vite, etc.)
  - Streaming: `tail -f`, `journalctl -f`, `docker logs -f`, `kubectl logs -f`
  - Dev servers: `serve`, `http-server`, `nodemon`, `pm2 run dev`, dev-server frontends
- Integrated into hook path (`audit_hook()`) and dispatch path (`run_inherited_passthrough()`)
- **TTY gate removed**: indefinite detection fires unconditionally (non-TTY sub-agents and CI pipelines now also detect indefinite commands)
- **Passthrough exception**: Piped stdin skips daemon guard so piped compression still works
- Exit code mapping: ENOENT → 127 (consistent with shell semantics)

**Documentation:**
- Updated `security.md` to clarify: no internal time cap, CI/CD external timeout pattern, dev-server inherited-stdio behavior
- CHANGELOG entries for timeout removal, daemon passthrough, and killing processes

**Commits:**
- 667b217: remove wrapper timeout caps, add kill-on-drop guard, daemon passthrough
- 18bc8ff: drop is_terminal gate on daemon passthrough — stream indefinite commands regardless of stdin
- d2d0117: simplify(indefinite): collapse duplicate match arms in is_indefinite_command
- b3764f5: treat --help/--version as finite in daemon guard

**Trade-off Accepted:**
- `cat output | skim vitest` now runs vitest live (decompressed)
- Use `skim vitest run` or `SKIM_PASSTHROUGH=1` to compress piped output

**Benefits:**
- No more internal timeout kills masking actual application hangs
- Prevents orphan processes across all code paths
- Dev-server commands run with live output (expected user behavior)
- Non-TTY environments (CI, sub-agents, PATH wrappers) now get full daemon detection
- External timeout mechanism in CI/CD pipelines is now the only source of truth

## Breaking Changes

None. Behavior unchanged for finite commands. Indefinite commands now run with inherited stdio by default (expected behavior).

## Testing

All 5,213 existing tests pass. New tests added for:
- `is_indefinite_command()` coverage (vitest, jest, webpack, tsc, tail -f, docker logs, kubectl logs, serve, http-server, nodemon, pm run, watch patterns, --help, --version)
- `ChildGuard` RAII drop behavior on panic/early-return
- `run_inherited_passthrough()` ENOENT → 127 mapping (deterministic)
- Piped stdin bypassing daemon guard (compression verification)
- TTY gate removal (non-TTY indefinite detection)

## Related Issues

Closes #121 (ADR-008 implementation)
Closes #126 (test infrastructure cleanup)

### DIFF
--- a/.devflow/.gitignore
+++ b/.devflow/.gitignore
@@ -1,46 +1,27 @@
-# Per-developer session state (fully transient)
-memory/
+# .devflow/ git-tracking policy
+# ---------------------------------------------------------------------------
+# Only curated, shared team knowledge is committed to git:
+#   - decisions/decisions.md, decisions/pitfalls.md      (ADR / pitfall records)
+#   - features/index.json, features/<slug>/KNOWLEDGE.md  (feature knowledge bases)
+#
+# Everything else under .devflow/ is per-developer or transient (memory, dream,
+# docs, locks, runtime state, manifest, scratch results) and is
+# intentionally ignored. Model: ignore-by-default, then re-include the curated
+# files. Any NEW file under .devflow/ is ignored unless explicitly listed below.
 
-# Sidecar dispatch system (fully transient)
-sidecar/
+# 1. Ignore everything under .devflow/ by default
+*
 
-# Per-developer observation logs and transient state
-learning/learning-log.jsonl
-learning/learning-log.v1.jsonl.bak
-learning/learning.json
-learning/.learning-manifest.json
-learning/.learning-notified-at
-learning/.learning-notifications.json
-learning/.learning-runs-today
-learning/.learning-session-count
-learning/.learning-batch-ids
-learning/debug/
+# 2. Keep this policy file
+!.gitignore
 
-# Per-developer decisions observation log and transient state
-decisions/decisions-log.jsonl
-decisions/decisions.json
-decisions/.decisions-manifest.json
-decisions/.decisions.lock/
-decisions/.decisions-usage.json
-decisions/.decisions-usage.lock/
-decisions/.decisions-notifications.json
-decisions/.decisions-runs-today
-decisions/.decisions-batch-ids
-decisions/.disabled
+# 3. Track the decisions knowledge (not its log / config / locks / usage state)
+!decisions/
+!decisions/decisions.md
+!decisions/pitfalls.md
 
-# Ephemeral doc artifacts
-docs/handoff-*.md
-docs/reviews/*/.last-review-head
-
-# Transient files in features
-features/.knowledge.lock/
-features/.knowledge-last-refresh
-features/.knowledge-refresh.lock
-features/.disabled
-features/.gitignore-configured
-
-# Install state (local-scope only)
-manifest.json
-
-# Init marker
-.gitignore-configured
+# 4. Track the feature knowledge bases (not locks / sentinels / scratch results)
+!features/
+!features/index.json
+!features/*/
+!features/*/KNOWLEDGE.md
--- a/.devflow/decisions/pitfalls.md
+++ b/.devflow/decisions/pitfalls.md
@@ -1,4 +1,4 @@
-<!-- TL;DR: 5 pitfalls. Key: PF-001, PF-002, PF-003, PF-004, PF-005 -->
+<!-- TL;DR: 6 pitfalls. Key: PF-002, PF-003, PF-004, PF-005, PF-006 -->
 # Known Pitfalls
 
 Area-specific gotchas, fragile areas, and past bugs.
@@ -47,3 +47,12 @@ Area-specific gotchas, fragile areas, and past bugs.
 - **Resolution**: before enforcing an inherited numeric criterion, trace it to a measured basis
 - **Status**: Active
 - **Source**: self-learning:obs_acqv8m
+
+## PF-006: Feature-knowledge files (KNOWLEDGE.md, index.json) go stale after a rename refactor, leaving broken import-path references
+
+- **Area**: .devflow feature-knowledge maintenance after refactors
+- **Issue**: a module or symbol rename (test_support to test_utils, PR #279) propagated through 77 compiled .rs files but left build-parsers KNOWLEDGE.md and index.json referencing the old name and a now-broken crate::cmd::test_support import path, because doc and metadata files are not type-checked by cargo
+- **Impact**: feature knowledge surfaced to future agents points at non-existent symbols, eroding trust in the pre-computed context and causing wasted exploration
+- **Resolution**: when a rename refactor touches symbols named in feature knowledge, grep KNOWLEDGE.md and index.json for the old identifier with word boundaries and update referencedFiles and import paths in the same PR — treat feature-knowledge drift as part of the refactor blast radius
+- **Status**: Active
+- **Source**: self-learning:obs_knstal
--- a/.devflow/features/build-parsers/KNOWLEDGE.md
+++ b/.devflow/features/build-parsers/KNOWLEDGE.md
@@ -1,9 +1,9 @@
 ---
 feature: build-parsers
 name: Build Tool Output Parsers
-description: "Use when adding a new build tool parser, modifying cargo/tsc/make/gradle/maven compression, or debugging three-tier parse degradation for build commands. Keywords: build, cargo, tsc, make, gradle, maven, clippy, ParseResult, BuildResult, three-tier, NDJSON, flag injection, run_check, run_fmt, parse_fmt."
+description: "Use when adding a new build tool parser, modifying cargo/tsc/make/gradle/maven compression, or debugging three-tier parse degradation for build commands. Keywords: build, cargo, tsc, make, gradle, maven, clippy, ParseResult, BuildResult, three-tier, NDJSON, flag injection, run_check, run_fmt, ChildGuard, indefinite, is_indefinite_command."
 category: component-patterns
-directories: [crates/rskim/src/cmd/build/]
+directories: [crates/rskim/src/cmd/build/, crates/rskim/src/cmd/]
 referencedFiles:
   - crates/rskim/src/cmd/build/mod.rs
   - crates/rskim/src/cmd/build/cargo.rs
@@ -20,9 +20,10 @@ referencedFiles:
   - crates/rskim/src/cmd/registry.rs
   - crates/rskim/src/cmd/test_utils.rs
   - crates/rskim/src/runner.rs
+  - crates/rskim/src/cmd/rewrite/indefinite.rs
 created: 2026-05-14
-updated: 2026-06-07
-version: 11
+updated: 2026-06-08
+version: 14
 ---
 
 # Build Tool Output Parsers
@@ -38,12 +39,12 @@ The module is invoked via flat dispatch (`skim tsc`) or multi-category dispatch 
 `cmd/mod.rs` was refactored to reduce complexity. Functionality previously inline in `mod.rs` is now split across dedicated submodules:
 
 - `cmd/dispatch.rs` — `dispatch()`, `run_raw_passthrough()`, and per-tool dispatcher helpers (`dispatch_cargo`, `dispatch_go`, `dispatch_swift`, `dispatch_dotnet`, `passthrough_subcmd`, `extract_subcmd`, `prepend_without`)
-- `cmd/execution.rs` — `OutputFormat`, `RunContext`, `ParsedCommandConfig<'_>`, `ToolRunConfig<'_>`, `run_tool<T>`, `run_parsed_command_with_mode`, `format_analytics_label`
+- `cmd/execution.rs` — `OutputFormat`, `RunContext`, `ParsedCommandConfig<'_>`, `ToolRunConfig<'_>`, `run_tool<T>`, `run_parsed_command_with_mode`, `format_analytics_label`, `combine_output`, `obtain_output`, `render_output<T>`, `record_and_report`, `passthrough_raw`
 - `cmd/security.rs` — `sanitize_for_display`, `scrub_db_args`, `scrub_infra_args`
 - `cmd/registry.rs` — `KNOWN_SUBCOMMANDS` (sorted, binary-searchable), `is_known_subcommand`, `is_meta_subcommand`, `wrapper_targets`
 - `cmd/test_utils.rs` — standalone `pub(crate)` module (compiled under `#[cfg(test)]` gate in `mod.rs`); canonical test helper source for all `cmd` subtree tests
 
-`cmd/mod.rs` remains the coordination point (declares all submodules, re-exports public API) and still houses the inline helpers: `user_has_flag`, `inject_flag_before_separator`, `extract_show_stats`, `extract_json_flag`, `extract_output_format`, `combine_output`, `should_read_stdin`, `read_bounded`, `read_stdin_bounded`, `MAX_STDIN_BYTES`, passthrough checks, `resolve_cache_dir`, `obtain_output`, `render_output<T>`, `DEFAULT_CMD_TIMEOUT`.
+`cmd/mod.rs` remains the coordination point (declares all submodules, re-exports public API) and still houses the inline helpers: `user_has_flag`, `inject_flag_before_separator`, `extract_show_stats`, `extract_json_flag`, `extract_output_format`, `should_read_stdin`, `read_bounded`, `read_stdin_bounded`, `MAX_STDIN_BYTES`, passthrough checks (`is_passthrough_mode`, `check_passthrough_str`, `check_passthrough_value`), `resolve_cache_dir`, and `skim_wrappers_dir`.
 
 When adding a new parser, register the tool in `cmd/registry.rs` (`KNOWN_SUBCOMMANDS` array) and `cmd/dispatch.rs` (`dispatch()` match arm), not in `cmd/mod.rs`.
 
@@ -165,7 +166,7 @@ Tier 2 noise patterns strip `[INFO] Downloading/Downloaded from`, `[INFO] ---` s
 ## `run_parsed_command` — Shared Infrastructure
 
 All five parsers call `super::run_parsed_command(...)` in `mod.rs` instead of spawning processes themselves. This function handles:
-1. Command spawn with a 600-second timeout (compile times can be long)
+1. Command spawn — blocks on a plain `wait()` with no internal timeout (see ADR-008)
 2. ANSI escape code stripping from both stdout and stderr before parsing
 3. Calling the parser function pointer
 4. Emitting degradation markers to stderr (only when `--debug` / `SKIM_DEBUG=1` is active — silent by default)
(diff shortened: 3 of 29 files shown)


dean0x:
## Code Review: High-Confidence Findings

### 1. **CRITICAL** — pm_is_indefinite: yarn/bun branch misreads script when flags precede run

**Files**: `crates/rskim/src/cmd/rewrite/indefinite.rs:213-225` (Rust 85% + Complexity 82%)

**Issue**: The `yarn` / `bun` branch recomputes `first` via `.find(|s| !s.starts_with('-'))`, which is correct. However, when flags precede `run`, the code re-scans with `rest.iter().skip_while(...)` starting from the **beginning of `rest`**, not from the position of `run`. 

For a command like `yarn --silent run dev`:
1. `first` = `None("--silent")` (first non-flag token)
2. Match arm: `Some("run") | Some("run-script")` — not matched because `first` is `"--silent"`
3. Falls through to `other => other` arm, returning `"--silent"` as the "script"

This misclassifies `yarn --silent run dev` as **finite** instead of indefinite. The dev server would be routed to buffered capture, appearing to hang until the 64 MiB cap or external timeout fires — exactly the footgun ADR-008 aims to prevent.

**Confidence**: 85% (functional bug independently identified by Rust and Complexity reviewers)

**Fix**: Collapse to a single positional walk using `filter()` to skip flags, then `skip_while()` to skip `run`/`run-script`:

```rust
_ => {
    let mut positionals = rest.iter().copied().filter(|s| !s.starts_with('-'));
    match positionals.next() {
        Some("run") | Some("run-script") => positionals.next(),
        other => other,
    }
}
```

This removes the double-scan and the footgun.

---

### 2. **PERFORMANCE** — Double tokenization on hook rewrite hot path

**Files**: `crates/rskim/src/cmd/rewrite/hook.rs:204` and `:219` (Performance 90%)

**Issue**: The `run_hook_mode` function tokenizes the command string twice in immediate succession:
- Line 204: `let quick_tokens: Vec<&str> = command.split_whitespace().collect();` (for indefinite check)
- Line 219: `let tokens: Vec<&str> = command.split_whitespace().collect();` (for rewrite engine)

The first Vec is dropped after the indefinite block, then an identical tokenization runs again. This is waste on skim's hottest path (every agent shell command).

**Confidence**: 90%

**Fix**: Hoist a single tokenization and reuse it:

```rust
let tokens: Vec<&str> = command.split_whitespace().collect();
if tokens.is_empty() {
    audit_hook(&command, false, "");
    return Ok(ExitCode::SUCCESS);
}

// Fast-path: indefinite commands must not be rewritten
if super::indefinite::is_indefinite_command(&tokens) {
    audit_hook(&command, false, "");
    return Ok(ExitCode::SUCCESS);
}

let has_operator_chars = command.contains("&&") || /* ... */;
// ... existing rewrite logic, reusing `tokens`
```

---

### 3. **CONSISTENCY** — Duplicate ChildGuard type

**Files**: `crates/rskim/src/runner.rs:258` vs `crates/rskim/src/cmd/infra/gh/streaming.rs:285` (Consistency 90%)

**Issue**: The `ChildGuard` struct is now defined identically in two places:
- `runner.rs:258`: new, ADR-008
- `gh/streaming.rs:285`: pre-existing

Both have the same `struct ChildGuard(std::process::Child)` and identical `Drop` impl. Per CLAUDE.md (consistency beats cleverness, single responsibility), this should be one canonical definition.

**Confidence**: 90%

**Fix**: Promote `runner.rs`'s `ChildGuard` to `pub(crate)` and import it in `gh/streaming.rs`, deleting the duplicate. `runner.rs` is the canonical owner (already owns `CommandOutput`, `MAX_OUTPUT_BYTES`, `is_spawn_error`).

---

### 4. **CLARITY** — Conflicting guards on precondition

**Files**: `crates/rskim/src/cmd/rewrite/indefinite.rs:40-43` (Complexity 88%)

**Issue**: The function has three overlapping layers protecting the same precondition:
- Doc comment: "must NEVER be called on zero-length slices"
- `debug_assert!(!tokens.is_empty(), ...)`
- Runtime `if tokens.is_empty() { return false; }`

Both callers already guarantee non-empty input, so the runtime `if` is dead code. This muddies the contract: is empty a programmer error (assert) or a recoverable input (return)?

**Confidence**: 88%

**Fix**: Pick one contract. Recommend keeping the conservative `if` (defense-in-depth) and dropping the `debug_assert!`. Update the doc comment from "must NEVER" to "callers should pass non-empty; empty returns false."

---

### 5. **DOCUMENTATION** — Missing doc comment on run_inherited_passthrough ownership model

**Files**: `crates/rskim/src/cmd/dispatch.rs:97-108` (Architecture 82%)

**Issue**: `run_inherited_passthrough` spawns via `Command::status()` and does NOT wrap in `ChildGuard`, unlike every `CommandRunner` spawn. This diverges from ADR-008's stated "uniform child-process lifetime mechanism" via ChildGuard. In practice, `status()` blocks to completion so there is no early-return window — the real-world orphan risk is low. But the architectural exception is not documented, leaving a coherence gap.

**Confidence**: 82%

**Fix**: Add a doc comment explaining why the guard is intentionally absent:

```rust
/// Unlike `CommandRunner::run_with_env`, this path does NOT wrap the child in
/// `ChildGuard`: `status()` spawns and blocks to completion in one call, with no
/// early-return window between spawn and reap. The child is always reaped before
/// this function returns.
```

---

### 6. **TEST QUALITY** — Tautological assertion

**Files**: `crates/rskim/tests/cli_e2e_daemon_passthrough.rs:100-119` (Testing HIGH 95%)

**Issue**: The `test_hook_mode_jest_ci_is_rewritten` assertion is:
```rust
.stdout(predicate::str::contains("skim").or(predicate::str::is_empty()))
```

The `.or(is_empty())` makes this a tautology — any stdout (empty, "skim ...", or arbitrary text) satisfies it. The test names a behavior it cannot verify.

**Confidence**: 95% (HIGH severity)

**Fix**: Either make the assertion real:
```rust
.stdout(predicate::str::contains("skim jest"));
```

Or delete this test if jest's rewrite presence is environment-dependent. The negative case (`jest --watch` NOT rewritten) is already covered elsewhere.

---

**Summary**: 5 inline findings + 1 summary comment below for medium-confidence items.

Prepared by Claude Code / Anthropic.

dean0x:
## Code Review: Medium-Confidence & Polish Items (60-79%)

### Documentation & Feature Knowledge

**Committed build-parsers/KNOWLEDGE.md stale** (Documentation MEDIUM, 95% → already fixed in working tree)
- The v11 version committed on HEAD still documents deleted `DEFAULT_CMD_TIMEOUT` as live, specifically: "Timeout is 600 seconds (10 minutes)..." (line 278)
- The working-tree v14 already fixes this with an anti-pattern note
- **Action**: Commit the working-tree `.devflow/features/build-parsers/KNOWLEDGE.md` edits onto the branch before merge to prevent stale documentation from landing

---

### Reliability & Resource Cleanup

**ChildGuard::drop grandchild leak** (Reliability MEDIUM, 82%)
- `ChildGuard` calls `kill()` then `wait()` on the direct child only — grandchildren (common for `npm → node`, `cargo → rustc`) are orphaned and may leak
- Impact: reader threads may block on pipes held open by surviving grandchildren; process-lifetime bounded but not main-thread hanging
- **Optional hardening**: spawn child in its own process group (`Command::process_group(0)`) and signal the group on drop (Unix-only)
- **Note**: this is a known limitation consistent with feature knowledge scope, not a regression

**run_inherited_passthrough swallows non-ENOENT spawn errors** (Reliability MEDIUM, 68%)
- Line 102: `Err(_) => ExitCode::FAILURE` discards the error detail without logging
- Violates the project's "fail loud" philosophy
- **Fix**: Add `eprintln!("error: failed to spawn {}: {}", program, e)` to match the streaming gh path style

---

### API Contract & Layering

**is_indefinite_command consumed at two layers without shared contract** (Architecture LOW, 80%)
- `hook.rs:205` tokenizes raw compound string; `dispatch.rs:425` tokenizes pre-parsed argv
- Behavior is correct but the compound-handling contract is implicit
- **Fix**: Add one-line doc note to `is_indefinite_command` stating that a leading indefinite program in a compound command is treated as whole-command indefinite

**should_read_stdin["run"] exception silently broadens beyond vitest/jest** (Regression LOW, 70%)
- The new `args.len() == 1 && args[0] == "run"` check is consumed by 7 call sites, not just test runners
- Practically zero risk (no tool uses `run` as a valid target), but the doc frames it as test-runner-only
- **Fix**: extend the comment to note cross-family reach, or document the decision to keep the scope global

---

### Testing Gaps

**No positive test for the core ADR-008 guarantee** (Testing MEDIUM, 85%)
- ADR-008's headline is "skim never kills a command the user explicitly invoked"
- The three deleted timeout tests correctly removed the old killing behavior
- But nothing replaced the inverse assertion — the suite has no positive guard against a reintroduced cap
- **Fix**: Add one cheap test asserting completion past a duration that the old cap would have killed:
```rust
#[test]
fn run_does_not_impose_internal_timeout() {
    let runner = CommandRunner::new();
    // ~1.5s exceeds any plausible accidental short cap
    let result = runner.run("sleep", &["1.5"]).unwrap();
    assert_eq!(result.exit_code, Some(0), "finite command must not be killed");
    assert!(result.duration >= std::time::Duration::from_millis(1400));
}
```

**Dispatch-level daemon guard non-firing not asserted** (Testing MEDIUM, 80%)
- `is_indefinite_command` is unit-tested thoroughly, but the dispatch wiring (finite commands NOT diverted to `run_inherited_passthrough`) is only indirectly covered
- A future broadening of `is_indefinite_command` could silently route `cargo test` to passthrough with no failing test
- **Fix** (acceptable today): The existing E2E tests (`cli_e2e_exit_codes`, `cli_e2e_test_parsers`) implicitly cover this (they produce compressed output, proving the guard fell through). To make it explicit, add an assertion that a finite build command still produces compressed structured output

---

### Pre-Existing Issues Noted (Not Blocking)

**Stale pitfall citations in gh/streaming.rs** (Consistency LOW, 85%)
- Pre-existing `ChildGuard` comment cites `PF-025`/`PF-023`, which don't resolve in current decisions corpus
- Not in your changed code, but if the ChildGuard consolidation (item #3 above) proceeds, the surviving definition should cite ADR-008 (already correct in runner.rs)

**Non-ENOENT spawn error handling** (Performance/Reliability MEDIUM, 62% combined)
- `run_inherited_passthrough` swallows errors without logging (noted above)
- `should_read_stdin` doc scope (noted above)

---

**All findings recorded for user decision. No additional inline comments created for 60-79% confidence items per the comment-pr SOP.**

Prepared by Claude Code / Anthropic.

dean0x:
## Exit-code mapping divergence (88% confidence)

The exit code from `run_inherited_passthrough` (dispatch.rs:126) uses `.clamp(0, 255)` to safely handle out-of-range exit codes, but the streaming twin at streaming.rs:488 uses unclamped `code as u8`, which silently wraps values >255 (e.g., 256 → 0). 

The doc comment claims it "mirrors the streaming `gh` path" (dispatch.rs:107), but the two implementations no longer match.

**Fix:** In `streaming.rs:488`, change:
```rust
Some(code) => ExitCode::from(code as u8),
```
to:
```rust
Some(code) => ExitCode::from(code.clamp(0, 255) as u8),
```

This aligns the streaming path with the new safer dispatch idiom and the existing `run_raw_passthrough` pattern (dispatch.rs:155).

dean0x:
## ENOENT branch is silent; should fail loud (80% confidence)

The ENOENT arm here maps to exit 127 with **no diagnostic**, but the non-ENOENT arm (line 122) fails loud with `eprintln!`. The doc comment also claims to "mirror the streaming `gh` path", which *does* print `error: {cmd} not found on PATH` on ENOENT.

This violates the project philosophy (CLAUDE.md: fail loud with clear error messages) and creates an inconsistency: `skim vite` without vite installed exits silently with code 127, while `skim gh run watch` without gh installed prints a helpful diagnostic.

**Fix:** Add a diagnostic to match the twin:
```rust
Err(e) if e.kind() == std::io::ErrorKind::NotFound => {
    eprintln!("error: {program} not found on PATH");
    ExitCode::from(127u8)
}
```

dean0x:
## ENOENT→127 mapping is not actually asserted (92% confidence)

The test at line 752 calls `run_inherited_passthrough("__skim_guaranteed_absent_binary__", &[])` but binds the result to `let _code = ...` — it never asserts the code is 127. The doc comment claims it "verifies the ENOENT branch is reached," but the test degrades to "does not panic" only.

A regression that mapped ENOENT to `ExitCode::FAILURE` (exit 1) or dropped the `NotFound` arm entirely would NOT be caught. The review specifically asks for a deterministic 127 test.

**Fix:** Extract the ENOENT→code mapping into a pure helper and unit-test the 127 value:
```rust
fn spawn_status_to_code(status: io::Result<ExitStatus>) -> u8 {
    match status {
        Err(e) if e.kind() == std::io::ErrorKind::NotFound => 127,
        Err(_) => 1,  // FAILURE
        Ok(s) => s.code().unwrap_or(1).clamp(0, 255) as u8,
    }
}
```

Then unit test this pure function with a synthesized `NotFound` error — no subprocess dependency needed.

dean0x:
## CHANGELOG missing user-visible behavior change (82% confidence)

The "Added — Daemon / streaming command passthrough" entry (lines 30-38) lists examples and mentions the buffered fallback, but it does **not** document the concrete behavior change: **bare `skim vitest` (and piped `cat output | skim vitest`) now runs vitest live instead of compressing**.

This is documented thoroughly in docs/security.md and build-parsers KNOWLEDGE.md, but users scanning only the CHANGELOG won't know the regression. Keep-a-Changelog convention treats behavior changes to existing invocations as notable.

**Fix:** Add a bullet under the "Changed" section (line 10):
```markdown
- **`skim vitest` now runs in watch mode by default (ADR-008 Part C)** — bare
  `skim vitest` and `cat output | skim vitest` are detected as indefinite and
  passed through live (no compression). To compress a one-shot run, use
  `skim vitest run` (or set `SKIM_PASSTHROUGH=1` to forward piped content).
```

dean0x:
## Code Review Cycle 2: Summary

This is **review cycle 2**. All 10 reviewers (architecture, complexity, consistency, documentation, performance, regression, reliability, rust, security, testing) have approved with the following conditions.

### Overall Recommendation: APPROVED_WITH_CONDITIONS

✅ All reviewers approve. The changes are architecturally sound, well-tested, and thoroughly documented.

**Conditions:** Four should-fix items with 80%+ confidence:

1. **Exit-code clamp divergence** (streaming.rs:488) — posted as inline comment
2. **ENOENT silent failure** (dispatch.rs:117) — posted as inline comment
3. **ENOENT→127 untested assertion** (dispatch.rs:752) — posted as inline comment  
4. **CHANGELOG missing `skim vitest` behavior change** — posted as inline comment

These are should-fix items but fall below the inline-comment threshold. Consider addressing them:

- **Dead allocation in non-compound path** (performance, 65%) — `hook.rs:223` allocates `tokens.join(" ")` before the compound-branch gate, so non-compound calls do a dead allocation. Move the `let original = …` into the `else` branch for a negligible but real speedup.

- **`run_hook_mode` function length** (complexity, 65%) — Pre-existing, only lightly touched. ~160 lines is long but the control flow is linear (read → parse → rewrite → emit with early returns). Not blocking; refactoring is out of ADR-008 scope.

- **Build-parsers KNOWLEDGE.md miscount** (documentation, pre-existing) — Line 108 says "four LazyLock<Regex> noise patterns" but make.rs:200-217 has exactly three. Fix: change "four" to "three". (Pre-existing, not introduced here, but worth fixing while the file is open.)

- **`kubectl_logs_is_indefinite` naming asymmetry** (consistency, 62%) — Unlike sibling `docker_is_indefinite`, this embeds the subcommand in the name. Optional; the split is otherwise clean and intentional.

### Cycle-1 Fixes All Hold

Cross-cycle verification confirmed all six cycle-1 fixes remain intact:
- pm flag-before-run single-pass scan (indefinite.rs:227-234)
- ChildGuard consolidated to runner.rs (streaming.rs:75 imports canonical)
- Double split_whitespace hoisted to one Vec (hook.rs:199)
- Fail-loud on non-ENOENT spawn error (dispatch.rs:118-124)
- `should_read_stdin` family-generic doc (mod.rs:93-96)
- Feature knowledge synced to v14

No regressions detected. The diff is clean for a 14-call-site refactor.

---

**Review Stats:**  
| Category | CRITICAL | HIGH | MEDIUM | LOW |
|----------|----------|------|--------|-----|
| Blocking | 0 | 0 | 0 | — |
| Should Fix | — | 0 | 6 | 4 |
| Pre-existing | — | — | 1 | 0 |

**Average Reviewer Score:** 9.1 / 10  
**Confidence:** Very High (all reviewer concerns directly addressable)

---

*Generated by Claude Code review cycle 2 — PR ready for merge once should-fix items are addressed.*

Issue #290, opened 2026-06-08
AST-index incremental build cache (skip re-extraction of unchanged files)


dean0x:
**Decision #7 from Wave 3g (#199).**

#199 ships the AST index with an **always-build, no-cache** baseline: every index build/refresh (including the backgrounded \`skim search --update\` run by git hooks) re-extracts the AST n-grams for **all** files via a second tree-sitter pass. On large repos this makes interactive rebuild-after-change and post-commit \`--update\` slower than necessary.

**Goal:** persist per-file AST n-grams keyed by content SHA so unchanged files skip re-extraction, keeping \`--update\` fast.

**Plan:**
- Cache \`(file_sha) -> AstNgramSet + StructuralMetrics + node_count\` keyed by content hash.
- On refresh, reuse cached entries for unchanged files; re-extract only changed/new files.
- Preserve the FileId-alignment invariant (cache lookups must not reorder files relative to the lexical index).

Relates to #199 (Wave 3g), #286 (query-engine perf), #273 (on-disk AST compression).

dean0x:
Resolved on `wave/wave4-search` (Wave-4 compound engine + AST/temporal follow-ups). Tracked in North Star #174. Not yet on `main` — ships with PR #368 (wave→main).

Issue #289, opened 2026-06-08
Temporal populate path + join unified auto-refresh (latent #257 gap)


dean0x:
**Discovered during Wave 3g (#199).**

\`temporal.db\` is never populated by the CLI. \`TemporalDb::sync\` / \`store_*\` are defined but only ever called from tests; \`heatmap\` only prints metrics and never persists. As a result the search \`--hot\` / \`--cold\` / \`--risky\` / \`--blast-radius\` flags read a database that nothing writes — they have silently no-op'd on real repositories since #257.

**Goal:** build the missing temporal writer and join it to the unified auto-refresh authority shipped in #199.

**Plan:**
- Implement \`rebuild_temporal(root, cache_dir, head)\`: \`parse_history\` → \`compute_file_risk_scores\` + \`compute_file_temporal_stats\` + co-change matrix → \`TemporalDb::sync(.., head)\`.
- Call it from the clearly-marked, ticket-referenced hook point left in \`ensure_indexes_fresh\` (\`crates/rskim/src/cmd/search/staleness.rs\`), after the lexical+AST rebuild, under the same \`.skim-build.lock\`, reusing the already-read HEAD (one call, no rework).
- Include an incremental walk (\`parse_history\` \`lookback_days\`) so refreshes stay cheap.

**Outcome:** temporal search works for the first time; all three indexes (lexical, AST, temporal) refresh through one freshness authority.

Relates to #199 (Wave 3g), #257 (temporal CLI flags), #202 (full compound CLI).

dean0x:
Resolved on `wave/wave4-search` (Wave-4 compound engine + AST/temporal follow-ups). Tracked in North Star #174. Not yet on `main` — ships with PR #368 (wave→main).

PR #257, opened 2026-05-26
feat(#189): temporal CLI flags for skim search (--hot, --cold, --risky, --blast-radius)

dean0x:
## Summary

- Add composable temporal search flags to `skim search`: `--hot`, `--cold`, `--risky` (mutually exclusive sort modes) and `--blast-radius <FILE>` (co-change pre-filter)
- These flags work both standalone (no text query — lists top hotspots/risks/partners) and combined with text queries (enriches and re-sorts BM25F results with temporal scores)
- Temporal annotations appear in both text output (`hotspot: 0.950  risk: 0.800`) and JSON output (`"temporal": {"hotspot_score": 0.95, ...}`)

## Key Changes

**rskim-search library:**
- `SearchQuery.file_filter`: new `Option<HashSet<FileId>>` field enabling blast-radius pre-filtering in the BM25F scoring loop — results are restricted to co-change partners before scoring
- `TemporalDb` per-file lookups: `hotspot_for_file`, `risk_for_file`, `cochanges_for_file` (OR-keyed SQL for symmetric file_a/file_b pairs)
- `TemporalDb` top-N queries: `top_hotspots`, `top_risks`, `top_coldspots`
- Schema migration v1→v2: performance indexes on `hotspot.score`, `risk.risk_score`, `cochange.file_b`

**rskim CLI:**
- New `temporal.rs` module: path normalization (project-root-first), standalone dispatch, enrichment, text/JSON formatters
- `ResolvedResult.temporal: Option<TemporalAnnotation>` — carries hotspot/risk scores to output layer; `#[serde(skip_serializing_if = "Option::is_none")]` keeps JSON clean when absent
- `format_text_output` appends temporal score suffix inline per result
- `parse_flags` validates mutual exclusion of `--hot`/`--cold`/`--risky`
- `--blast-radius` normalization resolves relative paths against project root first, CWD fallback

## Breaking Changes

None. `SearchQuery.file_filter` is `#[serde(skip)]` and defaults to `None`. `ResolvedResult.temporal` is `None` in all existing call sites. All 4 `ResolvedResult` struct literals with no temporal data updated to `temporal: None`.

## Reviewer Focus Areas

- **`temporal.rs:normalize_blast_radius_path`**: project-root-first resolution was added to fix blast-radius tests where the path exists under the temp root but CWD is elsewhere. The logic: join to root first, fallback to CWD if not found. Review that this is the right UX for symlinked repos.
[...]

PR #291, opened 2026-06-08
feat(#199): Wave 3g — skim search --ast structural pattern CLI

### LINKED ISSUE #199: Wave 3g: CLI integration (--ast "pattern")

dean0x:
## Summary

Wave 3g (#199): exposes the AST structural-query engine on `skim search` via `--ast`. Builds the AST n-gram index alongside the lexical index in one pipeline (same FileIds), wires `--ast` for standalone structural queries and text+AST intersection, and adds unified auto-refresh + self-heal across both indexes. Closes #199.

## What Shipped

- **AST index built alongside lexical** in the `consume` loop with the FileId-alignment invariant (every file — including non-tree-sitter JSON/YAML/TOML, oversized, parse-fail, and linearize-Err files — gets exactly one aligned `add_file_ngrams` call; empty-aligned entry on failure). Commit ordering: lexical build → AST build → manifest.save; AST build error ⟹ manifest NOT saved (old manifest left for retry). A desync between lexical and AST FileIds now aborts the build loudly before commit (fixes a latent silent-corruption path).

- **`--ast <pattern>` / `--ast=<pattern>`**: standalone structural queries (named patterns like `try-catch`, or containment like `"for_statement > await_expression"`). Text+AST intersection (`"auth" --ast try-catch`) applies the AST FileId set as a pure `file_filter` on the lexical query — lexical ranking, snippets, and line numbers preserved (interim file-set filter; composite cross-layer ranking deferred to #198).

- **Unified auto-refresh + self-heal**: lexical+AST rebuild atomically under one build lock at the same HEAD; a missing or below-FORMAT_VERSION AST index self-heals on the next query via a cheap 6-byte version probe.

- **AST-only output**: file-level (pattern name + description header, `path  score`, no line numbers); typed `--json` `{mode:"ast",pattern,description,total,results:[{path,score}]}`. Intersection reuses the lexical formatters (full snippets). `--help`, README, and CLAUDE.md document `--ast` with real examples.

## Validation Order (Deterministic, Test-Locked)

`--ast` + temporal (`--hot/--cold/--risky/--blast-radius`) → #202 error; single-node (`try_expression`) → #283 error; unknown pattern → lists available; all before opening the index.

## Deferrals (Filed Up Front)

- **#289** — Temporal populate path + join unified auto-refresh (documented hook point left in `ensure_indexes_fresh`; temporal search currently still no-ops — pre-existing since #257).
- **#290** — AST-index incremental build cache (`--update` currently re-extracts all files; no AST cache yet).
- Interim notes added to #198 (intersection ranking), #201 (line-span formatting), #202 (`--ast`+temporal).

## Measured Performance

(Corpus: this repo, 546 indexed files)

- **Build/rebuild**: 1.4s at 546 files → linearly extrapolates to ~25–26s at 10k (AC-P1 gate <30s: PASS).
- **Warm AST-only query**: ~30ms (gate <50ms: PASS). **Warm text+AST intersection**: ~31ms (gate <50ms: PASS).
- **On-disk**: AST index ~1.97 MB (skidx 176 KB + skpost 1.8 MB) — ~3% over the lexical footprint (combined lexical+AST budget; compression tracked in #273).
- **Incremental `--update`** after 1-file change: 1.68s full re-index (no AST cache yet — #290; report-only, not a gate).
- **Build memory**: both builders retain posting lists until `build()` (tens of MB at 10k files; chunked AST build tracked in #273).

## Tests

+34 AST CLI tests in `ast_tests.rs` (parse/validate, build & FileId alignment, standalone query, intersection, auto-refresh/self-heal, JSON contract). Full workspace suite green: **5,284 passed / 0 failed / 5 skipped**. `cargo clippy --all-features -- -D warnings` clean; `cargo fmt --check` clean.

## Quality Gates

- **Validator** (build/test/lint/fmt): ✅ PASS
- **Simplifier** (PF-004 widening + nesting cleanup): ✅ PASS
- **Scrutinizer** (caught+fixed the FileId-desync silent-corruption P0): ✅ PASS
- **Evaluator** (ALIGNED, 13/13 functional + 7/7 API ACs): ✅ PASS
- **Tester** (20/20 scenarios PASS): ✅ PASS

Not auto-merged (ADR-005 — hand-off only).


dean0x:
## PR Code Review — High-Confidence Findings

Based on 10 specialized reviews (security, architecture, performance, complexity, consistency, regression, testing, reliability, rust, documentation), here are the ≥80% confidence findings organized by priority.

### BLOCKING ISSUES (Fix before merge)

#### 1. Producer thread not joined on AST desync abort — panic-surfacing guarantee defeated
**File**: `crates/rskim/src/cmd/search/index.rs:293-303`
**Confidence**: 88% (reliability.md HIGH)

When `consume()` returns `Err` on the new ADR-006 abort path (line 293, `return Err(...)?`), the error unwinds immediately and the producer join at line 296-303 is never reached. This defeats the panic-surfacing invariant on this specific path.

**Impact**: If the producer thread panics after the abort point, the panic is swallowed instead of surfaced.

**Fix**: Capture the consume result, join the producer unconditionally, then propagate the error:
```rust
let consume_result = Self::consume(...);  // Don't use ? yet
producer_handle.join().map_err(...)?;     // Always join, catch panics first
let ConsumeResult { ... } = consume_result?;  // Then propagate consume error
```

---

#### 2. Text+`--ast` combined path fails loud instead of self-healing
**File**: `crates/rskim/src/cmd/search/mod.rs:570-595`
**Confidence**: 90% (regression.md HIGH) + 85% (performance.md HIGH)

The AST engine is opened at line 572 **before** `auto_refresh_if_stale` runs. The comment at line 571 ("auto_refresh_if_stale ran above") is factually wrong — the only refresh is inside `execute_query` at line 607, AFTER the AST work.

**Impact**:
- Pre-existing installs (lexical index present, no AST index yet) get "AST index not found" error and fail to self-heal, violating documented behavior
- The AST engine mmaps the old (stale) index, executes the query, then the rebuild discards the stale FileId set anyway (performance waste)
- Inconsistent with standalone path (`mod.rs:108-110`) which correctly refreshes first

**Fix**: Hoist `auto_refresh_if_stale(&root, &cache_dir, analytics)` before `open_ast_engine(...)` in `run_query`, matching the standalone pattern. Correct the stale comment at line 571.

---

#### 3. ADR-006 fail-loud abort path is completely untested
**File**: `crates/rskim/src/cmd/search/index.rs:494-508`
**Confidence**: 95% (testing.md HIGH)

The consume-loop abort (when `add_file_ngrams` fails after lexical success) is the central correctness fix of this PR. There is no test that exercises this path and asserts the observable outcomes: (1) `build_index` returns `Err`, (2) manifest is NOT overwritten (old manifest survives), (3) next `check_staleness` reports stale.

**Impact**: A future regression that swaps `return Err(...)` back to `continue` would silently corrupt the index with no test failure.

**Fix**: Add a regression test that injects an AST builder failure and asserts all three outcomes above. If the builder cannot be easily faked, extract a trait seam to make it testable.

---

### HIGH PRIORITY (Should Fix)

#### 4. Wasted AST materialization + discarded sort in intersection query
**File**: `crates/rskim/src/cmd/search/ast.rs:104-114`
**Confidence**: 88% (performance.md MEDIUM)

`resolve_ast_file_filter` sets `limit = usize::MAX` and calls `engine.search()`, which builds a `Vec<SearchResult>` (~64 bytes each, ~640KB for 10k files), sorts DESC, then extracts only FileIds and discards the order.

**Impact**: Unnecessary transient allocation + O(N log N) sort whose result is thrown away on every `text + --ast` query.

**Fix**: Add a lighter library entry point that returns just `Vec<FileId>` without constructing `SearchResult` or sorting.

---

#### 5. Dead `lang` parameter on `resolve_ast_file_filter`
**File**: `crates/rskim/src/cmd/search/ast.rs:98-108`
**Confidence**: 90% (architecture.md MEDIUM) + 80% (consistency.md MEDIUM)

The parameter is never used by the only caller (`mod.rs:573` passes `None`). The sibling `run_ast_standalone` never sets `lang` either. This is speculative generality and creates an asymmetry between the two AST paths.

**Fix**: Drop the parameter until a real `--lang` flag exists.

---

#### 6. `--stats` misreports git HEAD for pre-existing installs
**File**: `crates/rskim/src/cmd/search/staleness.rs:212-225`
**Confidence**: 88% (regression.md MEDIUM)

The AST self-heal probe short-circuits and returns `(NoStoredHead, None)` before loading the manifest. For installs with a lexical index + recorded HEAD but no AST index yet (every pre-PR install), `skim search --stats` now prints "git HEAD: (none)" and "stale (no HEAD recorded)" — the HEAD *is* recorded, only the AST index is missing.

**Fix**: Load the manifest before the AST-stale short-circuit and return it even when reporting `NoStoredHead`.

---

#### 7. Standalone AST path silently drops unmatched FileIds
**File**: `crates/rskim/src/cmd/search/ast.rs:155-163`
**Confidence**: 82% (consistency.md MEDIUM)

`run_ast_standalone` silently discards results whose FileId is out of range via `filter_map`. The combined path warns when filters resolve to fewer files than expected. This violates fail-loud and is inconsistent.

**Fix**: Emit a debug notice when `resolved.len() < results.len()` to surface the desync case consistently.

---

#### 8. `consume()` function complexity: 119 lines interleaving 4 control-flow policies
**File**: `crates/rskim/src/cmd/search/index.rs:424-543`
**Confidence**: 82% (complexity.md MEDIUM)

The loop interleaves (a) lexical fail-soft `continue`, (b) AST linearize/match, (c) fail-loud ADR-006 abort, (d) FileId overflow checks. The doc comment is 30+ lines to explain the control flow.

**Fix**: Extract AST n-gram derivation (lines 460-484) into `fn derive_ast_entry(...) -> (AstNgramSet, StructuralMetrics, u32)`. Loop becomes: lexical-add-or-continue → derive → ast-add-or-abort → advance.

---

### LOW PRIORITY (Long-term risk)

#### 9. Duplicated format-version constant (silent drift risk)
**File**: `crates/rskim-search/src/lib.rs:46-51` vs `crates/rskim-search/src/ast_index/store/format.rs:60`
**Confidence**: 88% (architecture.md LOW, but flagged as long-term risk)

`AST_INDEX_FORMAT_VERSION = 2` is newly declared at the crate root; `FORMAT_VERSION = 2` already exists inside `store/format.rs`. Two independent literals must be kept in sync by hand. When the format bumps to v3, drift is silent — the CLI self-heal stops detecting old versions.

**Fix**: Re-export the single source of truth or add a compile-time guard asserting equality.

---

### Summary

| Severity | Count | Blocking |
|----------|-------|----------|
| CRITICAL | 3 | YES (items 1-3) |
| HIGH | 4 | NO (should fix) |
| MEDIUM | 2 | NO (nice to have) |

---

**Attribution**: Synthesized by Claude Code devflow:git agent, integrating findings from 10 specialized reviews (security, architecture, performance, complexity, consistency, regression, testing, reliability, rust, documentation). Commit: 3aaa99f.

dean0x:
**🔴 BLOCKING: Standalone `--ast … --blast-radius …` silently drops `--blast-radius`**

The `--help` advertises `skim search --ast try-catch --blast-radius src/auth.rs` as "AST ∩ co-change", but when there's no text query, `run_ast_standalone` (line 128) never receives the `blast_radius` argument and performs no co-change intersection. The flag is silently discarded.

**False confidence**: The inline comment at mod.rs:101-105 claims "so the flag is honored," but only widened the dispatch guard for `--ast`—it didn't add `--blast-radius` to `run_ast_standalone`.

**Recommended fix (pick one)**:
- **(A) Implement it**: Thread `flags.blast_radius` into `run_ast_standalone`, resolve peer paths → `FileId`s (reuse `query.rs` blast resolution), and intersect with the AST `HashSet<FileId>` before formatting. This delivers the advertised "AST ∩ co-change."
- **(B) Fail loud**: Reject `--ast` + `--blast-radius` without a text query with a `#NNN`-tracked error (mirroring `#202`), and remove the `AST ∩ co-change` example from `--help` until it works.

Option A is the better fit since `query.rs` already has the intersection primitive.

See functional.md for full analysis.

dean0x:
**⚠️ SHOULD-FIX (90%): Regression test gives false confidence**

The test `ast_plus_blast_radius_no_text_does_not_silently_drop_ast` asserts only exit-code success, never that the `--blast-radius` filter was applied. Its own comment waves it off: "the AST path still runs." The test passes precisely *because* `--blast-radius` is ignored—it locks in the broken behavior.

**Correct test**: Build a project with known co-change structure, run `--ast <pattern> --blast-radius <file>`, and assert the result set equals `AST(pattern) ∩ peers(file)` (strictly smaller than the unfiltered AST set when peers exclude some matches).

This is the same coverage gap the testing reviewer flagged for the three-way `(Some(blast), Some(ast))` branch at `query.rs:104`.

See functional.md and testing.md for full analysis.

dean0x:
**MEDIUM (88% confidence): Redundant double `auto_refresh_if_stale` on combined text+`--ast` path**

On a `text --ast <pattern>` query:
- `run_query` calls `auto_refresh_if_stale` at mod.rs:617 (to refresh before opening the AST engine)
- `execute_query` calls `auto_refresh_if_stale` again at query.rs:56

The second call short-circuits at `StalenessCheck::Current` (no rebuild), but still re-runs `check_staleness`—a manifest load + git HEAD read—even though the index is already fresh. The first manifest is discarded; the second is loaded fresh.

**Impact**: Redundant manifest parse + git HEAD read per combined query. At 10k files, the second manifest parse is the dominant per-query cost (the measured ~31ms intersection).

**Fix**: Hoist the single refresh + manifest load to the `run_query` boundary, pass the loaded `FileManifest` into `execute_query` so it skips the internal refresh. The standalone AST path already does this correctly (mod.rs:115).

**Raised by**: architecture, performance, regression, and rust reviews. One consolidated comment here.

dean0x:
**HIGH (90% confidence): Four tests leak index artifacts into global cache**

These tests drive `run()` with `--root <tempdir>` but **no cache override**:
- `ast_tests.rs:496` (`intersection_disjoint_text_and_ast_returns_empty_exit_0`)
- `ast_tests.rs:901` (`ast_plus_blast_radius_no_text_does_not_silently_drop_ast`)
- `ast_tests.rs:201` (`run_ast_plus_hot_returns_202_error` — via `run()` which resolves cache)

The cache key is hashed from the tempdir path (unique per test), so **no cross-test collision**, but the index is written to the real `~/.cache/skim/search/` and **never cleaned up** — orphaned directories accumulate across every CI run.

**Why not flaky**: Each tempdir is unique → unique cache key → no state-leakage race. Impact is hygiene/test-isolation, not non-determinism.

**Fix**: Route these through an isolated cache like the hermetic tests (Groups 7/8/10 in `ast_tests.rs:636,682,733,844`). Either:
- (a) Set `SKIM_CACHE_DIR` to a tempdir for the test duration (guarded for parallel safety), or
- (b) Prefer the hermetic pattern: use lower-level entry points (`build_project_index(project, cache)` + `run_ast_standalone`/`execute_query` with explicit `cache_dir`) rather than top-level `run()`.

See testing.md for full analysis.

dean0x:
**MEDIUM (85% confidence): Avoidable `ast.clone()` of the matched-FileId HashSet**

In `query.rs:117`, the `(None, Some(ast))` match arm does:
```rust
sq.file_filter = Some(ast.clone())
```

This deep-clones the entire `HashSet<FileId>` (which can hold tens of thousands of entries for broad patterns like `try-catch` on a large corpus). `config.ast_file_ids` is owned by the `QueryConfig` and is not used again after this point, so the clone purely satisfies borrow rules.

**Fix**: Take `ast_file_ids` by value into `execute_query` (e.g. accept `QueryConfig` by value, or move `ast_file_ids` out via destructuring), so the set is moved into `sq.file_filter` instead of cloned.

One extra heap allocation + full copy per query on the AST-only path; bounded by match-set size.

See performance.md for full analysis.

dean0x:
**MEDIUM (82% confidence): Build lock acquisition is an unbounded blocking wait**

At `index.rs:178-180`, `lock_file.lock()` blocks indefinitely if another process holds the exclusive advisory lock. Every query path can reach this via `auto_refresh_if_stale → build_index`. If a peer `skim` process crashes while holding the lock, or a stale-but-live process hangs mid-build on a huge repo, an interactive `skim search` invocation blocks with no upper bound and no user-visible reason.

The reliability Iron Law requires every wait to be bounded—this is the one unbounded wait in the PR.

**Impact**: Silent stall with no diagnostic. Correctness is preserved (the lock guarantees serialization).

**Fix**: Prefer a bounded acquisition—`try_lock()` in a fixed-iteration retry loop:
```rust
const LOCK_MAX_WAIT: Duration = Duration::from_secs(120);
const LOCK_POLL: Duration = Duration::from_millis(200);
// … try_lock with timeout and clear diagnostic
```

At minimum, emit a one-line "waiting for concurrent build to finish…" notice to stderr before blocking so the stall is explained.

See reliability.md for full analysis and pseudocode.

dean0x:
**MEDIUM (85% confidence): `parse_flags` uses three different "consume next token" idioms**

At `mod.rs:297-377`, the `while i < args.len()` loop body matches ~12 flag arms, several consuming a following token with different patterns:
- `--limit` advances inline via `i += 1` (line 318)
- `--root` / `--ast` use the `consumed` return from `take_flag_value`
- `--blast-radius` uses the `consumed_next` return from `parse_temporal_flag`

Three different idioms for the same "consume next token" concern increases the chance a future arm gets the off-by-one wrong. The ordering and token-advance behavior is test-locked (per KNOWLEDGE), so a subtle regression is the most likely future failure mode.

**Fix** (non-breaking): Route `--limit` through `take_flag_value` too, so every value-flag uses one idiom. This is a readability consolidation only—arm order and semantics stay identical, locked tests continue to pass.

Alternatively: add a one-line comment at each manual `i += 1` site documenting the advance.

See complexity.md for full analysis.

dean0x:
**MEDIUM (82% confidence): TOCTOU between build-lock release and `open_ast_engine`**

At `mod.rs:617-618`, on the text+AST path:
- `auto_refresh_if_stale` acquires `.skim-build.lock`, rebuilds if needed, **releases the lock**
- `open_ast_engine` then opens `ast_index.skidx` with **no lock held**

A second `skim search` process could begin a rebuild in the gap and be mid-write when this process mmaps the file.

**Impact**: Under the local single-user threat model this is **benign**, not a vulnerability: index writes use `atomic_write` (temp + `sync_all` + rename), so readers see either the old coherent file or the new one—never torn writes. The reader validates size and CRC32 on `open()` and returns `IndexCorrupted` rather than reading garbage. But the load-bearing invariant (atomic-rename + validate guarantee) is implicit.

**Fix**: No code change required for correctness. Optionally add a one-line comment at `mod.rs:618` noting that lock-free `open_ast_engine` is safe *because* of atomic-rename + CRC/size validation in `AstIndexReader::open`, so a future refactor doesn't weaken either side. (avoids PF-002—make the implicit invariant explicit.)

See security.md for full analysis.

dean0x:
## Summary: PR #291 Code Review

**Overall Posture**: The 11 dimensional reviews (architecture, performance, complexity, consistency, regression, testing, reliability, rust, security, documentation, and cross-cutting functional) are **APPROVED / APPROVED_WITH_CONDITIONS** with **one blocking functional bug**.

### Inline Comments Posted: 8

1. **BLOCKING (95%)**: Standalone `--ast … --blast-radius …` silently drops `--blast-radius` (functional.md)
2. **SHOULD-FIX (90%)**: Regression test `ast_plus_blast_radius_no_text_does_not_silently_drop_ast` gives false confidence (functional.md)
3. **MEDIUM (88%)**: Redundant double `auto_refresh_if_stale` on combined text+`--ast` path (architecture, performance, regression, rust)
4. **HIGH (90%)**: Four tests leak index artifacts into global cache (testing.md)
5. **MEDIUM (85%)**: Avoidable `ast.clone()` of matched-FileId HashSet (performance.md)
6. **MEDIUM (82%)**: Build lock acquisition is unbounded blocking wait (reliability.md)
7. **MEDIUM (85%)**: `parse_flags` uses three different "consume next token" idioms (complexity.md)
8. **MEDIUM (82%)**: TOCTOU between build-lock release and `open_ast_engine` (security.md)

### 60-79% Suggestions Consolidated (not posted as inline)

**Lower-confidence findings for team consideration:**

- **Double `parse_ast_query` per `--ast` invocation** (rust, complexity, 65-72% confidence): pattern is parsed in `validate_ast_pattern`, then again in `resolve_ast_file_filter` or `run_ast_standalone`. Parse is cheap and re-validation is documented as intentional/idempotent. Minor DRY observation.
- **`take_flag_value` edge case with flag-like next token** (rust, 68%): `skim search --ast --json` treats `--json` as the pattern string. Fails loud with "unknown pattern" rather than mis-parsing, so safe, but message is confusing. Mirrors pre-existing `--root` space-form behavior—consistency choice, not regression.
- **`run_query` swallows AST-engine errors as warning** (security, 65%): On text+AST path, an `Err` from `resolve_ast_file_filter` degrades to "no AST filter" with stderr warning, so results may appear broader than `--ast` intent. Standalone path fails loud instead. Consider documenting the divergence or making combined path also fail loud.
- **Per-file path-key normalization allocated twice** (performance, 85%): `replace('\\', "/")` runs once in producer and again in consumer. ~10k small-string allocations at 10k-file scale, negligible against 25-26s build budget but avoidable.
- **Manifest-parse memoization suggestion** (performance, 70%): even after fixing the double refresh, `read_git_head` does loose-ref + packed-refs file I/O on each staleness check. Memoizing HEAD per process invocation would remove repeated `.git` reads.
- **`run_ast_standalone` out-of-range FileId warning lacks aggregate cap** (reliability, 65%): On a desynced index, every out-of-range FileId emits a stderr line (bounded by default `--limit 20`, user-capped). Rare due to self-heal, but very large `--limit` on corrupt index would spam. Consider aggregate warning.
- **Stale KNOWLEDGE.md: `resolve_ast_file_filter` signature drift** (documentation, 96%): `.devflow/features/cmd-search/KNOWLEDGE.md:118` documents 3-arg signature + `SearchQuery`-based query that was refactored away in this PR. A maintainer will hit confusion. **Fix**: Update line 118 to match shipped 2-arg direct `search_ast` call. (Note: This may have uncommitted edits per background maintenance agent—verify/commit after review.)

---

### Dimension Summary

| Dimension | Recommendation | Score | Notes |
|-----------|---|---|---|
| Architecture | APPROVED | 9/10 | Layering clean, ADR-006 abort sound, intersection seam well-scoped. Two MEDIUM cohesion refinements on auto-refresh. |
| Performance | APPROVED_WITH_CONDITIONS | 8/10 | Algorithmically correct. One cost-scaling concern (duplicate manifest load) should be addressed before 10k-file gate. |
| Complexity | APPROVED | 8/10 | Code long in places but flat (nesting ≤3), well-decomposed, test-locked. Two MEDIUM readability consolidations. |
| Consistency | APPROVED | — | Flagged by orchestrator as APPROVED with no MEDIUM+ findings in changed lines. |
| Regression | APPROVED | 9/10 | Lexical-only path byte-for-byte preserved, self-heal on v1/missing works, FileId-alignment sound. One MEDIUM double-refresh micro-cost. |
| Testing | APPROVED_WITH_CONDITIONS | 8/10 | Behavior-focused, high coverage on invariants. One HIGH hygiene issue (global-cache leaks), one MEDIUM coverage gap (three-way intersection test). |
| Reliability | APPROVED_WITH_CONDITIONS | 9/10 | Producer-thread lifecycle, lock release, overflow handling correct. One MEDIUM gap: unbounded blocking `.lock()` wait (no timeout). |
| Rust | APPROVED | 9 | Idiomatic ownership/borrowing, exhaustive matches, no unwrap on fallible I/O, PF-004 widening applied. |
| Security | APPROVED | 9/10 | No exploitable defect. One MEDIUM TOCTOU (benign under local threat model thanks to atomic-rename + validate). |
| Documentation | APPROVED_WITH_CONDITIONS | 9/10 | User-facing docs accurate, examples valid/test-backed, deferral tickets real. One MEDIUM stale KNOWLEDGE.md signature (may have uncommitted edits—verify). |
| Functional | REQUEST_CHANGES | — | **One blocking functional bug** (95%): standalone `--ast … --blast-radius …` silently drops the flag, violating the fail-loud principle and the advertised `AST ∩ co-change` feature. Also one should-fix false-confidence test (90%). |

---

### Next Steps for Author

1. **Fix the blocking functional bug** (standalone `--ast --blast-radius` intersection): Implement it (option A: thread blast into `run_ast_standalone`, reuse query.rs intersection) or fail loud (option B: reject the combination until implemented, file follow-up issue).
2. **Address the test-hygiene leak**: Route the four `run()` tests through isolated cache or use hermetic lower-level API.
3. **Consider the MEDIUM findings** (especially the double auto-refresh and unbounded lock wait) for prioritization — none block merge but several improve correctness/maintainability/reliability.
4. **Verify/commit the stale KNOWLEDGE.md line** (or confirm it has been addressed by background maintenance).

---

**Reviewer Notes**: 
- All 11 dimension reviewers provided substantive, well-reasoned findings. The cross-cutting functional review (orchestrator) identified the blocking feature-drop pattern that the dispatch-layer specialists did not own individually.
- Deduplication of the "double auto-refresh" finding (raised by 4 reviewers) consolidated into one comment to avoid repetition.
- The PR ships with comprehensive test coverage, correct error handling, and sound architectural layering. The findings are mostly refinements and one functional correctness bug that the advertised `--help` text promises.

---

**PR URL**: https://github.com/dean0x/skim/pull/291

dean0x:
## Blast-Radius Resolution: Duplication Across Three Dispatch Paths

**Issue**: The blast-radius → FileId resolution pipeline is duplicated across three locations, introducing DRY and single-responsibility violations:

- **`ast.rs:178-196`** (new, `run_ast_standalone`) — resolves paths → FileIds verbatim
- **`query.rs:111-135`** (existing, `execute_query_with_manifest`) — identical path→FileId loop
- **`mod.rs:536-573`** (existing, `resolve_blast_radius_filter`) — owns the DB/path resolution half

**Problem**: The duplication has already drifted — the "no temporal data" warning behavior differs between copies (`ast.rs:160-165` always emits plain stderr, while `mod.rs:547-559` is JSON-aware under `--json`). Any future change to blast-radius semantics must now be applied in three places or risk silent divergence.

**Recommended Fix**: Extract a single `resolve_blast_radius_file_ids()` helper in `temporal.rs` (which already owns `normalize_blast_radius_path` + `cochange_partner_paths`) that returns `Option<HashSet<FileId>>`. This:
- Centralizes the path→FileId pipeline (eliminates duplication)
- Resolves the JSON-awareness divergence (one warning code path)
- Reduces `run_ast_standalone` from 8 params to ~5 (removes the `#[allow(clippy::too_many_arguments)]`)
- Eliminates the `cache_dir` being passed twice

This change is in-scope for cycle-2 since the standalone path was newly added in this PR.

**Confidence**: 88-90% (architecture, complexity, consistency reviewers all flagged)

dean0x:
## Wasted Clone on the Intersection Arm (query.rs:138)

**Issue**: The upfront `let ast_file_ids = config.ast_file_ids.clone();` clones the AST FileId set unconditionally, but in the `(Some(blast), Some(ast))` intersection arm, the cloned set is only used for read-only `contains()` checks. The owned value is never moved into `sq.file_filter` — the intersection is a fresh HashSet.

**Impact**: One O(n) allocation + copy of the AST FileId set (bounded by 50k files) on every combined-flag query. The prior code used `config.ast_file_ids.as_ref()` and cloned nothing on the intersection arm.

**Recommended Fix**:
```rust
match (blast_file_ids, config.ast_file_ids.as_ref()) {
    (Some(blast), Some(ast)) => {
        // ast borrowed for contains() — no clone needed
        let intersection: HashSet<_> =
            blast.iter().filter(|id| ast.contains(*id)).copied().collect();
        sq.file_filter = Some(intersection);
    }
    (None, Some(ast)) => sq.file_filter = Some(ast.clone()), // only clone where needed
    // ...
}
```

This restores zero-clone on the intersection arm while keeping the single clone on the AST-only arm.

**Confidence**: 88% (performance reviewer flagged as regression from cycle-1 hoist)

dean0x:
## Test Assertion Decoupled from Production Output (ast_tests.rs:1140)

**Issue**: The `ast_blast_radius_intersection_is_applied_not_silently_dropped` test at line 1140 is the headline regression guard for the original PF-006 bug (standalone `--ast … --blast-radius …` silently dropped `--blast-radius`). However:

1. The test calls `run_ast_standalone()` (line 1206) and asserts only its exit code (`ExitCode::SUCCESS`)
2. The strict-subset assertion proving the intersection (lines 1248–1261) is computed from a **test-local re-derivation** that calls the library function independently, not from the production dispatch's actual output

**Problem**: If `run_ast_standalone` regressed to silently dropping `--blast-radius`, the test would still pass because the re-derived intersection is independent of the production code path.

**Recommended Fix**: Make the production intersection observable:

**Option 1** (preferred): Have `run_ast_standalone` return the resolved `Vec<AstResult>` or filtered `HashSet<FileId>` alongside (or instead of) the bare `ExitCode`, and assert the returned set equals the expected results.

**Option 2**: Thread a `&mut impl Write` into `run_ast_standalone` (it already constructs `BufWriter(stdout)` at line 242) so the test can pass a `Vec<u8>` buffer, then assert the captured text contains the expected file and not the excluded file. This mirrors how the formatter tests already use an injected writer.

Either way, the strict-subset assertion should be derived from **production output**, not from a test-local re-derivation.

**Confidence**: 88% (testing reviewer flagged; applies PF-007 avoid-exit-0-only-tests principle)

dean0x:
## Summary: PR #291 Code Review — Cycle 2 (Incremental)

**Status**: ZERO BLOCKING ISSUES — Merge recommendation: **APPROVE pending duplication cleanup**

This is an incremental cycle-2 review of PR #291 with zero correctness defects. All 10 specialist reviewers (architecture, complexity, consistency, performance, Rust, testing, security, reliability, documentation, regression) returned either **APPROVED** or **APPROVED_WITH_CONDITIONS**. The changes correctly implement the `--ast` + `--blast-radius` intersection on the standalone path, honor PF-006 and ADR-006, and pass 5,213 tests.

---

## Cross-Cycle Verification

✅ **All cycle-1 fixes verified sound** (architecture, performance, security, reliability reviewers each re-verified):
- Bounded build-lock (200ms poll, 120s deadline, `TryLockError` exhaustive match) — correct
- Producer-join happens-before before error propagation — correct
- `u32::try_from` widening on all FileId construction paths — present and applied
- Single `auto_refresh_if_stale` on text+`--ast` path — verified no double-refresh

✅ **No cycle-1 FALSE_POSITIVEs re-introduced** — KNOWLEDGE.md stale-signature and `manifest_count != file_count` guard both confirmed correct in current code; not re-raised.

✅ **No cycle-1 fixed issues reverted** — all 9 cycle-1 fixes remain in place and unregressed.

---

## Blocking Findings

**CRITICAL**: 0  
**HIGH**: 0

---

## Should-Fix Findings (High Confidence, Non-Blocking)

### 1. Blast-Radius Resolution: Duplication (88-90% confidence)
**Files**: `ast.rs:178-196`, `query.rs:111-135`, `mod.rs:536-573`  
**Severity**: MEDIUM (arch, complexity, consistency impact)  
**Status**: 3 inline comments created

The blast-radius → FileId resolution is duplicated across three dispatch paths. The duplication has already drifted (DB-absent JSON-mode warning differs between `ast.rs:160-165` and `mod.rs:547-559`). Extract a single `resolve_blast_radius_file_ids()` helper in `temporal.rs` to consolidate the pipeline. This also resolves two companion findings:
- Reduces `run_ast_standalone` parameter count from 8 to ~5 (removes `#[allow(clippy::too_many_arguments)]`)
- Eliminates `cache_dir` being passed twice to `run_ast_standalone`

**Condition for merge**: Extract the shared resolver so the path→FileId pipeline lives in one place.

---

### 2. Unconditional Clone on Intersection Arm (88% confidence)
**File**: `query.rs:138`  
**Severity**: MEDIUM (performance regression, minor allocation)  
**Status**: 1 inline comment created

The cycle-1 clone-hoist replaced per-arm clones with an unconditional upfront clone. On the `(Some(blast), Some(ast))` intersection arm, the cloned AST set is only read via `contains()` and never moved. Restore the match-on-borrow pattern: clone only in the `(None, Some(ast))` arm that needs ownership.

---

### 3. Test Assertion Decoupled from Production (88% confidence)
**File**: `ast_tests.rs:1140`  
**Severity**: MEDIUM (testing coverage)  
**Status**: 1 inline comment created

The headline regression test (`ast_blast_radius_intersection_is_applied_not_silently_dropped`) asserts the intersection via a test-local re-derivation, not production output. If the bug regressed, the test would still pass. Make `run_ast_standalone` output observable (return the resolved set or inject a writer for captured output) so assertions derive from production behavior, not re-implementation.

---

## Lower-Confidence Suggestions (Summary Only, Not Inline)

**60–80% confidence** findings from the review specialist reports:

- **Performance (72%)**: `sorted_paths()` allocated twice in `run_ast_standalone` (lines 179 and 219). Hoist once and reuse, matching the existing pattern in `query.rs:103`.
- **Rust (72%, 68%)**: Redundant `sorted_paths()` alloc on standalone branch; `parse_ast_query` invoked twice on standalone path (validate + derive). Both cosmetic/efficiency; low priority.
- **Consistency (62%)**: Double-space-after-period in new `index.rs` comments (style only, not enforced).
- **Security (62%)**: Blast-radius error messages echo raw `--blast-radius` path + canonicalized root to stderr. Pre-existing, benign in local-CLI threat model.
- **Reliability (82%)**: Build-lock manual recovery on deadline expiry is manual only (pre-existing design; bounded lock is correct). Informational.

**None of these block merge.** They are tracked in the specialist reports for future cleanup or explicit deferral.

---

## Architecture & Design Assessment

| Category | CRITICAL | HIGH | MEDIUM | Score | Recommendation |
|----------|----------|------|--------|-------|-----------------|
| Architecture | 0 | 2 | 2 | 7/10 | APPROVED_WITH_CONDITIONS |
| Complexity | 0 | 0 | 2 | 8/10 | APPROVED_WITH_CONDITIONS |
| Consistency | 0 | 0 | 3 | 8/10 | APPROVED_WITH_CONDITIONS |
| Performance | 0 | 0 | 3 | 8/10 | APPROVED_WITH_CONDITIONS |
| Rust | 0 | 0 | 0 | 9/10 | APPROVED |
| Testing | 0 | 0 | 1 | 8/10 | APPROVED_WITH_CONDITIONS |
| Security | 0 | 0 | 0 | 9/10 | APPROVED |
| Reliability | 0 | 0 | 0 | 9/10 | APPROVED |

**Unanimous Finding**: The three MEDIUM findings (duplication, clone regression, test decoupling) are all addressable in one coordinated refactor: extract the shared blast-radius resolver, fix the clone pattern in query.rs, and make `run_ast_standalone` output observable. None block merge; all are in-scope cycle-2 changes (the standalone path was newly added).

---

## Merge Readiness

✅ Zero blocking issues  
✅ All tests passing (5,213)  
✅ All cycle-1 fixes verified sound  
✅ No regressions introduced  
✅ Security, reliability, Rust all APPROVED  
⚠️ Three MEDIUM non-blocking conditions (see above)

**Merge Recommendation**: **APPROVED** — Merge now and file fast-follow tickets for the three MEDIUM conditions, OR **APPROVED_WITH_CONDITIONS** — Apply the duplication cleanup before merge (addresses all three MEDIUM findings at once, as the primary is the driver for the other two).

---

## Cross-Cycle Context

This review continues a multi-cycle feature implementation for PR #291 (Wave 3g, skim search `--ast` structural patterns with `--blast-radius` intersection on standalone dispatch). Cycle 1 introduced the standalone intersection fix and supporting commits; this cycle verifies correctness under incremental change and surfaces structural duplications that emerged from the initial implementation.

**Claude Code Review Harness — Cycle 2 — 2026-06-09**

Issue #317, opened 2026-06-11
tech-debt: command wrappers truncate output instead of compressing (compress, never truncate)


dean0x:
## Summary

Several skim command wrappers TRUNCATE tool output (hard line/row/char caps, tail-N) rather than COMPRESSING it the way skim's core code/prose transforms do. Truncation silently drops content an agent may need — e.g. the `gh issue view` body cap that motivated the gh output transparency work. This issue tracks the broader principle: **compress, never silently truncate.** Where skim must bound output, it should route the content through a real compressor (or be loud about what it dropped), not lop off the tail.

## Two Philosophies

- **Compress (preferred):** reduce tokens while preserving structure/signal — what skim's tree-sitter/markdown transforms already do. Reversible-in-spirit: the reader still sees what exists.
- **Truncate (current, in wrappers):** hard-cut after N lines/rows/chars. Cheap but lossy and silent; the reader cannot tell whether the cut hid the important part.

## Group A — Automatic Truncation to FIX (Silent, Content-Bearing)

Inventory (verify each file:line during implementation — these are from an audit and may have drifted):

- gh issue/pr body: 10-line cap (`crates/rskim/src/cmd/infra/gh/shared.rs` `MAX_BODY_LINES`, ~:43)
- gh comments: count cap + 120-char preview (`gh/shared.rs` ~:48 / ~:244)
- gh run step detail cap (`gh/shared.rs` ~:51)
- gh list-item cap (`gh/shared.rs` ~:54)
- gh api items/strings caps (`crates/rskim/src/cmd/infra/gh/api.rs` ~:229 / ~:308)
- gh release assets/body caps (`crates/rskim/src/cmd/infra/gh/release_view.rs` ~:191 / ~:298)
- DB table: 100 rows / 40-char cells (`crates/rskim/src/cmd/.../output/canonical.rs` ~:691 / ~:694)
- curl/aws: 100 items / 20 keys caps
- file-wrapper "… and N more" elision
- log field caps (`log.rs` ~:352 / ~:367)
- **test-failure tail-50** (`crates/rskim/src/cmd/.../test/shared.rs` ~:470) — HIGH PRIORITY: a tail cut can hide the actual first error / panic location

## Group B — Opt-In / Bounded-By-Request Truncation (ACCEPTABLE, Keep)

- `--max-lines` / `--tokens` user-requested caps (AST-aware truncation)
- top-N reports (heatmap insights, search `--limit`) where N is the point of the command

[...]

PR #318, opened 2026-06-11
fix: gh output transparency — pass through --json/--jq/-q/--template/-t byte-faithfully (skip-list + handler gate)

dean0x:
## Summary

Two independent bugs caused `gh issue view 93 -q .body` and similar invocations with output-steering flags to be reformatted and truncated by skim instead of passed through with exact bytes. This PR fixes both layers.

## Root Cause (Both Layers)

**Layer 1 — Hook rewrite skip-list (`rules.rs`):** The `skip_if_flag_prefix` arrays for all gh rules only contained the long-form flags (`--jq`, `--template`, `--web`) but missed the short aliases (`-q`, `-t`, `-w`) and the `--json` flag entirely. This caused the hook to rewrite `gh issue view 93 -q .body` → `skim gh issue view 93 -q .body`, which then hit the handler and got reformatted.

**Layer 2 — Handler gate (`cmd/infra/gh/mod.rs`):** The wrapper/argv0 path and direct `skim gh …` invocations bypass the rewrite skip-list entirely. There was no output-steering check in the handler, so any invocation with `--json`, `--jq`, `-q`, `--template`, or `-t` would still be reformatted + truncated by the sub-parsers.

## Changes

**Part 1A — Skip-list fix (`rules.rs`):**
- Every gh rule that skips on a long-form flag now also skips on the short alias: `-q` wherever `--jq` skips, `-t` wherever `--template` skips, `-w` wherever `--web` skips
- `--json` added to every gh rule except `gh api` (which has no `--json` flag — responses are always JSON natively) and `gh run watch` (streaming TUI, no output steering)
- Added `test_gh_rules_output_flag_parity` — an invariant test that enforces these alias relationships going forward; prevents the original bug from regressing
- Updated the design-decision comment block documenting the `-w`=`--workflow` safety reasoning and the parity invariant

**Part 1B — Handler transparency gate (`cmd/infra/gh/shared.rs` + `mod.rs`):**
- New `GH_OUTPUT_STEERING_FLAGS` constant and `user_steers_output(args)` function in `shared.rs` using strict equality matching (mirrors the engine's `should_skip_by_flag`)
- Gate fires at the top of `run()` before the subcommand match: when an output-steering flag is present (before `--`), `run_raw_passthrough` is called for byte-faithful passthrough
- `--web`/`-w` intentionally excluded from the gate: `--web` emits no stdout to corrupt; `-w` is ambiguous (`--workflow` on `gh run list`)

## `-w` = `--workflow` Safety Reasoning

`gh run list` uses `-w` as shorthand for `--workflow` (a filter flag), NOT `--web`. Adding `-w` to that command's skip-list or gate would suppress compression for workflow-filtered runs, which is incorrect. The fix therefore only adds `-w` to skip-lists where `--web` is present (commands that genuinely open a browser), and excludes `-w` from the handler gate entirely.

## Glued-Flag Documented Limitation

Glued short-value forms like `-q.body` (no space) are not matched on either layer — consistent with the existing strict-match semantics across the codebase. An invocation like `gh issue view 93 -q.body` would still compress. This is low-severity (users can always write `-q .body` with a space) and matches the documented engine behavior.

## Behavior Change

- **`gh` invocations with `--json`, `--jq`/`-q`, `--template`/`-t`**: previously reformatted + truncated on both hook and handler paths; now pass through with exact bytes on both paths
- **Plain `gh issue view N`** (no output flag): still compresses as before — this is correct and by design
- Not a breaking API change

## Test Matrix

- 3 existing tests flipped: `test_rewrite_gh_json_rewrites` → `_skipped`, `test_rewrite_gh_pr_list_json_roundtrip` → `_skipped`, `test_would_rewrite_gh_pr_list_json_rewrites` → `_skips`
- New unit tests (rules.rs / tests.rs): all short-alias skip cases, `--json` skip, api skip exceptions, `-w`=workflow guard, compound/pipe cases
- New e2e tests (cli_e2e_rewrite.rs): `--suggest`/`"match":false` for all skipped variants; `"match":true` guards for cases that must still rewrite
- `test_gh_rules_output_flag_parity`: parity invariant enforced at the type/test level
- `EXPECTED_RULE_COUNT` unchanged — no rules added or removed
- Both engine strict-match tests remain green

## Reviewer Focus Areas

- `shared.rs`: `user_steers_output` strict-match semantics vs. the `--` separator handling
- `rules.rs`: the `gh run list` and `gh release list` rows intentionally have no `-w` — verify against the `-w`=`--workflow` reasoning above
- `mod.rs`: the gate placement (before subcommand match, before api args strip) and the `ctx` parameter is not needed by `run_raw_passthrough`

## Notes

Tech-debt tracking issue for the broader "compress, never truncate" principle: #317


dean0x:
**[HIGH · testing.md, 90% confidence]**

Layer 2 (handler gate → `run_raw_passthrough` byte-faithful passthrough) has NO behavioral test. Every new e2e test routes through `skim rewrite --suggest` and only exercises Layer 1 (the rewrite skip-list). The gate `if` block could be deleted and the entire suite would stay green (PF-007).

**Suggested fix**: Add a fake-gh-on-PATH behavioral test that:
- Places a fake `gh` binary on a temporary `PATH`
- Invokes `skim gh issue view 93 -q .body` with a known byte sequence piped to the fake gh
- Asserts that stdout is byte-identical (the gate fired and passed through raw)

The pattern already exists in `test_passthrough_forwards_raw_content` (cli_e2e_rewrite.rs:751-775) — copy that model.

**Why it matters**: PF-007 says "if a test would still pass with the feature deleted, it asserts nothing." That applies here — the gate's *wiring* (if steering flag → invoke `run_raw_passthrough`) is uncovered. A future edit deleting or reordering that `if` block would compress `-q .body` output and every existing test would stay green.

dean0x:
**[HIGH · architecture.md + consistency.md, 88% confidence]**

The two layers disagree on whether `--json` is a steering flag for `gh api` and `gh run watch`:
- **Gate** (lines 44, 54-60): flat, subcommand-blind scan — treats `--json` as steering for *every* `gh` subcommand
- **Rules** (rules.rs:1455): subcommand-aware; deliberately exempts `gh api` and `gh run watch` from `--json` membership

**The mismatch:**
- `test_gh_api_json_still_rewrites` asserts `gh api ... --json x` **rewrites** to `skim gh api`
- But once it reaches the handler at runtime, `user_steers_output(["api", "repos/o/r", "--json", "x"])` returns `true` and routes to `run_raw_passthrough` instead
- The two layers make opposite local decisions for the same input

**Impact**: Currently benign end-to-end (raw passthrough of `gh api --json x`, which gh would reject anyway, is harmless). But it's a latent correctness trap — any future steering flag valid on some subcommands and non-steering on others will be handled correctly by the rewrite layer and **incorrectly** by the gate.

**Suggested fix** (priority order):
1. Make the gate subcommand-aware: derive the steering set from the matched rule's `skip_if_flag_prefix` (filtered to steering flags) rather than `GH_OUTPUT_STEERING_FLAGS`, so `gh api` automatically excludes `--json`
2. At minimum: add a `gh api`/`gh run watch` `--json` carve-out in `user_steers_output` + a cross-layer parity test mirroring `test_gh_rules_output_flag_parity` so the layers cannot silently drift

dean0x:
## Review Summary: fix/gh-output-transparency (PR #318)

**Overall**: Strong review from 8 of 10 reviewers (APPROVED), scores 7–10. **Zero blocking issues.** Two high-confidence should-fix items flagged by multiple reviewers relate to the gate-vs-skip-list split; addressing them strengthens the architecture.

### Key Findings

**2 HIGH-confidence should-fix items** (both posted above as comments):
1. **testing.md (90%)**: Layer 2 handler gate lacks behavioral test covering `run_raw_passthrough` invocation — gate could be deleted and suite stays green (PF-007 gap)
2. **architecture.md + consistency.md (88%)**: Subcommand-blind gate vs. subcommand-aware skip-list — layers disagree on `--json` steering for `gh api` and `gh run watch`, creating latent correctness trap

**Medium-confidence maintainability items** (no blocking, mention for context):
- Steering-flag list duplicated across ~10 sites (const + 9 gh rules); should be a shared `&'static [&'static str]` const like `DOCKER_GLOBAL_FLAGS`
- `user_steers_output` duplicates the strict-match predicate from `user_has_flag` and `should_skip_by_flag`; could reuse `user_has_flag(before_dashdash, GH_OUTPUT_STEERING_FLAGS)`
- Dangling tracking-issue reference at mod.rs:89-90 — comment says "compress, never truncate tracking issue" with no number; it is #317

### Review Scores
| Reviewer | Score | Status |
|----------|-------|--------|
| testing | 7/10 | APPROVED_WITH_CONDITIONS |
| architecture | 7/10 | APPROVED_WITH_CONDITIONS |
| consistency | 8/10 | APPROVED_WITH_CONDITIONS |
| (6 others) | 8–10 | APPROVED |

### Notes
- Documented `-q.body` glued-form data-loss limitation (intentional unmatch on both layers) is acceptable and tracked
- `.devflow/` diff entries are auto-generated Dream maintenance artifacts, not review substance
- Layer 1 (hook/rewrite path) coverage is strong and PF-007-compliant; Layer 2 gate lacks behavioral test (see HIGH #1 above)

**Review artifacts**: `/Users/dean/Sandbox/skim-search/.devflow/docs/reviews/fix-gh-output-transparency/2026-06-12_1103/`

---

*Review comments created by Claude Code (devflow:git agent)*

Issue #321, opened 2026-06-12
infra parsers fold child stderr only in Passthrough tier — success-with-warnings loses stderr


dean0x:
Follow-up from #317 (Workstream 6, item 3).

After #317, *unexpected* non-zero exits raw-forward full stdout+stderr, and file/db wrappers forward child stderr verbatim in the compressed path. The remaining gap: infra parsers (curl, gh, docker, kubectl, aws, terraform, …) only surface child stderr when the parse degrades to Passthrough tier (via `combine_output`). A command that **succeeds (exit 0) but emits warnings on stderr** and parses at Full/Degraded tier silently drops those warnings.

Examples: `docker build` deprecation warnings, kubectl server-side warnings (`Warning: ...` on stderr with exit 0), terraform provider warnings.

Disposition to evaluate per-tool: forward stderr verbatim (as file/db now do), or fold stderr warnings into the compressed rendering. Blanket `forward_stderr=true` was deliberately NOT applied to infra in #317 because several infra tools write progress/noise to stderr (curl progress meter, docker pull layers) that compression exists to remove.

Issue #320, opened 2026-06-12
file wrappers: MAX_DISPLAY_ENTRIES=100 truncates df/du/env/find/ls/ps/wc output


dean0x:
Follow-up from #317 (Workstream 6, lower-priority sibling of the grep/rg fix).

The file-family wrappers (df, du, env, find, ls, ps, wc) cap rendered entries at `MAX_DISPLAY_ENTRIES = 100` with a visible footer. Unlike the (now fixed) grep/rg silent caps, this truncation is at least *visible*, but it still violates the "compress, never truncate" principle from #317: skim must never show less information than the raw tool, only re-encode it.

Disposition options:
- Remove the cap entirely (input is already bounded by the runner's 64 MiB cap, so no unbounded allocation).
- If a bound is kept for a specific tool, route it through the shared `elision_marker()` helper (exact counts + `SKIM_PASSTHROUGH=1` hint).

Same principle, lower frequency/impact than grep — agents rarely need >100 `ls` rows, but `find` results are diagnostics and should not be capped.

Issue #319, opened 2026-06-12
PATH wrappers: pipe segments and harness Grep tool double-interception


dean0x:
Follow-up from #317 (Workstream 5c known limitation).

When PATH wrappers are installed (`skim init --wrappers`), a pipeline like `git diff | grep TODO` becomes `skim git diff | skim grep TODO` — each segment is independently intercepted, so the grep consumer sees *compressed* producer output rather than raw output. The rewrite hook now refuses to rewrite piped commands (#317), but PATH wrappers operate per-process and cannot see the pipeline shape.

This is **not solvable process-locally**: agent harnesses pipe all stdio, so isatty heuristics would disable wrappers entirely.

Scope to investigate:
- Can the wrapper detect it is mid-pipeline at all (e.g. via parent shell introspection)? Likely no — document and decide.
- Should PATH wrappers intercept the harness Grep tool's underlying rg invocations? (Open question carried from #317.)
- Whether wrappers should auto-passthrough when stdin is a pipe from another skim wrapper (e.g. env-var breadcrumb `SKIM_IN_PIPELINE`).

Until resolved, the limitation is documented in `cmd/init/wrappers.rs` and README.

PR #322, opened 2026-06-12
fix(#317): compress, never truncate — wrapper transparency + rewrite-hook correctness

### LINKED ISSUE #317: tech-debt: command wrappers truncate output instead of compressing (compress, never truncate)

dean0x:
Closes #317.

Skim's contract is now structural: **wrappers may re-encode output, never show less than the raw tool; rewrites must reconstruct the command byte-faithfully or bail.** Seven commits, one per workstream — each lands green.

### 1. Loud elision helper (`output::elision_marker`)
Any genuinely unavoidable safety bound must state exact counts + the escape hatch: `[skim] N <unit> omitted (X of Y shown) — SKIM_PASSTHROUGH=1 for full output`.

### 2. grep/rg: every match, truthful labels
- `MAX_MATCHES_PER_FILE=5` / `MAX_FILES_SHOWN=50` deleted — grouping stays (re-encoding), truncation goes; `shown == total`, no `(showing K)`.
- `GrepArgs::scan` classifies the argv (value flags, `--`, clusters, GNU option permutation): single-file output is attributed to the real operand path — not `<stdin>` — and the `file:line:content` regex is skipped when grep could not have produced that format (kills misattribution).
- Unattributable lines abort structured parsing → lossless passthrough; `Binary file … matches` notices preserved.

### 3. Exit-disposition matrix (execution path)
- `classify_exit` runs on the raw `Option<i32>` before any default — signal kills are always unexpected.
- **Unexpected failure → raw forward**: full stdout+stderr byte-faithfully (before ANSI strip), `[skim] {tool} exited N; raw output (not compressed).`, zero-savings `raw` analytics tier. `grep pat /nonexistent` now shows `No such file` and exits 2.
- `expected_exit_codes` + `forward_stderr` added with **no defaults** — all 36 `ToolRunConfig` sites + 3 direct sites carry explicit audited values.
- Notice matrix: expected failure at Passthrough tier is silent (grep no-match = raw-grep parity); Full/Degraded keep the escape-hatch notice.
- `forward_stderr=true` (file+db families): success-with-warnings stderr is no longer dropped.

### 4. Test wrappers: failure detail always survives
- Stable-toolchain `cargo test` (the default path): `parse_failure_details` state machine over `---- <name> stdout/stderr ----` blocks attaches each failing test's captured output — `panicked at <file:line:col>` always survives (Addendum-2 repro is a fixture + regression test). `TestResult.context` safety net renders a `--- failure context ---` block when block parsing yields nothing.
- `failure_context_body`: section runs from the FIRST failure marker to EOF; ≤350 lines whole, else head(300) + loud elision + 50-line recap. go/pytest/vitest/playwright inherit.
- `MAX_ENTRIES=100` scrape cap removed (failing test names are diagnostics).
- `run_parsed_command_with_exit`: piped failing runs derive exit 1 instead of the fabricated stdin exit 0.

### 5. gh/db/curl/aws: Group A caps removed
- gh: full issue/PR bodies (was 10 lines), ALL comments in full (was 3 × 120 chars), every failed step, no list/checks `MAX_ITEMS`, no api array `take(50)`; api depth/string elision exact+loud; release markers aligned; streaming 64KiB cap now splits losslessly (+1 byte no longer dropped).
- db: `MAX_DB_ROWS=100` + 40-char cell chop removed — every row, full cells; loud marker only when the tool's own footer claims more rows than parsed.
- curl/aws: `MAX_ITEMS=100` removed (16MiB JSON gate already bounds input).
- Kept (Group C): `MAX_JSON_BYTES`, JSON depth limits, `MAX_SUITE_DEPTH`.

### 6. Rewrite hook: never corrupt a command
- **5a** `--session-id` accepted on all paths — every hook-rewritten `cat <file>` with session attribution used to error. New e2e test class executes the emitted command.
- **5b** `rewrite_would_corrupt` bails on newlines / heredocs / `$()` `${}` / backticks / unmatched quotes / lossy whitespace, on every rewrite path — fixes the 72-session / 180-failure multi-line `git commit` corruption. One-line commits still rewrite.
- **5c** Pipes wholly rewrite-free (user-approved): any top-level `|` passes through. Pipe-rewrite machinery deleted and the `exclude_pipe_source` field stripped from all 154 rule instances — `rules.rs` is **modified (−154 lines), not removed**; the rule table itself is intact and the skipped-specific-rule suppression behavior is pinned.
- **5d** `.d.ts`/`.d.mts`/`.d.cts`/`.pyi` → `--mode=structure` (pseudo strips declaration files to nothing); mixed lists bail.
- **5e** Session-id injection works on the structured parts vector — correct for `;` too.

### 7. Docs
CLAUDE.md design constraint; wrapper pipe limitation documented (`wrappers.rs`, README) and tracked in #319.

## Verification
- `cargo test --all-features` (46 targets), `cargo clippy -- -D warnings`, `cargo fmt --check` — all green.
- All 9 issue repros from the plan run against this clone's release build (grep counts/labels/errors/silence, cargo panic + exit code, heredoc hook bail, session-id round-trip exec, pipe passthrough, `.d.ts` structure mode).
- Hot path untouched: 1600-line transform at ~8ms (target <50ms).

## Follow-ups (filed up front)
#319 (wrapper pipeline double-interception), #320 (file-family display cap), #321 (infra stderr on success-with-warnings).


dean0x:
**Post-push self-review finding (fixed in c6ebec9):** an adversarial review pass on this branch found one HIGH correctness bug in the redirect handling that predates this PR but falls squarely in its byte-faithfulness contract:

`strip_segment_redirects` strips only *recognized* redirect forms and re-appends them at segment end, so a recognized redirect preceding an unrecognized `>file` got reordered past it:

```
cargo build 2>&1 >log.txt && cargo test     # stderr→terminal, stdout→log
→ skim cargo build >log.txt 2>&1 && …       # BOTH→log (fd routing flipped)
```

`rewrite_would_corrupt` now bails on that shape (recognized redirect followed anywhere by an unrecognized `>`-bearing token — deliberately coarse, over-bail-by-design). Safe orders (`>log 2>&1`, recognized-only combos) still rewrite. Covered by 4 new unit tests (including one pinning the underlying reorder so the guard is provably load-bearing) + 2 hook e2e tests. Full suite, clippy, fmt green.

dean0x:
## Code Review Summary

**10-reviewer pass: NO BLOCKING ISSUES**

All reviews completed green across security, architecture, complexity, consistency, regression, rust, testing, and documentation. Scores: 8–10/10 across reviewers. Test suite verified: `cargo test --workspace` 5432 pass / 0 fail. `cargo clippy --all-targets -- -D warnings` clean.

---

## Should-Fix Findings (7 items, all ≥80% confidence)

### 🔒 Security (2)

**hook.rs:277 — Watchdog atomic-write race (Confidence: 80%)**
- **Risk**: The watchdog thread calls `process::exit(0)` after 5s. If it fires while the main thread emits the hook-response JSON via `println!` (line 277), the agent could receive truncated JSON on stdout.
- **Impact**: Low probability, but silent partial-JSON is a correctness risk on the security-sensitive output path.
- **Fix**: Build the full `json_out` string and emit atomically as `write_all().flush()` before the watchdog window:
  ```rust
  let json_out = serde_json::to_string(&response)?;
  let stdout = std::io::stdout();
  let mut handle = stdout.lock();
  handle.write_all(json_out.as_bytes())?;
  handle.flush()?;
  ```

**grep.rs:187–193, 287 — Output misattribution (Confidence: 82%)**
- **Risk**: When grep runs on one non-`-` operand with no `-r`/`-H`, output lines are attributed to `file_operands[0]` (the argv-derived label) without verifying it matches grep's actual filesystem access.
- **Impact**: If the operand contains an unexpanded glob (edge cases the corrupt guard allows), the displayed file attribution can name a path that didn't produce the lines — output-integrity/misattribution concern. Bounded: compress-never-truncate is honored, no bytes dropped.
- **Fix**: Add a doc-comment to `single_unprefixed_target` noting that the label is argv-derived and NOT verified against grep's actual filesystem access, so downstream consumers should not treat it as authoritative. No code change strictly required.

---

### 🏗️ Architecture (1)

**pkg/mod.rs:126 — `forward_stderr` hardcoded (Confidence: 92%)**
- **Issue**: The PR states "thread `expected_exit_codes` + `forward_stderr` with NO defaults" across ToolRunConfig. This holds for most families, but `run_pkg_subcommand` hardcodes `forward_stderr: false` (line 126) while exposing `expected_exit_codes` explicitly.
- **Impact**: One of two #317 fields is silent-defaulted; defeats the "explicit audit every site" invariant. Low runtime risk (pkg genuinely wants `false`), but architectural inconsistency.
- **Fix (option b, cheaper)**: Document in `PkgSubcommandConfig` doc-comment that `forward_stderr: false` is a deliberate family-level constant, making the asymmetry intentional and discoverable.

---

### 🎯 Consistency (1)

**gh/api.rs:269, 318, 329 — Elision format divergence (Confidence: 85%)**
- **Issue**: This PR centralized on `output::elision_marker` for all truncation notices. However, `compact_json_value` hand-rolls depth/string elision notices instead of using the shared helper. Compare to sibling `curl.rs:592` which handles the identical "JSON object exceeds key cap" case WITH the shared helper.
- **Impact**: Two JSON-compaction wrappers, two different elision idioms — inconsistent with the PR's uniformity goal (#317).
- **Fix**: Express bounds via `elision_marker` where a count is available (e.g. `elision_marker(map.len(), total, "object keys")`), so the `[skim] … SKIM_PASSTHROUGH=1` suffix is identical across all sites.

---

### 📊 Complexity (1)

**grep.rs:100–183 — `GrepArgs::scan` approach limit (Confidence: 82%)**
- **Issue**: ~83 lines, cyclomatic complexity 12–15 (borderline HIGH). Handles three distinct branches (positionals, long-flag 14-entry table, short-cluster `-rn`/`-A3`) with 4–5 nesting depth in short-cluster arm.
- **Mitigation**: Exhaustive unit tests + clear doc-comments. Function is correct.
- **Optional low-risk refactor**: Extract short-flag cluster loop into `fn scan_short_cluster(...)` and long-flag dispatch into `fn scan_long_flag(...)`, removing duplicated `i += 1` logic and dropping each piece below 30 lines / depth 3.

---

### 🦀 Rust (1)

**streaming.rs:133–142 — UTF-8 lossy at chunk boundary (Confidence: 85%)**
- **Issue**: `read_line_lossy` splits overlong lines into 64 KiB chunks and decodes each with `from_utf8_lossy`. When a multi-byte UTF-8 codepoint straddles the boundary, trailing partial bytes → `U+FFFD` + leading partial bytes of next chunk → `U+FFFD` — reassembled content is NOT byte-faithful (4-byte char can become 6 bytes).
- **Doc/reality gap**: Test name asserts "every byte of the raw line must survive across chunks" but only holds for ASCII.
- **Fix (low-risk option)**: Soften doc-comment and test names to "every byte of valid UTF-8 survives; multi-byte chars split at 64 KiB cap degrade to U+FFFD per AD-STR-6", matching the intentional lossy contract.

---

### 🧪 Testing (1)

**cargo.rs:487–524 — `parse_failure_details` coverage gap (Confidence: 88%)**
- **Issue**: State machine (block-header regex with lazy `.+?` for doctest names with spaces, multi-block append) has only one e2e test with a single-block fixture. The append-on-second-block branch, lazy quantifier, and bare `failures:` terminator are unexercised.
- **Impact**: Regression in any of those branches would pass the full suite green. By contrast, sibling `failure_context_body` is thoroughly unit-tested (8+ scenarios).
- **Fix**: Add `#[cfg(test)]` unit tests directly calling `parse_failure_details` with:
  - (a) Two-block same-test input (stdout + stderr blocks) asserting both bodies are concatenated
  - (b) Doctest-style header `---- src/lib.rs - module (line 10) stdout ----` asserting names with embedded spaces are captured

---

## Documentation Note

The PR description states "rewrite/rules.rs entirely removed (-154)" but the file was **modified, not deleted** (2216 → 2062 lines). The `exclude_pipe_source` field was stripped from the struct and all 154 rule instances; the rule table itself is intact. Consider correcting the PR body to avoid reviewer confusion.

---

## Recommendation

**APPROVED** with optional should-fix follow-ups. The 7 findings above are architectural/maintainability improvements, not blockers. Per ADR-001 ("if you see something, do something"), the author may choose to address them now or defer to follow-up PRs. The core #317 invariant (compress, never truncate) is well-implemented across all wrappers, tests are comprehensive (no regressions), and the security boundaries are sound.

---

*Reviewed by: security · architecture · complexity · consistency · regression · rust · testing · documentation*

*Standards applied: SOLID, PF-002, PF-003, PF-007, PF-008, ADR-001, AD-STR-6, feature-knowledge:build-parsers*

Issue #337, opened 2026-06-17
rewrite engine: full composed-command support (parse instead of bail)


dean0x:
## Context

The fix/rewrite-hook-falseneg PR adds a conservative safety bail: the rewrite engine passes a command through unchanged when, after trimming leading and trailing whitespace, it contains an interior newline (statement-merge corruption, where multiple statements were being merged into one rewritten line). That is a stopgap that trades compression for correctness on multi-line input.

## The deeper gap

The engine still lacks full composed-command support. It tokenizes on whitespace without being quote-aware, which leaves several known corruption and false-negative classes unhandled:

- Quote-aware tokenization: multi-space-in-quotes corruption (collapsing runs of spaces inside a quoted argument), and tabs as separators (audit item AD-RW-13).
- Multi-line parsing: today the engine bails on interior newlines rather than parsing each statement; a proper parser would compress each line.
- Heredocs and line continuations (trailing backslash at end of line).

## Direction

Replace bail-on-what-we-cannot-prove-round-trips with a real shell-aware tokenizer and parser that preserves byte-faithful semantics across quotes, tabs, multiple spaces, newlines, heredocs, and continuations, so these commands can be compressed safely instead of passed through raw.

## Acceptance (high level)

- Quote-aware tokenizer: spaces and tabs inside quotes preserved byte-for-byte.
- Multi-line commands parsed per-statement (no blanket newline bail).
- Heredoc and line-continuation bodies preserved verbatim.
- Round-trip property tests over the corruption classes above.

## Related

Follow-up to the interior-newline bail introduced in fix/rewrite-hook-falseneg (fix C).

PR #340, opened 2026-06-21
fix: rewrite-hook false-negatives — root-cause fixes A–F + recurrence guards

dean0x:
## Summary

Fixes a class of **silent false-negatives** in skim's command-rewrite hook. skim rewrites `grep` → `skim grep` (wrapping the real tool); three independent reviewer reports hit the same defect. Empirical reproduction against the built binary showed the trigger was **not** pattern characters (`<`, anchored `+`, quotes) but a small set of root-cause bugs that all produce the same dangerous outcome — output an agent reads as *"no matches / not present."*

Six fixes (A–F) on one branch, spanning the rewrite engine (A/C/E — the agent PreToolUse hook and the `skim rewrite` CLI) and the shared tool handlers that **both** interception surfaces run (B/D/F — the hook/CLI engine **and** the OS-level PATH wrappers). The rewrite engine and the wrapper surface dispatch differently; see `CLAUDE.md` → *Two interception surfaces*.

## Fixes

- **A — Flag-drop root cause + recurrence guard.** Rules baked a required flag into the match `prefix`, then `rewrite_to` replaced the whole prefix with `skim <tool>`, **dropping the flag** (`grep -rn` → `skim grep` lost `-rn` → non-recursive → *"Is a directory"* → false-negative; same for `grep -r`, `ls -la`, `ls -R`). Flags are now preserved in `rewrite_to`. **Generalization:** a structural guard test (`no_rule_silently_drops_prefix_flags`) **fails the build** if any rule drops a prefix flag that isn't on an explicit `WRAPPER_REINJECTS` allowlist (each entry citing the compensating wrapper fn). The guard immediately caught a **pre-existing** bug — `ruff format --check` dropped both `format` and `--check`, silently turning a dry-run into a destructive file rewrite — now fixed too.
- **B — Error visibility.** grep/rg/diff exit **1** (benign "no match"/"differs") no longer prints the misleading `[skim] compressed output (exit 1)` hint, while **still propagating the exit code**; exit **≥2** (a real error) forwards the tool's stderr to our stderr. (Surfacing stderr on error was already provided by #317's `forward_stderr`; this PR makes the hint exit-code-aware per tool via `should_emit_compressed_hint`.)
- **C — Composed-command safety.** Passthrough-unchanged when a command, after trimming trailing whitespace, contains an **interior newline** (statement-merge corruption: `git log -1 SHA⏎printf done` was merging into one command). Applied at **both** engine entry points — the agent hook **and** the `skim rewrite` CLI path (where `split_whitespace` had been silently merging statements). Trailing newlines still rewrite. The deeper composed-command parser gap is tracked in #337.
- **D — git show file-content.** `git show <ref>:<path>` now uses **pseudo** mode (preserves body logic) instead of structure mode (which collapsed bodies to `{...}` and hid the logic). The inflation guardrail still falls back to raw.
- **E — Pipe-source.** `git diff|log|show | grep …` leaves the source **raw** when piped (skim's reformatted lines would break the downstream filter). Enforced globally by the pipe short-circuit in `compound.rs`; misleading per-rule comments corrected.
- **F — Display wording.** `GREP: N matches in M files` — removed the misleading `(showing N)` (N was the *file* count; a reviewer read it as truncation). Truncation wording appears only when a cap actually truncated. **JSON envelope unchanged.**

## Recurrence prevention

1. **Structural guard test (A)** — CI fails if any rule drops a prefix flag.
2. **Error visibility (B)** — any residual failure becomes loud stderr, not an empty "no match."
3. **Bail-don't-corrupt (C)** — the engine passes through anything it can't provably round-trip.

## Testing

- **rskim: 3921 tests pass** (full `--no-fail-fast`); **rskim-core: 470 tests + 15 doctests pass**.
- `cargo clippy --all-targets -D warnings` clean; `cargo fmt --check` clean.
- **AC-A4 is mutation-proven:** the negative fixture drives the *real* shared `check_rule_for_flag_drop` helper — inverting the check turns **both** the guard and the fixture red.
- **Surfaces tested:** the `--hook` JSON path and the `skim rewrite` CLI are tested for flag preservation (A) and the B/D/F behavioral changes. The PATH-wrapper surface (`~/.skim/bin/<tool>` → skim binary) takes a separate dispatch path (`detect_argv0_dispatch` → `cmd::dispatch`) that **bypasses the rewrite engine** — it runs the tool handler directly with the original argv, so rewrite-path flag preservation (A/C/E) does not apply to it. It shares the B/D/F handler code with the tested paths but is not independently exercised by a wrapped-`argv[0]` test. See `CLAUDE.md` → *Two interception surfaces*.

## Scope (deliberately not in this PR)

No full multi-line/quote-aware parser, no engine-level per-rule flag opt-outs, no `--mode` flag on `git show` — all per the plan. Round-tripping constructs (`$()`, backticks, `${}`, `<()`, redirects, `&`, `<<<`, comments, env-prefixes, quoted pipes) are **not** bailed (verified to round-trip — bailing would lose compression for nothing).

## Tracking

- #337 — *rewrite engine: full composed-command support (parse instead of bail)* — the deeper quote-aware tokenization / multi-line / heredoc work.

### DIFF
--- a/.devflow/features/build-parsers/KNOWLEDGE.md
+++ b/.devflow/features/build-parsers/KNOWLEDGE.md
@@ -1,7 +1,7 @@
 ---
 feature: build-parsers
 name: Build Tool Output Parsers
-description: "Use when adding a new build tool parser, modifying cargo/tsc/make/gradle/maven compression, or debugging three-tier parse degradation for build commands. Keywords: build, cargo, tsc, make, gradle, maven, clippy, ParseResult, BuildResult, three-tier, NDJSON, flag injection, run_check, run_fmt, ChildGuard, indefinite, is_indefinite_command, expected_exit_codes, forward_stderr, ExitDisposition, classify_exit, run_parsed_command_with_exit, elision_marker, failure_context_body, parse_failure_details, compress-never-truncate."
+description: "Use when adding a new build tool parser, modifying cargo/tsc/make/gradle/maven compression, or debugging three-tier parse degradation for build commands. Keywords: build, cargo, tsc, make, gradle, maven, clippy, ParseResult, BuildResult, three-tier, NDJSON, flag injection, run_check, run_fmt, ChildGuard, indefinite, is_indefinite_command, expected_exit_codes, forward_stderr, ExitDisposition, classify_exit, run_parsed_command_with_exit, elision_marker, failure_context_body, parse_failure_details, compress-never-truncate, spawn_status_to_code, run_with_node_fallback, should_emit_compressed_hint, BENIGN_EXIT1_PROGRAMS, FileResult."
 category: component-patterns
 directories: [crates/rskim/src/cmd/build/, crates/rskim/src/cmd/]
 referencedFiles:
@@ -24,8 +24,8 @@ referencedFiles:
   - crates/rskim/src/cmd/test/cargo.rs
   - crates/rskim/src/cmd/test/shared.rs
 created: 2026-05-14
-updated: 2026-06-17
-version: 16
+updated: 2026-06-21
+version: 17
 ---
 
 # Build Tool Output Parsers
@@ -51,7 +51,7 @@ Rules that flow from this constraint:
 `cmd/mod.rs` was refactored to reduce complexity. Functionality previously inline in `mod.rs` is now split across dedicated submodules:
 
 - `cmd/dispatch.rs` — `dispatch()`, `run_raw_passthrough()`, and per-tool dispatcher helpers (`dispatch_cargo`, `dispatch_go`, `dispatch_swift`, `dispatch_dotnet`, `passthrough_subcmd`, `extract_subcmd`, `prepend_without`)
-- `cmd/execution.rs` — `OutputFormat`, `RunContext`, `ParsedCommandConfig<'_>`, `ToolRunConfig<'_>`, `run_tool<T>`, `run_parsed_command_with_mode`, `run_parsed_command_with_exit`, `format_analytics_label`, `combine_output`, `obtain_output`, `render_output<T>`, `record_and_report`, `passthrough_raw`
+- `cmd/execution.rs` — `OutputFormat`, `RunContext`, `ParsedCommandConfig<'_>`, `ToolRunConfig<'_>`, `run_tool<T>`, `run_parsed_command_with_mode`, `run_parsed_command_with_exit`, `format_analytics_label`, `combine_output`, `obtain_output`, `render_output<T>`, `record_and_report`, `passthrough_raw`, `RecordReport<'_>`, `should_emit_compressed_hint`, `BENIGN_EXIT1_PROGRAMS`
 - `cmd/security.rs` — `sanitize_for_display`, `scrub_db_args`, `scrub_infra_args`
 - `cmd/registry.rs` — `KNOWN_SUBCOMMANDS` (sorted, binary-searchable), `is_known_subcommand`, `is_meta_subcommand`, `wrapper_targets`
 - `cmd/test_utils.rs` — standalone `pub(crate)` module (compiled under `#[cfg(test)]` gate in `mod.rs`); canonical test helper source for all `cmd` subtree tests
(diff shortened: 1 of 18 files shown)


dean0x:
**File**: `crates/rskim/src/cmd/rewrite/rules.rs:17` (and lines 1820, 1837, 2188, 2195, 2205)

**Finding**: Hardcoded cross-file line-number citations create brittle structural coupling.

Four comments hardcode `"lines 543-545 of compound.rs"` and three cite `"line ~54 in gofmt.rs"`, `"lines ~92-94 in prettier.rs"`, `"lines ~94-97 in rustfmt.rs"`. These are accurate today, but they couple comments in one file to absolute line ranges in other files with no compiler or test enforcement. Any insertion above line 543 in `compound.rs` silently invalidates all four citations (PF-008 drift trap).

**Fix**: Cite the function name instead: `"enforced by has_pipe_operator in try_rewrite_compound (compound.rs)"` rather than line numbers. Symbol references survive line drift and are greppable. For the WRAPPER_REINJECTS entries, drop the line numbers entirely and rely on the function name + file context.

**Confidence**: 96% (architecture), 72% (consistency), 70% (documentation) — three reviewers flagged this as a maintenance hazard.

**Why it matters**: This is exactly the "drift trap" PF-008 warns about. A future maintainer reading a stale line ref will trust incorrect information and make incorrect decisions.

dean0x:
**File**: `crates/rskim/src/cmd/git/show.rs:794`

**Finding**: Fix D uses struct-literal syntax instead of the established `TransformConfig::with_mode()` constructor.

**Current (line 794-798)**:
```rust
let config = TransformConfig {
    mode: Mode::Pseudo,
    ..TransformConfig::default()
};
```

**Issue**: The codebase has a dedicated constructor `TransformConfig::with_mode(mode)` (defined at `rskim-core/src/types.rs:670`) used at ~20 call sites, including the sibling `git diff` command in `diff/render.rs:446`. Fix D is the only non-test production site constructing via struct literal—an inconsistency with the project's "match existing patterns — consistency beats cleverness" rule.

**Fix**:
```rust
let config = TransformConfig::with_mode(Mode::Pseudo);
```

This is behavior-identical but aligns `git show` with `git diff` and other call sites.

**Confidence**: 88% (consistency review). This is the line the PR changed (was `TransformConfig::default()` before), so it's categorized as Should-Fix.

dean0x:
**File**: `crates/rskim/src/cmd/execution.rs:262` (Fix B `should_emit_compressed_hint`)

**Finding**: Fix B's behavioral change has no end-to-end test for the path it actually modifies.

**The issue**: Fix B only alters behavior when a `BENIGN_EXIT1_PROGRAMS` tool (grep/rg/diff) exits with code 1 AND produced a Full/Degraded-tier compressed body. However:

- The pre-existing E2E `test_grep_no_match_exits_1_silently` tests the passthrough tier (empty body), which was already silent via the pre-existing `tier_name != "passthrough"` condition. It does **not** exercise the new `!is_benign_exit1` term.
- For `diff` (the clearest real case), the only E2E is `skim diff` (a git-wrapper), not the raw `diff` tool routed through `run_parsed_command_with_exit`.

**Impact**: The unit tests pin the decision function, but nothing proves the wired-up integration (`record_and_report` → `should_emit_compressed_hint` → returned `ExitCode`) suppresses the hint AND propagates exit 1 for a real differing-files `diff a b`.

**Fix**: Add an E2E that runs `diff` on two actual differing files through the skim binary, asserting:
- `.code(1)` — exit code is propagated
- stdout contains the compressed diff body (Full tier)
- stderr does NOT contain `"compressed output"`

**Confidence**: 88% (testing review). This is a MEDIUM-confidence finding, but it matters because the threading of `RecordReport::program` is now load-bearing.

dean0x:
**File**: `crates/rskim/src/cmd/git/show.rs:1483` (Fix D unit tests)

**Finding**: Fix D's unit tests assert a locally-constructed config, not the production path behavior (PF-007 shape).

**The issue**: The two new tests (`fix_d_file_content_transform_uses_pseudo_mode`, `fix_d_pseudo_mode_preserves_function_body`) assert that:
1. A locally-constructed `TransformConfig{mode: Pseudo}` has mode = Pseudo ✓
2. `rskim_core::transform(src, Rust, Mode::Pseudo)` keeps function bodies ✓

But neither drives **`run_show_file_content`** — the production function that actually sets the mode. These tests would **still pass** if line 794 were reverted to `TransformConfig::default()` (Structure mode), making them vacuous guards per PF-007.

**Impact**: A regression that silently reverts Fix D to Structure mode would not be caught by these tests.

**Fix**: Add an E2E that runs `skim git show HEAD:<a-real-supported-source-file>.rs` (or similar) and asserts the output contains body/logic tokens that Structure mode would strip, proving Pseudo is live. Alternatively, a unit test that directly calls `run_show_file_content` against a fixture and asserts the rendered output retains body content.

**Confidence**: 85% (testing review). This is a should-fix because it mirrors the PF-007 anti-pattern: a test that would still pass with the fix deleted.

dean0x:
## Summary: Unanimous APPROVE (0 Blocking Issues, 10 Reviews, Scores 8–9/10)

All 10 reviewers (security, architecture, performance, complexity, consistency, regression, testing, reliability, Rust, documentation) unanimously recommend **APPROVE**. Behavior changes are intentional fixes. No blocking issues found.

---

### Inline Comments Posted (4/6 eligible)

1. **Brittle cross-file line citations** (rules.rs) — cite function names, not line numbers (PF-008 drift trap)
2. **Fix D idiom deviation** (show.rs:794) — use `TransformConfig::with_mode()` not struct literal
3. **Fix B no real-tool E2E** (execution.rs:262) — add `diff a b` end-to-end test
4. **Fix D vacuous test** (show.rs:1483) — add production-path test through `run_show_file_content`

**Skipped (60-79% confidence)**: Two additional findings scored 60-79% and are listed below in "Non-Blocking Should-Fix" instead of inline.

---

### Non-Blocking Should-Fix & Observations

**MEDIUM (60-79% confidence, non-blocking):**

- **"Both surfaces verified" claim is partially false** (testing 82%): The PR description asserts that wrapped-tool (argv[0]=tool) and `--hook` JSON paths "behave identically for B/D/F + flag preservation." Only the `--hook` path is tested; no E2E exercises the wrapped argv[0] surface. Either add a wrapped-argv E2E or soften the prose claim (the wrapper path uses the same `try_rewrite`/handler code, so it is covered indirectly, but not directly tested).

- **WRAPPER_REINJECTS invariant not test-enforced** (architecture 82%): Each allowlist entry exempts a rule on the basis that a wrapper re-injects the flag, but the proof is a hand-maintained comment, not code. If a wrapper stops injecting, the allowlist silently lies and the guard becomes ineffective. **Candidate follow-up**: have each wrapper expose `const INJECTS: &[&str]` and assert in tests that every `WRAPPER_REINJECTS` entry's flag is actually in the cited handler's `INJECTS`.

- **Fix-A e2e predicates are non-discriminating** (regression 88%): Tests like `test_rewrite_grep_r` assert only `contains("skim grep")`, not the preserved flag itself. They pass whether or not `-r` appears in the output. The load-bearing guard is the structural `no_rule_silently_drops_prefix_flags` unit test (which correctly iterates all rules and would panic on regression), so this is non-blocking, but tightening the e2e assertions to `contains("skim grep -r TODO src/")` would close the e2e gap.

**LOW (60-79% confidence, informational):**

- **`prefix[1..]` indexing lacks empty-slice guard** (Rust 70%): `check_rule_for_flag_drop` at line 2129 does `prefix[1..]` which would panic if a rule ever has an empty prefix. No current rule has one, so unreachable today, but a defensive `.get(1..).unwrap_or(&[])` would future-proof the guard (matters because panics here only surface under `--all-targets` CI per PF-009).

- **CLI rewrite path scans input twice** (performance 72%): Fix C adds `command_needs_passthrough` before `collect_input_tokens`, which then re-splits the same args. This is the explicit CLI `skim rewrite` entry (one call per user invocation), NOT the hot `--hook` path. Impact on <50ms budget is negligible; flagging only for awareness. The correctness justification is sound.

- **Module-doc cross-file line citations** (documentation 70%): Same issue as inline comments #1 above — hardcoded line refs in rustdoc comments (e.g., "lines 543-545 of compound.rs") will silently drift. Consider citing the function name instead.

- **Wrapper path struct-field redundancy** (Rust 65%): Fix D builds a full `TransformConfig` then discards all fields except `mode`. Matches the pre-existing pattern in the same function (no regression), so style-only, non-blocking.

---

### Pre-Existing Issues (Out of Scope, Do Not Block This PR)

**NOTE (PF-008 observation, not a finding against this PR):**

Three reviewers independently encountered a **pre-existing rendering bug** unrelated to these fixes: skim's `git diff` handler duplicates changed lines (shows both `-old` and `+new` on the same display line) when the diff tier is Full. Reproducible in this PR's own binary AND independently on the diff path this PR does NOT modify. Recommendation: **file a separate GitHub issue** to track this existing behavior, but it does not block this PR since it's pre-existing and outside the scope of the rewrite-hook fixes.

---

### Coverage Verdict (Testing Aggregate)

| Aspect | Coverage | Status |
|--------|----------|--------|
| Fix A (flag preservation) | Structural guard + unit tests ✓; e2e predicates weak but guarded | PASS |
| Fix B (exit-1 hint suppression) | Unit tests ✓; no real-tool Full-tier E2E | PASS (unit solid; e2e gap noted) |
| Fix C (interior-newline bail) | All three coverage types (unit, e2e, both surfaces) ✓ | PASS |
| Fix D (Pseudo mode) | Unit only; production path not tested | PASS (unit tests cover intent; path test recommended) |
| Fix E (pipe passthrough docs) | Doc correction + E2E ✓ | PASS |
| Fix F (header rendering) | Unit + serde roundtrips ✓ | PASS |
| Two-entry-point invariant | All fixes wired correctly; Fix C/E verified on both surfaces ✓ | PASS |
| PF-007 sweep (vacuous tests) | Clean; no exit-0-only assertions ✓ | PASS |

---

### Recommendation

**APPROVED** with four concrete improvement suggestions (all non-blocking):

1. Add real-tool E2E for Fix B (`diff` exit 1 Full tier)
2. Add production-path E2E for Fix D (`git show HEAD:<file>.rs`)
3. Replace hardcoded line citations with function names (rules.rs + rustdoc)
4. Tighten Fix-A e2e predicates to assert full preserved-flag form

The PR is regression-safe: all six fixes are tightly scoped, two-entry-point invariant is respected, and load-bearing guards (structural flag-drop test, pure `should_emit_compressed_hint`, `command_needs_passthrough`) are backed by discriminating tests.

**Per ADR-005, merge is the user's decision.**

---

*Reviewed by: security (9/10), architecture (8/10), performance (9/10), complexity (9/10), consistency (9/10), regression (9/10), testing (8/10), reliability (9/10), Rust (9/10), documentation (9/10)*

PR #350, opened 2026-06-22
fix(#317): compression issue batch — net-savings guard, test-exit, diff-dedup, session-id

dean0x:
## Summary

Triages and fixes a batch of rewrite-hook / output-compression issue reports from coding agents, holding skim's vision — **compress, and when we can't beat raw, pass through (never expand, never lose meaning)** — and keeping **both interception surfaces** (the rewrite hook *and* the PATH wrappers) correct.

A **Phase 0 diagnostic pass** (built HEAD, reproduced every report) split the batch into *confirmed-live* vs *version-skew / not-skim*, so we only wrote code for real defects. Triage note: `.devflow/docs/phase0-triage-rewrite-compression-batch.md`.

| Report | Verdict | Action |
|---|---|---|
| 7.1 `ls -la` net-expands · 7.2 `wc -c` net-expands | **LIVE (C)** | net-savings guard |
| 3.1 `vitest run` passthrough forces exit 1 | **LIVE (D)** | real exit code |
| 6.1 `git diff` duplicates changed lines | **LIVE (F)** | per-file dedup cursor |
| 1.1 `--session-id` rejected on skew | skew + hardened (B) | drop flag injection |
| 1.2, 2.1, 5.1, 5.3 | already fixed on HEAD | confirmed, no code |
| 4.1 "denied" | **not a skim bug** (agent-harness perms / skew) | none |
| 5.2 `export …` `(eval):` error | **not skim** (agent shell) | none |

## What changed

**Cluster C — universal net-savings guard** (`cmd/execution.rs::savings_decision`, wired into the shared sinks: execution / git / build / test / log). Conservative rule (an explicit user decision): keep compressed **only if strictly smaller** than raw — measured in tokens below the size cap, byte-length above it; tie / empty / larger → passthrough (emit raw verbatim, no fabricated `tool N` / `LS:` / `GREP:` header). Exit code + stderr preserved on passthrough. Lives below the dispatch front-end, so it covers **both** surfaces. `--json`, `gh` streaming, and `heatmap` are exempt. Strengthens #317 (output only ever moves toward *more-complete* raw).

**Cluster C-7 — flag-substituting handlers.** `git status --short`/`-s`/`--porcelain` is judged against the **user's literal-flag** output (not skim's internal `--porcelain=v2`), so on a small/clean repo it passes through to raw and never expands relative to what the user asked for. Porcelain stays machine-faithful (empty = clean).

**Cluster D — test-runner exit fidelity** (`cmd/test/shared.rs`). The passthrough arm no longer hardcodes `ExitCode::FAILURE`; it returns `resolve_exit_code(0, exit_source)` — a passing suite skim merely couldn't parse now exits 0; a genuinely failing suite preserves its non-zero code; the PASS/FAIL summary leads the output.

**Cluster F — git diff de-duplication** (`cmd/git/diff/render.rs`). A per-`FileDiff` monotonic `EmittedCursor { last_new, last_old }` skips already-emitted patch lines (`+`/context on the new-line axis, `-` on the old-line axis), so changed lines on a hunk shared by two adjacent ranges are emitted **exactly once**. Context intact; non-overlapping diffs unchanged; `render_raw_hunks` untouched.

**Cluster B — session-id hardening** (`cmd/rewrite/hook.rs`, `main.rs`, `cmd/dispatch.rs`, `cmd/session_sidecar.rs`). Stops injecting `--session-id` into rewritten commands entirely — the only thing an older binary could hard-reject (the 1.1 failure class) — making the failure impossible regardless of version skew. Attribution flows out-of-band: sidecar (hook surface, ancestry walk) → `SKIM_SESSION_ID` env (wrapper surface) → the hidden `_session_id` clap field (forward-compat fallback). `strip_session_id_flag()` at the dispatch entry tolerates a stray flag on every subcommand (defense-in-depth).

## Key decisions

- **Net-savings rule: strictly-smaller-to-keep, ties pass through** (per the explicit decision; an early implementation had this inverted — corrected in `e94c989`). 0 tokens saved → passthrough.
- **Tokenization size cap lowered 64 MiB → 64 KiB** (`90fca15`, chosen with the maintainer). The guard tokenizes raw+compressed on the main thread; measured cost is ~0.3 s/MB in release (~3 s/MB in debug), and pathological on degenerate single-token input. 64 KiB bounds the guard's added latency to ~tens of ms at the cap in release; larger outputs use byte-length comparison (still never-expand-safe — token accuracy matters most for *small* outputs, which stay below the cap).
- **Empty/tiny raw → faithful passthrough** (no fabricated summary). Consistent with the strict rule and git's porcelain contract; the build-success "OK warnings:" summary still appears whenever the build produces real output (verified).

## Findings to flag (out of scope here)

1. **Analytics tokenization is the dominant large-output cost**, and it is *pre-existing* (present on `main`): recording token-savings tokenizes the output on the main thread (~0.3 s/MB release). The new guard cap does not address this path. Worth a follow-up (move it to the existing fire-and-forget background thread, or cap it too).
2. **Rewrite engine drops `run` from `cargo nextest run`** → `skim cargo nextest …` (observed via `skim rewrite`). Pre-existing; not in this batch's scope, but it silently breaks `cargo nextest run` under the hook (use `SKIM_PASSTHROUGH=1`).
3. **`skim <markdown>` emits headings out of source order** (title last) — observed on the triage note.

## Testing

- Full suite green: **3963 passed, 0 failed** (`cargo nextest run -p rskim -j 4`); `clippy -D warnings` + `cargo fmt --check` clean.
- New coverage: `cli_no_expansion_317.rs` (7.1/7.2 invariant on real `ls -la`/`wc -c`), `cli_session_id_skew.rs` (stray `--session-id` stripped for git/grep/ls), `cli_wrapper_argv0.rs` (**PATH-wrapper / argv0 surface** — the previously-untested second surface), plus a deterministic CAP-ENGAGED perf test and unit/integration updates.
- 30 scenario-based acceptance checks pass (functionality, exit codes, `--json` validity, stdout/stderr split, #317 never-expand, flag preservation).
- Both-surface coverage: rewrite/hook path and the wrapper (argv0) path both exercised.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,8 +42,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `gh api` and `gh run watch` are exempt from the `--json` steering check at the
   handler gate (matching the rewrite skip-list); only `--jq`/`-q`/`--template`/`-t`
   trigger passthrough for those subcommands.
+- **Test-runner exit-code fidelity** (#350) — A passing suite whose output skim cannot
+  parse now exits 0 instead of 1. Previously `resolve_exit_code` treated an unparseable
+  exit-0 result as a failure (exit 1); it now propagates the child's zero exit code
+  verbatim. Genuine non-zero exits from failing suites are preserved unchanged on all paths.
+- **Git diff changed-line de-duplication** (#350) — Lines appearing in a hunk covered by
+  adjacent diff ranges are now emitted exactly once. A per-`FileDiff` `EmittedCursor` tracks
+  the last-written position; overlapping ranges advance the cursor rather than re-emitting
+  the shared lines. No change to diff output for non-overlapping hunks.
 
 ### Changed
+- **Session-id attribution priority inverted: sidecar > env > flag** (#350) — The hook no longer
+  injects `--session-id` into rewritten commands; flag injection caused hard failures
+  (`"unexpected argument --session-id"`) on older binaries. Attribution now resolves in order:
+  sidecar (written out-of-band by the hook; found via ancestry walk) → `SKIM_SESSION_ID` env var
+  (wrapper-surface attribution; export alongside `PATH`) → `--session-id=VALUE` flag
+  (forward-compat fallback; honoured so old hooks that still inject the flag are not lost).
 - **Net-savings guard token-decision cap raised 64 KiB → 256 KiB; new 4 KiB longest-run guard
   for degenerate inputs** (#317 / #350) — The cap controlling when `savings_decision` falls
   back from exact token counts to fast byte comparison is raised from 64 KiB to 256 KiB,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,27 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   trigger passthrough for those subcommands.
 
 ### Changed
+- **Net-savings guard token-decision cap raised 64 KiB → 256 KiB; new 4 KiB longest-run guard
+  for degenerate inputs** (#317 / #350) — The cap controlling when `savings_decision` falls
+  back from exact token counts to fast byte comparison is raised from 64 KiB to 256 KiB,
+  improving token-accurate decisions for moderately large outputs.  A complementary
+  longest-run guard is added: when either string contains a non-whitespace run exceeding
+  4 KiB (and both strings are below the size cap), the function falls back to byte comparison
+  to avoid O(n²) BPE merge cost on minified JS / base64 / binary-as-text single-line inputs.
+  Real line-oriented shell output never triggers the run guard; the "never expand" safety
+  invariant is unchanged on all paths.
+- **Analytics stores true (gross) compressed-token counts on expansion rows** (#317 / #350) —
+  Previously `compressed_tokens` was clamped to `raw_tokens` when the output expanded.  It is
+  now stored as the true value.  The `tokens_saved` aggregate (in `query_summary`, `query_daily`,
+  and all other aggregate queries) is floored per-row to 0 via `CASE WHEN`, consistent with
+  existing `query_by_command` / `query_by_language` / `query_by_mode` / `query_by_session`
+  behavior.  Row-level `raw_tokens` / `compressed_tokens` now carry true gross counts, allowing
+  accurate expansion-rate analysis.  **Note:** rows written before this change remain clamped
+  (mixed historical data); this is acceptable for cumulative analytics — no migration is needed.
+- **`--show-stats` token counts reused for analytics recording** (#317 / #350) — When
+  `--show-stats` is active, the token counts already computed for the stats display are reused
+  to record the analytics row via `try_record_command_with_counts`, avoiding redundant
+  background re-tokenization.  The common path (no `--show-stats`) is unchanged.
 - **`serde_json` `preserve_order` feature enabled workspace-wide — key ordering changes** — (#302)
   Enabling `preserve_order` switches `serde_json::Map` from `BTreeMap` (alphabetical) to
   `IndexMap` (declaration/insertion order) for every crate in the workspace. Visible effects:
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -96,7 +96,7 @@ skim intercepts a sub-agent's shell command through **two independent mechanisms
 
 - `SKIM_PASSTHROUGH=1` — bypass all compression (use when compressed output hides an error). Indefinite commands (`vite dev`, `jest --watch`, bare `skim vitest`) auto-pass-through live; use `skim vitest run` for a compressed one-shot.
 - `SKIM_DEBUG=1` (or `--debug`) — warnings/notices on stderr.
-- `SKIM_SESSION_ID` — analytics session attribution; priority `--session-id` > sidecar > env > none. Set it alongside the PATH export so sub-agents inherit it.
+- `SKIM_SESSION_ID` — analytics session attribution; priority sidecar > env > `--session-id` flag (flag is a forward-compat fallback only — the hook no longer injects it). Set it alongside the PATH export so sub-agents inherit it.
 - `SKIM_CACHE_DIR` / `SKIM_ANALYTICS_DB` — override the cache dir / analytics DB path.
 - `SKIM_DISABLE_ANALYTICS=1` — disable recording. `SKIM_INPUT_COST_PER_MTOK` — $/MTok for cost estimates (default 3.0).
 - Session-provider overrides for `discover`/`learn`/`agents`: `SKIM_PROJECTS_DIR`, `SKIM_CODEX_SESSIONS_DIR`, `SKIM_COPILOT_DIR`, `SKIM_CURSOR_DB_PATH`, `SKIM_GEMINI_DIR`, `SKIM_CRUSH_DIR`.
--- a/crates/rskim/Cargo.toml
+++ b/crates/rskim/Cargo.toml
@@ -56,4 +56,3 @@ serde_json = { workspace = true }
 serial_test = { workspace = true }
 filetime = { workspace = true }
 rskim-contract = { path = "../rskim-contract" }
-rusqlite = { workspace = true }
--- a/crates/rskim/Cargo.toml
+++ b/crates/rskim/Cargo.toml
@@ -56,3 +56,4 @@ serde_json = { workspace = true }
 serial_test = { workspace = true }
 filetime = { workspace = true }
 rskim-contract = { path = "../rskim-contract" }
+rusqlite = { workspace = true }
--- a/crates/rskim/src/analytics/mod.rs
+++ b/crates/rskim/src/analytics/mod.rs
@@ -2262,14 +2262,9 @@ mod tests {
         // Only the two compressing rows contribute: 2 × (1000 - 100) = 1800.
         assert_eq!(summary.tokens_saved, 1800);
 
-        let daily = db.query_daily(None).unwrap();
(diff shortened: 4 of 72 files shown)


dean0x:
## Follow-up: tokenization-accuracy pass (commits `5bf8339`, `7f29178`)

Builds on this PR's net-savings guard with three changes:

**1. Guard token-decision cap 64 KiB → 256 KiB + degenerate-input run guard.** `savings_decision` now tokenizes (the exact token decision) up to 256 KiB instead of 64 KiB; above that it still falls back to byte comparison. A new 4 KiB longest-non-whitespace-run guard routes BPE-pathological inputs (long no-split runs — minified JS, base64) to the byte path, bounding worst-case latency (~100–150 ms at the cap in release; a degenerate 256 KiB single run is decided in <50 ms via the byte path). Never-expand contract unchanged.

**2. `--show-stats` reuses its token counts for analytics** instead of re-tokenizing on the background thread. The common (no-`--show-stats`) hot path is unchanged by design — still zero main-thread tokenization.

**3. Faithful (gross) expansion analytics.** Removed the `compressed_tokens.min(raw_tokens)` clamp so true counts are stored; the summary/daily `tokens_saved` aggregates now floor per-row to 0 via the same `CASE WHEN` the by-command/lang/mode/session aggregates already use. ⚠️ Rows written **before** this change remain clamped (mixed historical data — acceptable for cumulative analytics; no schema migration).

### Verification
- 2940/2940 `-p rskim` unit/integration tests; new expansion + `--show-stats` integration tests (`tests/cli_analytics_expansion.rs`); clippy + fmt clean; remote **Test Suite (full regression) green**.
- `--show-stats` rows confirmed **field-identical** to the background path, e.g. `skim git log` records `git | raw=264 | compressed=277 | passthrough | 0.0%` both with and without `--show-stats` — and note that row is a real expansion (277 > 264) now stored **gross/true** with savings floored to 0.
- The `security/snyk` check failure is the known **account-quota limit** ("used your limit of private tests"), not a code finding (Rust is not Snyk-Code-supported).

dean0x:
The "emit raw verbatim + add trailing newline + flip tier to passthrough" logic is duplicated across ~6 sinks (execution.rs:703-715, build/mod.rs:211-218, git/mod.rs:451-457, git/status.rs:169-178, test/shared.rs:173-181, log.rs:176-183). This should be consolidated into a single helper `emit_raw_passthrough()` to ensure the #317 never-expand/always-newline invariant cannot silently diverge across sinks on future edits.

**Suggested fix:** Extract `pub(crate) fn emit_raw_passthrough(out: &mut impl Write, raw: &str) -> io::Result<&'static str>` in cmd/execution.rs, implementing the canonical "emit raw + trailing newline + return \"passthrough\"" behavior. (HIGH: architecture 90%, complexity 88%, consistency 96%)

dean0x:
The entire `run_parsed_command_with_raw_override` function (~90 lines) duplicates `git/mod.rs::run_parsed_command` just to add an optional `raw_override` baseline. This double-maintains the credential-scrub path (PF-024) and other shared logic.

**Suggested fix:** Add `raw_override: Option<String>` parameter to the existing shared `git::run_parsed_command` (defaulting `None` at other call sites). Delete `run_parsed_command_with_raw_override`. This matches the module's own `combine_stderr` pattern and keeps credential-scrub single-sourced. (HIGH: architecture 88%, consistency 92%)

dean0x:
The `savings_decision_degenerate_perf_under_50ms` test asserts `< 50 ms` for a 256 KiB tokenization boundary. This is flaky under pinned 4-thread CI with sccache — a 50 ms wall-clock budget has no margin for scheduler jitter. Raise to 500 ms (still proves tokenization ~3s was skipped) or assert behaviorally.

**Suggested fix:** Change `assert!(elapsed < Duration::from_millis(50))` to `assert!(elapsed < Duration::from_millis(500))` or move timing to a separate `#[ignore]`d benchmark and assert via a behavioral proxy. (HIGH: testing 88%)

dean0x:
The `argv0_ls_wrapper_dispatches_and_does_not_expand` test asserts `skim_len <= raw_len + raw_len` (2× slack), not the strict `<=` enforced elsewhere. The #317 never-expand contract is `skim_len <= raw_len` — a 2× allowance would pass a near-doubling regression.

**Suggested fix:** Trim both sides and assert strict `skim_stdout.trim_end().len() <= raw_output.trim_end().len()`, matching `cli_no_expansion_317.rs`. (HIGH: testing 85%)

dean0x:
Three user-visible behavior changes are missing from Unreleased CHANGELOG:
1. **Session-ID resolution priority inverted** — attribution now resolves `sidecar > env > flag` (inverse of previous flag-first order).
2. **Test passthrough exit-code fidelity fixed** — a PASSING suite whose output skim couldn't parse previously reported exit 1; now forwards the real process exit.
3. **git diff de-duplication** — adjacent nodes sharing a hunk no longer emit changed lines twice.

Add entries under `### Changed` and `### Fixed`. See documentation review for exact wording. (HIGH: documentation 84-88%)

dean0x:
`longest_nonwhitespace_run()` is called at line 159-161 (two full O(n) scans, up to 256 KiB each) **before** the cheap `comp_t.len() >= raw_t.len()` byte early-exit at line 176. When compression doesn't win on bytes (common for small/incompressible output), the function wastes 512 KiB of scanning before deciding to Passthrough anyway.

**Suggested fix:** Reorder so cheap byte checks gate the run scan. Do size/byte checks FIRST; only compute `longest_nonwhitespace_run` when bytes indicate compression wins AND below size cap (immediately before `count_token_pair` at line 186). (MEDIUM: performance 80%)

dean0x:
When `--show-stats` is active, the same `(raw, compressed)` pair is tokenized **twice on the main thread**: once in the guard (via `savings_decision`), then again in the stats display (via `record_and_report`). At ~0.3 s/MB release cost, a 200 KiB output pays ~240 ms total main-thread token cost.

**Suggested fix:** Have `savings_decision` optionally return the token counts it computed, and feed those into stats when acceptable. Or tokenize once on untrimmed strings and derive trimmed counts from length adjustment. (MEDIUM: performance 85%)

dean0x:
## Summary: PR #350 Review Complete

**Overall Posture:** 0 blocking issues across 11 reviewers; mergeable with should-fix follow-ups.

All behavior is sound and well-tested. The net-savings guard is correctly bounded, exit-code fidelity is fixed, git-diff dedup is accurate, and the session-id migration is defended on both interception surfaces. No regressions or comprehension barriers.

### Should-Fix Items (8 total)

**High-priority (blocks polish):**
1. **Duplication consolidation** — 6 sinks re-implement emit-raw-passthrough (HIGH ≥88%) → extract helper
2. **git/status.rs copy** — 90-line duplicate of git/mod.rs (HIGH ≥88%) → add `raw_override` param to shared function
3. **Flaky test** — 50 ms wall-clock assertion (HIGH 88%) → raise to 500 ms or use behavioral proxy
4. **Wrapper-surface guard** — 2× slack never-expand assertion (HIGH 85%) → tighten to strict `<=`

**Medium-priority (perf/optimization):**
5. **Whitespace-run reorder** — scan before byte gate (MEDIUM 80%) → compute only when compression wins
6. **Double tokenization** — --show-stats re-tokenizes (MEDIUM 85%) → reuse counts from guard

**Documentation-only (critical accuracy):**
7. **CLAUDE.md stale** — session-id priority inverted but doc shows old order (CRITICAL 96%) → update to `sidecar > env > flag`
8. **CHANGELOG omissions** — 3 user-visible changes not documented (HIGH 84-88%) → add session-id priority, exit-code fix, diff-dedup entries

### Lower-Confidence Items (7 total)
- Session-id priority phrasing reconciliation (70%)
- `EmittedCursor` naming consistency (`cursor` vs `emitted_cur` in render.rs) (90% — minor style)
- Redundant rusqlite dev-dep (72%)
- `const { assert!(TOKEN_SIZE_CAP > TOKEN_RUN_CAP) }` suggestion (70%)
- Git status double-spawn justification / single-invocation path (75%)
- Heatmap docstring exemption wording (78%)
- Analytics e2e test 300 ms sleep robustness (78% — inherent to fire-and-forget)

### Review Scores by Discipline
| Discipline | Score | Notes |
|-----------|-------|-------|
| Security | 9/10 | Clean; no injection surfaces; credential-redaction unchanged |
| Architecture | 8/10 | Policy (savings_decision) centralized; mechanism (emit) duplicated 6× → needs consolidation |
| Performance | 8/10 | Guard bounds sound; two MEDIUM efficiency wins available |
| Complexity | 8/10 | Functions under limits; duplication is maintainability, not comprehension |
| Consistency | 7/10 | Wrapper rollout uniform; guard invocation asymmetric across sinks |
| Regression | 9/10 | Every behavior change locked by targeted test; #317 invariant upheld |
| Testing | 8/10 | Strong coverage; fix timing assertion flakiness and wrapper-surface assertion tightness |
| Reliability | 10/10 | Every loop bounded; assertions present; errors degrade gracefully |
| Rust | 9/10 | Idiomatic; no unwrap on production paths; borrowing optimized |
| Dependencies | 10/10 | Clean; workspace-pinned; no transitive bloat |
| Documentation | 7/10 | Code comments 9+/10; project docs need 1 CRITICAL + 3 HIGH fixes |

**Recommendation:** APPROVED WITH CONDITIONS — Fix the 4 high-priority should-fix items (duplication, flaky test, guard tightness) and the 2 critical documentation updates (CLAUDE.md, CHANGELOG) before merge. The 2 MEDIUM perf items are recommended but not blocking. 

Generated by automated review across 11 specialized agents (security, architecture, performance, complexity, consistency, regression, testing, reliability, rust, dependencies, documentation). See inline comments above for file:line details.

PR #22, opened 2026-03-11
feat: add --mode=minimal transformation mode

dean0x:
## Summary

Adds `--mode=minimal` — a new light-touch transformation mode that fills the gap between `full` (no transformation) and `structure` (bodies stripped). Closes the first deliverable of #18.

- **Strips** non-doc comments (`//`, `#`, `/* */`) at module/class level
- **Preserves** doc comments (`/** */`, `///`, `//!`, Python docstrings, Go adjacency docs, Javadoc)
- **Preserves** all function bodies and comments inside them
- **Preserves** shebangs (`#!/usr/bin/env python3`)
- **Normalizes** excessive blank lines (3+ consecutive → 2)
- **Passthrough** for JSON, YAML, and Markdown (no comments to strip)
- **Expected savings**: 15-30% token reduction on typical source files

### Implementation

Three-pass algorithm in `crates/rskim-core/src/transform/minimal.rs`:
1. Walk tree-sitter AST collecting byte ranges of non-doc comment nodes to remove
2. Remove ranges from source, trim trailing whitespace on affected lines
3. Normalize blank lines (3+ consecutive → 2)

New shared utilities in `crates/rskim-core/src/transform/utils.rs`:
- `get_comment_prefix()` / `get_comment_suffix()` — language-aware comment markers
- `is_inside_function_body()` — AST ancestor walk for body detection

### Files changed (16 files, +1531/-4)

**New modules:** `minimal.rs` (414 lines), `utils.rs` (99 lines)
**New fixtures:** 6 comment-rich test fixtures (TS, JS, Python, Rust, Go, Java)
**Modified:** types.rs, mod.rs, lib.rs, main.rs, integration.rs, cli.rs, benchmarks

### Test results

- 138/138 core tests pass (41 unit + 87 integration + 10 doc)
- 23/23 CLI tests pass
- 39 new tests added (30 integration + 4 CLI + 5 unit)
- Clippy clean, rustfmt clean, Snyk scan clean

## Test plan

- [x] All 151+ existing tests still pass (no regressions)
- [x] TypeScript: regular comments stripped, JSDoc preserved, body comments preserved
- [x] Python: `#` comments stripped, docstrings preserved, shebang preserved
- [x] Rust: `//` stripped, `///`/`//!` preserved, body comments preserved
- [x] Go: standalone comments stripped, adjacency doc comments preserved
- [x] Java: regular stripped, Javadoc preserved, body comments preserved
- [x] JSON/YAML/Markdown passthrough (output == input)
[...]

Issue #21, opened 2026-03-08
Phase D: Platform & Ecosystem — Plugins, Context Budgeting & Import Graph Analysis


dean0x:
# Phase D: Platform & Ecosystem

**Parent:** #17 (Project Horizon)
**Priority:** Future — blue ocean opportunities
**Depends on:** Phases A-C establishing the foundation

## Context

Once skim owns code reading (Phase A), command output (Phase B), and analytics (Phase C), the next frontier is **intelligent context optimization** — features that no tool in the ecosystem currently offers.

Note: Token budget mode (`--tokens N`) for single files was pulled forward into Phase A (#18) due to its high impact. Phase D covers the **multi-file and cross-file** extensions of that concept.

### 1. Cross-File Context Budgeting
- [ ] `skim budget 8000 src/**/*.ts` — fit maximum information from multiple files into N tokens
- [ ] Intelligent file prioritization: which files are most relevant?
- [ ] Cross-file deduplication: don't repeat shared patterns
- [ ] Output: concatenated view with file headers, optimized for the budget
- [ ] Builds on single-file `--tokens N` from Phase A

### 2. Import Graph Analysis
- [ ] `skim deps file.rs` — show what this file imports and what imports it
- [ ] `skim context file.rs` — serve the file + all direct dependencies' signatures
- [ ] tree-sitter powered: parse import/use/require statements across all supported languages
- [ ] Use case: "I'm working on file.rs — give me the context I need, nothing more"

### 3. Diff-Aware Context (supersedes #9)
- [ ] `skim diff` — show only changed regions + surrounding structural context
- [ ] Parse git diff, locate changed lines, find containing function/class via tree-sitter
- [ ] Output: changed functions in full, unchanged functions as signatures-only
- [ ] Use case: "Show me what changed, with enough context to understand it"

### 4. Plugin System
- [ ] Define a `Filter` trait for community-contributed command output filters
- [ ] Plugin discovery: `~/.config/skim/plugins/` or cargo feature flags
- [ ] Standard interface: `fn filter(raw_output: &str, command: &str) -> FilterResult`
- [ ] Built-in filters for Phase B commands; community can add more

### 5. Semantic Deduplication
- [ ] Detect structurally similar code blocks across files
- [ ] Collapse duplicates: show first occurrence in full, subsequent as "similar to file.rs:42"
- [ ] tree-sitter powered: compare AST node structures, not text

## Acceptance Criteria

[...]

Issue #13, opened 2026-02-01
Tech Debt Backlog


dean0x:
# Tech Debt Backlog

Tracking pre-existing improvements and test coverage gaps deferred from code reviews.

### 1. Refactor getBinaryPath() into smaller functions
- **Location**: npm/rskim/bin/skim:114-209
- **Category**: Code quality
- **Priority**: Medium
- **Description**: Current function is too long and handles multiple concerns. Should be split into focused functions for clarity and maintainability.
- **Status**: Cosmetic improvement, working code
- **Added**: 2026-02-01 from feat-improved-linux-distribution review

### 2. Add unit tests for isMusl() function
- **Location**: npm/rskim/bin/skim
- **Category**: Test coverage
- **Priority**: Medium
- **Description**: Add comprehensive unit tests for musl detection strategies to ensure cross-platform binary selection works correctly.
- **Status**: Test coverage gap
- **Added**: 2026-02-01 from feat-improved-linux-distribution review

### 3. Add tests for findPlatformBinary() search paths
- **Location**: npm/rskim/bin/skim
- **Category**: Test coverage
- **Priority**: Medium
- **Description**: Add tests for binary lookup paths to verify correct fallback behavior across platforms.
- **Status**: Test coverage gap
- **Added**: 2026-02-01 from feat-improved-linux-distribution review

### 4. Add tests for postinstall download/extract logic
- **Location**: npm/rskim/scripts/postinstall.js
- **Category**: Test coverage
- **Priority**: Medium
- **Description**: Add test coverage for network operations, download verification, and extraction logic to catch edge cases.
- **Status**: Test coverage gap
- **Added**: 2026-02-01 from feat-improved-linux-distribution review

---

## Guidelines

- Items are ordered by review date
- Pull requests should reference this issue when addressing items
- Close item sections when work is complete with link to PR
- Archive older items (>6 months) if not addressed

---

_Tech debt backlog created by Claude Code on 2026-02-01_

PR #360, opened 2026-06-24
fix(#359): record file-op analytics independent of cache state; unify SKIM_CACHE_DIR

### LINKED ISSUE #359: analytics & cache-dir: plain `skim <file>` drops token-savings on cold cache; `SKIM_CACHE_DIR` ignored by parser cache + analytics default (fix together)
**Two related, pre-existing bugs in skim's cache/analytics layer, to be fixed together.**
They're bundled because Problem B's fix (`SKIM_CACHE_DIR` honored everywhere) is what lets
Problem A's tests isolate the parser cache cleanly. Neither blocks PR #350.

---

# Problem A — plain `skim <file>` silently drops token-savings analytics

A plain `skim <file>` invocation (no `--show-stats`) records a token-savings row **only if the
parser cache already holds token counts** for that file+mode — which only happens after a prior
`--show-stats` run on the same file. On a cold cache, a plain-warmed entry, or with `--no-cache`,
the file-operation path records **nothing**. Token savings for the most common invocation —
agents calling `skim <file>` via the hook *without* `--show-stats` — are silently dropped.

## Root cause (A)

- `process.rs` computes token counts on the main thread **only** when `--show-stats` is set
  (`(None, None)` otherwise) — intentional perf optimization (don't pay the tiktoken BPE cost on
  every invocation).
- `record_file_analytics` (main.rs) records **only when** `result.original_tokens` /
  `transformed_tokens` are `Some`. Without `--show-stats` they are `None` → nothing recorded.
- The only way a plain run gets `Some` counts is a cache **hit** on an entry written by a prior
  `--show-stats` run (`try_cached_result`; `needs_recount` recomputes only under `--show-stats`).
- Unlike subcommands (git/build/grep/…), the file-op path has **no background-tokenization
  fallback** (`try_record_command` / `record_fire_and_forget` are `cmd/`-only). So file-op
  analytics is accidentally gated on cache-carried counts. This contradicts the code's own
  comments (process.rs:444-445, 173-176) which say analytics computes counts on a background
  thread independent of `--show-stats`.

## Reproduction (deterministic; cold cache via HOME isolation)

| Scenario | Rows recorded |
|---|---|
| `skim file.ts` — cold cache (miss) | **0** |
| `skim file.ts` ×2 — miss then hit on plain-written entry | 0, 0 |
| `skim file.ts --show-stats` then `skim file.ts` (hit w/ counts) | 1, **1** |
| `skim file.ts --no-cache` | **0** |
| `skim file.ts --show-stats` | 1 |

## Confirmed scope (A)

- **Single file `skim <file>`** — affected.
- **stdin `skim -`** — affected (`record_file_analytics` at main.rs:856).
- **multi-file / glob / directory** — **CONFIRMED affected**: `multi.rs` sums each file's
  `process_result.original_tokens` (multi.rs:237-243), `None` without `--show-stats`, then gates
  recording on `total_original_tokens > 0` (multi.rs:283) → cold-cache multi runs record nothing.

## Why the removed test flaked (corrected attribution)

`test_show_stats_off_records_exactly_one_row` asserted 1 row for a plain run. Tests share the
default `~/.cache/skim`, so on macOS a sibling `--show-stats` run on the same `simple.ts` warmed
the cache with counts → plain recorded 1 → pass; cold Linux CI → 0 → fail (`left 0, right 1`).
This was **not** a writer race: `record_with_counts` registers its thread in `PENDING_THREADS`
and `flush_pending()` (main.rs:760, after the pipeline at main.rs:745) joins it before exit, so
post-exit reads are deterministic *given* cache state. The `8325d45` commit message and PF-001
attributing the flake to fire-and-forget timing are incorrect.

## Where the raw source text is available (the design crux for A)

Recording needs the raw token count → needs the raw source text. Availability differs by path:

| Path | Raw text available at record time? |
|---|---|
| file, cache miss | yes — `contents` read in `process_file` |
| file, cache **hit** | **no** — the cache deliberately skips the file read |
| stdin | in `process_stdin`'s local `buffer`, not currently returned |
| multi-file | per-file `contents` inside the rayon map, not retained |

The cache-hit case is the crux: re-read the file on the background thread, or persist counts
into the cache.

## Proposed fix (A)

Record file-op analytics independent of cache state via **background tokenization** (matching the
subcommand `record_fire_and_forget` path), off the main thread to preserve the "no BPE cost
without `--show-stats`" optimization:
- `record_file_analytics`: when counts are `None` and analytics enabled, spawn a **registered**
  background thread that tokenizes raw + compressed and records (`command_type=File`, mode, language).
- Supply raw: miss → `contents` in hand; **hit → re-read the file** (recommended; simplest), or
  backfill counts into the cache; stdin → retain the buffer; multi → retain per-file raw (mind
  memory on big globs) or record per file.

---

# Problem B — `SKIM_CACHE_DIR` is honored inconsistently (ignored by the parser cache + default analytics DB)

`SKIM_CACHE_DIR` is documented (CLAUDE.md) as overriding the cache dir, and it **does** work for
some subsystems — but it is **silently ignored by the parser cache, tee output, and the default
`analytics.db` path**. Two cache-dir resolvers drifted apart:

| Resolver | Reads `SKIM_CACHE_DIR`? | Used by |
|---|---|---|
| `cmd::resolve_cache_dir()` (`hook_log::CacheEnv::from_process`, hook_log.rs:34) | ✅ yes | search index (index.rs:763), hook log (hook_log.rs:61), rewrite stamps (rewrite/hook.rs:141/293/334/365), startup (main.rs:721) |
| `cache::get_cache_dir()` (cache.rs:86, `dirs::cache_dir()` only) | ❌ **no** | **parser cache** (cache.rs:156/191/224), tee output (output/tee.rs:39), **default `analytics.db`** (analytics/mod.rs:388) |

## Impact (B)

- **Consistency / least-surprise + doc mismatch**: the same env var relocates the search index and
  hook logs but not the parser cache or default analytics DB.
- **Testability**: you cannot isolate the parser cache via `SKIM_CACHE_DIR` (the Problem-A repro had
  to override `HOME` instead). Fixing B makes A's tests hermetic without `HOME` juggling.
- No data loss / no wrong output — the parser cache is regenerable. Low–moderate severity.

## Proposed fix (B)

Route both resolvers through a single source of truth so `SKIM_CACHE_DIR` is honored everywhere
(or, minimally, have `get_cache_dir()` check `SKIM_CACHE_DIR` before `dirs::cache_dir()`).
~5–10 LOC minimal; ~30–60 LOC for the unified-helper + test version.

**Decision needed:** `get_cache_dir()` also yields the *default* `analytics.db` path
(analytics/mod.rs:388, used when `SKIM_ANALYTICS_DB` is unset). Honoring `SKIM_CACHE_DIR` there
means a user who sets `SKIM_CACHE_DIR` (and not `SKIM_ANALYTICS_DB`) would see `analytics.db` move
to the new dir — old history stays on disk but skim reads the new location. Recommended: **yes**
(`SKIM_CACHE_DIR` = all skim cache state), documented as a behavior change.

---

# Combined fix plan

- **Branch off `main`** (keep PR #350 scoped). Suggested order:
  1. **Problem B** first (~30–60 LOC, low risk) — unify cache-dir resolution; gives A's tests
     clean `SKIM_CACHE_DIR`-based parser-cache isolation.
  2. **Problem A, phase 1**: single-file + stdin background tokenization; un-ignore the regression
     test (~90–140 LOC).
  3. **Problem A, phase 2**: multi-file / glob / dir (~40–70 LOC).
- **Decisions locked (recommended):** re-read file on cache hit (A); `SKIM_CACHE_DIR` also moves the
  default `analytics.db` (B).
- **Sizing (combined):** ~210–360 LOC incl. tests.
- **Risk:** main-thread latency unchanged (A's work is background, joined by `flush_pending`; <50ms
  bench unaffected). New per-invocation cost: thread spawn + tokenize (+ a small `fs::read` on cache
  hits) on every file op — the agent hook runs `skim` often; subcommands already pay this, but worth
  a bench sanity check. Behavior change: `analytics.db` grows much faster (every file op recorded) —
  intended (ADR-001); pruning (`maybe_prune`) already exists.

# Regression test status

`crates/rskim/tests/cli_analytics_expansion.rs::test_plain_file_op_should_record_analytics_no_cache`
already added as `#[ignore]`'d (asserts the desired 1 row; currently records 0) in commit
`a38676e`, along with the corrected NOTE. Remove `#[ignore]` when Problem A is fixed.

dean0x:
## Summary

Fixes two pre-existing bugs in skim's cache/analytics layer (#359), landed together because Problem B's fix is what lets Problem A's tests isolate the parser cache cleanly.

**Problem A — silent analytics loss (PF-001).** Plain `skim <file>` — the most common invocation (agents calling skim via the hook) — silently dropped token-savings analytics unless the parser cache already carried token counts (i.e. a prior `--show-stats` run). On a cold cache, a plain-warmed entry, or `--no-cache`, nothing was recorded, so most real savings data was lost. Root cause: token counts were computed on the main thread only under `--show-stats`, and the file-op path lacked the background-tokenization fallback every subcommand already had.

**Problem B — `SKIM_CACHE_DIR` drift (PF-002).** `SKIM_CACHE_DIR` was honored by the search index and hook log but silently ignored by the parser cache, tee output, and the default `analytics.db` path — two cache-dir resolvers had drifted apart.

## What changed

- **Unified background recorder** (`record_file_ops`, `analytics/mod.rs`): tokenizes off the main thread — registered via `register_thread` so `flush_pending()` joins it — AFTER stdout is flushed, preserving the streaming contract (analytics never delays or alters output). Emits per-file rows for single-file, stdin, and multi/glob/dir, each carrying the **detected** language, independent of cache state. Single-file re-reads the source through the 50 MB-guarded, TOCTOU-safe `read_source`; stdin retains its buffer; deleted/changed/non-UTF8 files are skipped best-effort (no panic).
- **Single cache-root resolver** (`cache::cache_root` / `cache_root_from`): `get_cache_dir` and `hook_log::CacheEnv` both delegate to it. `SKIM_CACHE_DIR` now relocates the parser cache, tee output, **and** the default `analytics.db`; `SKIM_ANALYTICS_DB` still wins when explicitly set; empty/whitespace `SKIM_CACHE_DIR` is treated as unset.
- **Shared, analytics-isolated test harness** (`tests/common/mod.rs`): analytics-OFF-by-default `skim()` plus opt-in `skim_with_analytics(db)`; ~49 integration test files migrated so the new file-op recording can never write to a developer's real `analytics.db`.

## Behavior changes (documented)

- The default `analytics.db` relocates with `SKIM_CACHE_DIR` (prior history at the old path is orphaned, not migrated).
- File-op analytics rows now carry the **detected** language (previously `--language`-only).
- Multi-file/glob/directory now emit **N per-file rows** (previously a single aggregate row).
- No `token_savings` schema change / no migration; the subcommand analytics path (`record_fire_and_forget`) is untouched; stdout is byte-identical for all file ops.

## Commits

`B` unify cache-dir resolution → `T` shared test harness + migrate integration tests → `A1` background recorder + single-file/stdin + `ProcessResult.language`/`stdin_raw` → `A2` multi/glob/dir per-file rows → `T-fix` repair a `mod common;` migration defect → `simplify` 1-line clarity → `test` B1 relocation test + CHANGELOG + comment cleanup.

## Verification

- **Suite green:** 2943 rskim integration tests + 470 rskim-core (4006 under `--all-targets`), 1 pre-existing unrelated hang filtered (`runner::tests::test_node_fallback_local_bin_found`); `cargo clippy -p rskim -p rskim-core --all-targets -- -D warnings` clean; `cargo fmt --check` clean.
- **Acceptance QA on the real binary (isolated DBs):** F1 `1|file|rust`; F3 plain vs `--show-stats` token-count parity; F7 mixed-language → 3 per-file rows; F8 directory → N rows; F11 empty file → `1|0|0.0` (no panic); F13 `SKIM_DISABLE_ANALYTICS=1` → 0 rows; B1 `SKIM_CACHE_DIR` alone → `analytics.db` + parser cache both land in the dir (1 row); B4 `SKIM_ANALYTICS_DB` precedence; B7 empty `SKIM_CACHE_DIR` → exit 0; C3 stdout byte-identical analytics on vs off.

## Out of scope

Unsupported-extension input currently exits `1` rather than the documented `3` — verified **pre-existing on `main`** and untouched by this branch (exit-code routing unchanged). Worth a separate issue.

Closes #359.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   default when explicitly set. Empty `SKIM_CACHE_DIR` is treated as unset (falls back to
   the platform default `~/.cache/skim`).
 
+- **Plain `skim <file>` now always records token-savings analytics (#359 Phase A)** —
+  Previously a plain `skim <file>` (and stdin, glob, and directory invocations) only
+  recorded analytics when token counts were already present in the parser cache from a
+  prior `--show-stats` run; cold-cache and plain-warmed-cache runs silently dropped all
+  data. The fix introduces a unified `record_file_ops` path that records token counts
+  independent of cache state, with the detected language attached to each row. Multi-file
+  invocations (glob, directory) now emit one analytics row per file instead of a single
+  aggregate.
+
 ### Added
 - **`rskim-tokens` crate (L3 Wave-1)** — Multi-provider token counting library (cl100k /
   o200k / Anthropic-offline / heuristic). Default build is HTTP-free; `net-anthropic` feature
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,25 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- **Unified `SKIM_CACHE_DIR` resolution — honored by all cache subsystems (#359 Phase B)** —
+  Previously `SKIM_CACHE_DIR` was silently ignored by the parser cache and the default
+  `analytics.db` path (`cache::get_cache_dir` read only `dirs::cache_dir()`) while the
+  search index and hook log respected it. This caused partial relocation: some skim state
+  moved, some stayed under `~/.cache/skim` (PF-002).
+
+  The fix introduces a single source of truth:
+  - `cache::cache_root_from(override_dir)` — pure resolver (no I/O); filters empty paths.
+  - `cache::cache_root()` — reads `SKIM_CACHE_DIR` and delegates to `cache_root_from`.
+  - `cache::get_cache_dir()` now calls `cache_root()` before the mkdir/chmod block.
+  - `cmd::hook_log::CacheEnv::resolve_cache_dir()` now delegates to `cache::cache_root_from`
+    instead of its own inline resolver.
+
+  **Behavior change:** `SKIM_CACHE_DIR` now also relocates the default `analytics.db`
+  (previously it did not). `SKIM_ANALYTICS_DB` still takes precedence over the relocated
+  default when explicitly set. Empty `SKIM_CACHE_DIR` is treated as unset (falls back to
+  the platform default `~/.cache/skim`).
+
 ### Added
 - **`rskim-tokens` crate (L3 Wave-1)** — Multi-provider token counting library (cl100k /
   o200k / Anthropic-offline / heuristic). Default build is HTTP-free; `net-anthropic` feature
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   **Behavior change:** `SKIM_CACHE_DIR` now also relocates the default `analytics.db`
   (previously it did not). `SKIM_ANALYTICS_DB` still takes precedence over the relocated
   default when explicitly set. Empty `SKIM_CACHE_DIR` is treated as unset (falls back to
-  the platform default `~/.cache/skim`).
+  the platform default `~/.cache/skim`). **Caveat:** pre-existing history at the old
+  `~/.cache/skim/analytics.db` is **not migrated** — setting `SKIM_CACHE_DIR` for the
+  first time causes `skim stats` to start from an empty DB at the new location; the old
+  file remains at `~/.cache/skim/analytics.db` and must be moved manually if you want to
+  preserve history.
 
 - **Plain `skim <file>` now always records token-savings analytics (#359 Phase A)** —
   Previously a plain `skim <file>` (and stdin, glob, and directory invocations) only
@@ -33,7 +37,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   data. The fix introduces a unified `record_file_ops` path that records token counts
   independent of cache state, with the detected language attached to each row. Multi-file
   invocations (glob, directory) now emit one analytics row per file instead of a single
-  aggregate.
+  aggregate. **Dashboard metric note:** `skim stats` computes `invocations` as a row count,
+  so a 3-file run now contributes +3 to the invocations counter instead of +1; historical
+  data recorded before this release used the old single-row-per-invocation convention, so
+  `invocations` comparisons across the upgrade point will reflect this change in counting
+  semantics.
 
 ### Added
 - **`rskim-tokens` crate (L3 Wave-1)** — Multi-provider token counting library (cl100k /
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -97,7 +97,15 @@ skim intercepts a sub-agent's shell command through **two independent mechanisms
 - `SKIM_PASSTHROUGH=1` — bypass all compression (use when compressed output hides an error). Indefinite commands (`vite dev`, `jest --watch`, bare `skim vitest`) auto-pass-through live; use `skim vitest run` for a compressed one-shot.
 - `SKIM_DEBUG=1` (or `--debug`) — warnings/notices on stderr.
 - `SKIM_SESSION_ID` — analytics session attribution; priority sidecar > env > `--session-id` flag (flag is a forward-compat fallback only — the hook no longer injects it). Set it alongside the PATH export so sub-agents inherit it.
-- `SKIM_CACHE_DIR` / `SKIM_ANALYTICS_DB` — override the cache dir / analytics DB path.
+- `SKIM_CACHE_DIR` — relocates **all** skim cache state: parser cache (`.json` files),
+  tee output (`tee/`), and the **default** `analytics.db` location. An empty value is
+  treated as unset (falls back to `~/.cache/skim`). The path is used as-is (no `skim`
+  suffix is appended by the resolver).
+- `SKIM_ANALYTICS_DB` — overrides the analytics DB path directly; **takes precedence over
+  `SKIM_CACHE_DIR`** for the DB location. When `SKIM_ANALYTICS_DB` is set, the DB is
+  opened at that exact path regardless of `SKIM_CACHE_DIR`. To isolate all skim state
+  in a sandbox it is sufficient to set `SKIM_CACHE_DIR` alone (the default analytics.db
+  moves with it).
 - `SKIM_DISABLE_ANALYTICS=1` — disable recording. `SKIM_INPUT_COST_PER_MTOK` — $/MTok for cost estimates (default 3.0).
 - Session-provider overrides for `discover`/`learn`/`agents`: `SKIM_PROJECTS_DIR`, `SKIM_CODEX_SESSIONS_DIR`, `SKIM_COPILOT_DIR`, `SKIM_CURSOR_DB_PATH`, `SKIM_GEMINI_DIR`, `SKIM_CRUSH_DIR`.
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -41,7 +41,7 @@ Streaming output (stdout, zero-copy via &str slices where possible)
 `transform_source()` routes each language to its parser via the Strategy Pattern, avoiding special-case conditionals — each language encapsulates its own strategy.
 
 **Non-obvious behavior (gotchas):**
-- **Analytics:** token savings persist to `~/.cache/skim/analytics.db` (SQLite/WAL), recorded fire-and-forget on background threads. `--clear-cache` clears only the parser cache, NOT `analytics.db` — use `skim stats --clear` for that. The `AnalyticsStore` trait + `MockStore` make the stats dashboard testable without a real DB.
+- **Analytics:** token savings persist to `~/.cache/skim/analytics.db` (SQLite/WAL; default location — relocates with `SKIM_CACHE_DIR`, see Environment Variables), recorded fire-and-forget on background threads. `--clear-cache` clears only the parser cache, NOT `analytics.db` — use `skim stats --clear` for that. The `AnalyticsStore` trait + `MockStore` make the stats dashboard testable without a real DB.
 - **Search DB:** `rskim-search` stores hotspot/risk/co-change data in `<root>/.skim/search.db`. Migrations are forward-only via `PRAGMA user_version`; a DB written by a newer version errors rather than corrupting data.
 - **AST index:** the n-gram index (`ast_index.skidx` / `.skpost`) is format v2 — v1 files are rejected with "please rebuild" (`skim search index --rebuild`). Synthetic n-gram markers (IDs ≥ 64900) resolve to `None` in `vocab_resolve()`, keeping them isolated from real vocabulary.
 
@@ -100,7 +100,10 @@ skim intercepts a sub-agent's shell command through **two independent mechanisms
 - `SKIM_CACHE_DIR` — relocates **all** skim cache state: parser cache (`.json` files),
   tee output (`tee/`), and the **default** `analytics.db` location. An empty value is
   treated as unset (falls back to `~/.cache/skim`). The path is used as-is (no `skim`
-  suffix is appended by the resolver).
+  suffix is appended by the resolver). **Caveat:** pre-existing analytics history at the
+  old `~/.cache/skim/analytics.db` is **not migrated** — setting this variable for the
+  first time causes `skim stats` to start from an empty DB at the new location; move
+  the old file manually if you want to preserve history.
 - `SKIM_ANALYTICS_DB` — overrides the analytics DB path directly; **takes precedence over
   `SKIM_CACHE_DIR`** for the DB location. When `SKIM_ANALYTICS_DB` is set, the DB is
   opened at that exact path regardless of `SKIM_CACHE_DIR`. To isolate all skim state
--- a/crates/rskim/src/analytics/mod.rs
+++ b/crates/rskim/src/analytics/mod.rs
@@ -23,6 +23,7 @@ use std::path::{Path, PathBuf};
 use std::sync::Mutex;
 use std::time::Duration;
 
+use rayon::prelude::*;
 use rusqlite::Connection;
 use serde::Serialize;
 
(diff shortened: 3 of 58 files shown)


dean0x:
## [Performance + Database Review] Per-row SQLite open and prune in persist loop

**File**: `crates/rskim/src/analytics/mod.rs`  
**Lines**: ~1127-1129 (persist loop in `record_file_ops`), ~975-980 (`persist_record`), ~636-657 (`maybe_prune`)  
**Confidence**: 88% (both reviewers flagged this consolidation)

### Issue
`record_file_ops` persists each record by calling `persist_record(&rec)` in a serial loop. Each call to `persist_record` invokes `AnalyticsDb::open_default()`, which performs a full connection setup per row:
- `Connection::open`
- `std::fs::metadata` + `set_permissions` syscall pair  
- `busy_timeout`
- `PRAGMA journal_mode=WAL`
- `schema::run_migrations`

Additionally, `db.maybe_prune()` runs a SELECT query against `analytics_meta` **once per row**.

For a `skim <directory>/` over N files, this results in N connection opens, N prune-checks, and N WAL handshakes in a single background thread—linearly scaling with directory size.

### Impact
- Not on the critical path (output is already flushed), so streaming targets are met.
- Inflates exit latency proportional to file count; on large multi-file runs the background join can become measurably slow.

### Suggested Fix
Open one connection per `record_file_ops` invocation and batch all inserts:
```rust
if let Ok(db) = AnalyticsDb::open_default() {
    for rec in &records {
        let _ = db.record(rec);
    }
    db.maybe_prune(); // Once per batch, not per row
}
```

This collapses N opens/migrations/prunes to 1 while preserving the single-writer guarantee (inserts remain serial within the one connection).

---
*[performance+database review]*

dean0x:
## [Complexity Review] Doubled enabled guard creates ambiguous ownership

**File**: `crates/rskim/src/analytics/mod.rs` (line ~1085-1088), `crates/rskim/src/multi.rs` (line ~291, 336), `crates/rskim/src/main.rs` (line ~949-950)  
**Confidence**: 88%

### Issue
`record_file_ops(enabled, rows, common)` re-checks `if !enabled || rows.is_empty()` internally, but **every** call site already guards before building `rows`:

1. **`multi.rs:291`**: Wraps the entire row-construction block in `if options.analytics_enabled { ... record_file_ops(true, rows, common) }`. This builds the full `rows` Vec (with per-file `path.clone()` moves) even though the callee will immediately re-test the literal `true`.

2. **`main.rs:949`**: Forwards the `enabled` param from `record_file_analytics` after already performing significant work to build the `FileCounts` variant.

This creates an ambiguous ownership contract: it's unclear whether `record_file_ops` is the single owner of the guard, or whether callers are responsible. A future reader may "optimize away" the redundant caller-side guard on the wrong side of the boundary.

### Impact
Low runtime cost (the work is small), but it is exactly the kind of "who is responsible" ambiguity that accumulates over time. The internal `!enabled` branch is partially dead from `multi.rs`'s perspective (always `true`).

### Suggested Fix
Pick one approach for consistency:

**Option A**: Make `record_file_ops` the single owner of the guard. Drop the `if options.analytics_enabled` wrapper in `multi.rs` and build `rows` unconditionally—the early `rows.is_empty()`/`!enabled` return still protects the spawn.

**Option B**: Make callers the owners. Drop the `enabled` parameter from `record_file_ops` entirely—callers already guard, so no need to re-test.

Apply the same convention to `record_with_counts` (`mod.rs:1031`) for symmetry.

---
*[complexity review]*

dean0x:
## [Complexity + Rust Review] stdin "unreachable in practice" branch risks silent PF-001 re-introduction

**File**: `crates/rskim/src/main.rs`  
**Lines**: ~927-943  
**Confidence**: 82% (both reviewers flagged)

### Issue
The `record_file_analytics` stdin adapter relies on a cross-file invariant:

```rust
FileCounts::Tokenize => {
    let Some(buf) = result.stdin_raw else { return }; // "unreachable in practice"
    ...
}
```

The invariant is that `stdin_raw` is `Some` if and only if `show_stats` was false. This invariant is enforced in `process_stdin` (`process.rs:423-427`):
- `stdin_raw = Some(buffer)` exactly when `!show_stats`
- `original_tokens`/`transformed_tokens` are `Some` exactly when `show_stats`

**Problem**: This invariant is implicit, documented only in a 4-line comment, and split across two files. A future change to either `show_stats` block could silently turn the "unreachable" path into real, invisible analytics loss—re-introducing PF-001 specifically for stdin.

### Impact
Correctness is sound today (the invariant holds). Maintainability risk: future changes have no compile-time signal that they break the contract.

### Suggested Fix
Make the invariant explicit at the boundary where the value is produced:

**Option A** (Recommended): Add a `debug_assert!` in `process_stdin` to document the invariant in code:
```rust
debug_assert!(stdin_raw.is_some() != (original_tokens.is_some() && transformed_tokens.is_some()),
    "stdin_raw and token counts must be mutually exclusive");
```

**Option B** (Larger refactor): Encode the invariant in the type. Instead of three loosely-coupled `Option` fields on `ProcessResult`, carry a single `enum Counts { Known(usize, usize) | NeedsTokenize(RawSource) }`. This makes the illegal state unrepresentable.

---
*[complexity+rust review]*

dean0x:
## [Documentation Review] Orphaned analytics.db history not disclosed in user-facing docs

**File**: `CHANGELOG.md` (lines ~24-27), `CLAUDE.md` (lines ~100-108)  
**Confidence**: 92%

### Issue
The CHANGELOG "Behavior change" block and CLAUDE.md correctly document that `SKIM_CACHE_DIR` now relocates the default `analytics.db`. However, **neither mentions that pre-existing analytics history under `~/.cache/skim/analytics.db` is left behind (orphaned), not migrated**, when a user first sets `SKIM_CACHE_DIR`.

The PR description itself acknowledges this: "old history orphaned not migrated." However, this caveat did not make it into the user-facing documentation.

### Impact
A user who sets `SKIM_CACHE_DIR` after having used skim will silently observe their `skim stats` history reset to empty (the dashboard now reads the relocated, fresh DB). Without the caveat, this looks like data loss or a bug.

This is exactly the class of "behavior change that the changelog must call out"—surprising behavior must be documented.

### Suggested Fix
Add one sentence to the CHANGELOG behavior-change block and the CLAUDE.md `SKIM_CACHE_DIR` entry:

> Existing analytics history under the previous location (`~/.cache/skim/analytics.db`) is **not migrated**—it is left in place and a fresh `analytics.db` is created at the new location. Point `SKIM_ANALYTICS_DB` at the old file to retain prior history, or copy it manually.

---
*[documentation review]*

dean0x:
## [Testing Review] F14 test does not exercise the path it documents

**File**: `crates/rskim/tests/cli_analytics_expansion.rs`  
**Lines**: ~968-1020 (test name and body)  
**Confidence**: 90%

### Issue
F14's doc comment claims it verifies "one file deleted **between transform and background re-read** … deleted/changed files are silently skipped."

However, the test deletes `ghost.ts` **before** invoking skim (line ~988), so the path fails during the initial `process_explicit_files` call and never reaches the background `record_file_ops` re-read. The test's own inline comment at lines ~1013-1014 admits this: "Ghost file is not found before process_file is even called."

The genuine best-effort skip branch—`RawSource::Reread(p) => crate::process::read_source(&p).ok()?` (analytics/mod.rs:1102), which handles a file deleted/changed **after dispatch but before the background re-read**—has **no integration coverage**.

The 50 MB guard rejection of a TOCTOU-grown file is also unverified.

### Impact
A regression that breaks the background-thread skip-on-read-error (e.g., changing `.ok()?` to `.unwrap()`) would panic the background thread in production yet pass this test suite.

### Suggested Fix
**Preferred**: Add a test that forces the `Reread` skip—e.g., a `#[cfg(unix)]` test whose `RawSource::Reread` path points at:
- A file made unreadable (chmod 000) after dispatch, OR
- A file replaced with >50MB content after dispatch

Assert that sibling rows still land and the process exits 0 without panic.

**Alternative**: Re-document F14 to reflect what it actually tests (a file that fails during arg resolution → N-1 rows, no crash), or rename it to something more accurate like `F14_early_dispatch_failure`.

---
*[testing review]*

dean0x:
## Summary: Code Review Complete

**PR**: #360 | **Base**: main | **Branch**: fix/analytics-loss-and-cache-dir  
**Review Date**: 2026-06-25 | **Artifact**: `.devflow/docs/reviews/fix-analytics-loss-and-cache-dir/2026-06-25_1029/`

---

### Verdict
- **Blocking Issues**: 0  
- **Should-Fix (High Confidence)**: 5 inline comments above  
- **Recommendations**: 11/11 reviewers APPROVED or APPROVED_WITH_CONDITIONS  
- **Status**: **MERGEABLE** with should-fix follow-ups

---

### Should-Fix Summary (from inline comments)

| Focus | Issue | File | Confidence |
|-------|-------|------|-----------|
| Performance + Database | Per-row DB open + prune | analytics/mod.rs:~1127 | 88% |
| Complexity | Doubled `enabled` guard | analytics/mod.rs + multi.rs + main.rs | 88% |
| Complexity + Rust | stdin unreachable branch risk | main.rs:~927 | 82% |
| Documentation | Orphaned analytics.db not disclosed | CHANGELOG.md + CLAUDE.md | 92% |
| Testing | F14 test misleading | cli_analytics_expansion.rs:~968 | 90% |

---

### Lower-Confidence Suggestions (60–79% — Summary comment only)

1. **File-size guard check timing** (Reliability, 88%) — Pre-read stat check before full allocation in `read_source` to avoid transient `num_cpus × filesize` memory spikes on large directories.

2. **Uncapped background tokenization** (Reliability, 80%) — Per ADR-001, apply run-cap fallback before tokenizing in the recorder or expose a `count_tokens_bounded` variant.

3. **`flush_pending` join unbounded** (Reliability, 90% — pre-existing) — Now that background work scales with directory size, bind the exit join with a deadline (was previously acknowledged for single-file case; multi-file changes the calculus).

4. **Whitespace-only SKIM_CACHE_DIR** (Rust, 70%) — Doc says "empty/whitespace treated as unset," but code only checks `.is_empty()`, not `.trim()`. Either trim or clarify the wording.

5. **Empty SKIM_ANALYTICS_DB asymmetry** (Database, 70%) — Unlike `SKIM_CACHE_DIR`, empty `SKIM_ANALYTICS_DB=""` is used as a literal path. Mirror the `SKIM_CACHE_DIR` guard for symmetry.

6. **Duplicate env read behind unified resolver** (Architecture, 85%) — `cache_root()` and `CacheEnv::from_process()` each read `SKIM_CACHE_DIR` independently. Consolidate the env read into one function both call, so the only difference is injected-vs-process.

7. **`summary.invocations` metric semantics shift** (Regression, 88%) — Multi-file runs now contribute N rows instead of 1 aggregate. Document this in CHANGELOG so historical metric comparisons across upgrade are understandable.

---

### Reviewer Consensus

**APPROVED**: 2 reviews  
**APPROVED_WITH_CONDITIONS**: 9 reviews (conditions are the 5 should-fix items above + suggested lower-confidence follow-ups)

**Key findings by focus**:
- **Security**: 9/10 — no SQL injection, no path traversal, TOCTOU-safe, owner-only perms.  
- **Performance**: 8/10 — streaming contract intact, hot path unregressed; exit latency O(N) for large dirs (should-fix).  
- **Database**: 9/10 — safe concurrency (WAL+busy_timeout), parameterized queries, efficient for single-file; per-row open/prune is the one should-fix.  
- **Reliability**: 8/10 — bounded thread count, no panic, poison-tolerant join; should-fix on pre-read sizing and unbounded tokenization.  
- **Architecture**: 8/10 — PF-001/PF-002 fixes are clean; streaming contract preserved; cache-root genuinely single-source.  
- **Regression**: 9/10 — migration complete, no unintended breakages; only the metrics-semantics shift needs CHANGELOG callout.  
- **Testing**: 9/10 — behavior-focused, deterministic, harness isolation correct; F14 test misleading (should-fix above).  
- **Consistency**: 9/10 — harness migration uniform, idioms match established patterns, all agreed-upon renames correct.  
- **Complexity**: 8/10 — no blocking issues; the double guard and implicit stdin invariant are clarification opportunities.  
- **Documentation**: 9/10 — unusually accurate; one caveat (orphaned history) missed user-facing docs.  

---

### Next Steps

1. Address the 5 high-confidence should-fix items via inline comments above.
2. Track the 7 lower-confidence suggestions for follow-up PRs (these do not block merge).
3. Run local test suite to confirm the 5 should-fixes do not introduce regressions.
4. Merge once should-fixes are addressed.

Review reports available at: `/Users/dean/Sandbox/skim-issues/.devflow/docs/reviews/fix-analytics-loss-and-cache-dir/2026-06-25_1029/`

---
*This summary was generated by the Git agent comment-pr operation. All findings are deduplicated across 11 parallel reviewers.*

dean0x:
## Tech Debt Update

Two pre-existing issues surfaced during the PR #360 code review have been added to the **Tech Debt Backlog** (issue #13):

### Items Deferred
- **#21** (HIGH): Unbounded `flush_pending` join with no timeout in analytics/mod.rs. PR #360 increased priority by expanding work from O(1) to O(directory size).
- **#22** (MEDIUM): `AnalyticsStore` trait abstracts reads but not writes, blocking injectable recording for unit tests.

Both are pre-existing design patterns, not regressions introduced by this PR. Review details: [resolution summary](.devflow/docs/reviews/fix-analytics-loss-and-cache-dir/2026-06-25_1029/resolution-summary.md).

See #13 for full backlog context.

PR #367, opened 2026-06-26
fix: markdown heading order, nextest rewrite, pseudo visibility, ls double-header

dean0x:
## Summary

Four independent bug fixes across `rskim-core` and `rskim`, each with TDD (failing test → implementation → passing):

- **Markdown headings emitted in reverse order** — LIFO visit stack returned siblings in reverse source order; fix sorts by `source_start_line` before output
- **`cargo nextest run` dropped the `run` token** — rewrite rule stripped `run`; fragile `is_nextest` sniff replaced with explicit parameter; nextest exit code corrected to `100`
- **Pseudo mode stripped visibility/export modifiers** — `pub`, `export`, `public`, `private`, etc. are API surface, not noise; removed from all per-language strip lists
- **`ls -la` double-counted `.`/`..` and emitted two headers** — dotdirs now skipped before counting; single footer replaces the prepended summary entry

### FIX 1 — Markdown heading order (`rskim-core`)
- `crates/rskim-core/src/transform/structure.rs`: added `headers.sort_by_key(|h| h.2)` before the texts/spans/line-map pipeline
- Tests: updated `test_markdown_line_map_consecutive_headers` (LIFO→doc-order); added `test_markdown_headers_document_order` (unit) and `test_markdown_headings_in_document_order` (E2E)

### FIX 2 — `cargo nextest run` rewrite (`rskim`)
- `cmd/rewrite/rules.rs`: `rewrite_to` now includes `"run"` so `cargo nextest run` → `skim cargo nextest run`
- `cmd/rewrite/engine.rs`: updated `test_cargo_nextest_run` assertion
- `cmd/test/mod.rs`: explicit `is_nextest = runner_args.first() == Some("nextest")` (A2 contract)
- `cmd/test/cargo.rs`: extracted `build_cargo_args(args, is_nextest)` helper; nextest exit code `100`, libtest `101`
- Tests: added 2 E2E rewrite tests, 4 `build_cargo_args` unit tests

### FIX 3 — Pseudo mode preserves visibility (`rskim-core`)
- `cmd/transform/pseudo.rs`: removed visibility keywords/node-kinds from all per-language `PseudoRules` strip lists (TS/JS `export`; Rust `visibility_modifier`; Java/C#/Ruby/Kotlin/Swift access modifiers); C++ `access_specifier` special case changed from removal to pass-through
- Tests: all "strips visibility" assertions inverted → "preserves visibility"; added verification tests for non-visibility modifiers still stripped (`static`/`final`/`open`/etc.) across csharp/kotlin/swift/ruby transform test files; integration.rs updated for C++/Java/TypeScript/Rust/auto-detection pseudo tests (A4 contract)

### FIX 5 — `ls -la` dotdir count + double header (`rskim`)
- `cmd/file/ls.rs`: skip `.` and `..` before counting (trims `-F`/`-p` trailing `/`); remove prepended `summary_entry`; footer folded to `"D dirs, F files"` or `"… — D dirs, F files"` when elided; empty-dir → Full result with 0 entries (not None)
- Tests: updated `test_tier1_ls_la` (count assertion); added 4 regression tests

## Breaking Changes

None. `FileResult` public type and `Display` contract unchanged. Exit codes (`0`/`1`/`2`/`3`) unchanged. `--mode` values unchanged.

## Reviewer Focus Areas

- **FIX 3 completeness**: All 8 affected languages updated? Check `pseudo.rs` `get_pseudo_rules` for any language that still has visibility tokens in `strip_keywords` / `strip_kinds`.
- **FIX 5 empty-dir path**: `has_long_lines` guard correctly distinguishes "no permission lines at all" (→ `None`, fall to Tier-2) from "permission lines but all were dotdirs" (→ `Full` with 0 entries). Verify `test_tier1_ls_la_empty_dir`.
- **FIX 2 exit codes**: nextest exits `100` on failure, libtest exits `101`. The old code used `101` for nextest. Confirm `expected_exit_codes` branch in `cargo.rs`.
- **ADR-001 compatibility**: FIX 5 footer path preserves the `elision_marker` call — net-savings guard is not bypassed.

### DIFF
--- a/crates/rskim/src/cmd/file/ls.rs
+++ b/crates/rskim/src/cmd/file/ls.rs
@@ -120,6 +120,18 @@ fn parse_ls(output: &CommandOutput) -> ParseResult<FileResult> {
 }
 
 /// Tier 1: long-form `ls -la` output — detect permissions, count dirs vs files.
+///
+/// Skips `.` and `..` dotdir entries so they are excluded from counts and from
+/// the display list.  `-F`/`-p` append a trailing `/` to directory names; names
+/// are trimmed before the dot-dir comparison.
+///
+/// Dir/file counts are folded into the FOOTER rather than a prepended summary
+/// entry, eliminating the duplicate `ls N` + `LS: N entries` double-header that
+/// appeared on large directories (A3 contract: FileResult public shape unchanged).
+///
+/// Empty dirs (only `total`/`.`/`..` lines): returns a Full result with 0 entries
+/// rather than None, preventing Tier-2 from mis-tokenising the `total`/`.`/`..`
+/// lines as plain filenames.
 fn try_parse_ls_long(stdout: &str) -> Option<FileResult> {
     let mut dirs = 0usize;
     let mut files = 0usize;
@@ -130,6 +142,18 @@ fn try_parse_ls_long(stdout: &str) -> Option<FileResult> {
         if !RE_LS_LONG.is_match(line) {
             continue;
         }
+
+        // Skip `.` and `..` dotdir entries.  The name is the last whitespace-separated
+        // token; trim a trailing `/` first (added by `-F`/`-p` flags).
+        match line
+            .split_whitespace()
+            .last()
+            .map(|s| s.trim_end_matches('/'))
+        {
+            Some(".") | Some("..") => continue,
+            _ => {}
+        }
+
         line_count += 1;
         if line.starts_with('d') {
             dirs += 1;
@@ -141,23 +165,35 @@ fn try_parse_ls_long(stdout: &str) -> Option<FileResult> {
         }
     }
 
-    if line_count == 0 {
+    // Empty dir after skipping dotdirs: return a well-formed Full result (0 entries,
+    // "0 dirs, 0 files" footer) instead of None.  Returning None would fall through
+    // to Tier-2 which mis-tokenises `total`/`.`/`..` lines.
+    // Note: we can only reach here if at least one line matched RE_LS_LONG (including
+    // the dotdir lines we skipped), so we need to check whether the original output
+    // had any matching lines at all before declaring it empty.
+    let has_long_lines = stdout
+        .lines()
+        .take(MAX_INPUT_LINES)
+        .any(|l| RE_LS_LONG.is_match(l));
+    if !has_long_lines {
         return None;
     }
 
     let shown_count = entries.len();
-    let footer = crate::output::elision_marker(shown_count, line_count, "entries");
 
-    let summary_entry = format!("LS: {line_count} entries ({dirs} dirs, {files} files)");
-    // Prepend summary as first entry
-    let mut all_entries = vec![summary_entry];
-    all_entries.extend(entries);
+    // Build footer: fold the dir/file breakdown into the elision marker (if any),
+    // or emit just "D dirs, F files" when everything fits.
+    let breakdown = format!("{dirs} dirs, {files} files");
+    let footer = match crate::output::elision_marker(shown_count, line_count, "entries") {
+        Some(elision) => Some(format!("{elision} — {breakdown}")),
+        None => Some(breakdown),
+    };
 
     Some(FileResult::new(
         "ls".to_string(),
         line_count,
         shown_count,
-        all_entries,
+        entries,
         footer,
     ))
 }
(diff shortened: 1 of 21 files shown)


crates/rskim/src/cmd/test/cargo.rs
```diff
@@ -44,22 +44,64 @@ static RE_CARGO_SUMMARY: LazyLock<Regex> = LazyLock::new(|| {
...
+/// flag is threaded explicitly from the dispatcher — never sniffed from args.
+pub(crate) fn build_cargo_args(args: &[String], is_nextest: bool) -> Vec<String> {
+    let mut cmd = if is_nextest {
+        Vec::new()
+    } else {
+        vec!["test".to_string()]
+    };
+    cmd.extend(args.iter().cloned());
+    cmd
+}
```
dean0x, line 66: ## 🚨 BLOCKING: `cargo test nextest` misrouted to `cargo nextest` (FIX 2 F2c regression)

The new `is_nextest` detection in `test::run` (`mod.rs:66`) is positional:
```
runner_args.first() == Some("nextest")
```

This catches `cargo nextest run` correctly, but silently misroutes **`cargo test nextest`** (a bare filter name, legitimate for running tests named "nextest"):

**Trace:**
- `cargo test nextest` → `dispatch_cargo` → `prepend_without("cargo", args, 0)` strips `"test"` → `test::run(["cargo","nextest"])`
- `runner_args.first() == Some("nextest")` → `is_nextest=true`
- `build_cargo_args(["nextest"], true)` returns `["nextest"]` (no `"test"` prefix)
- **Wrong:** spawns `cargo nextest` instead of `cargo test nextest`

**Impact:** violates #317 ("never emit a command that changes semantics"). The trigger is narrow but real in *this* repo (has nextest-named tests like `test_tier1_nextest_pass`). The F2c guard at line 1234 doesn't catch this—it only tests `--test nextest` (a flag form where first arg is `--test`, not `nextest`).

**Fix:** Thread `is_nextest` from `dispatch.rs:381` (the "nextest" arm that already knows it) instead of re-deriving it in `test::run`. Either:
1. Add `is_nextest: bool` parameter to `test::run`; dispatch arm at :381 passes `true`, `:378` passes `false`
2. Use an enum (`CargoRunner::{Test, Nextest}`) instead of `bool` for type safety

**Also add:** a regression test for `cargo test nextest` (positional filter) that drives **production dispatch** (not an inline logic copy), asserting the built argv contains `["test","nextest"]`.

---
_Review: rust.md (90%); corroborated by architecture/security/regression._

crates/rskim/src/cmd/file/ls.rs
```diff
@@ -120,16 +120,45 @@ fn parse_ls(output: &CommandOutput) -> ParseResult<FileResult> {
...
+        // Skip `.` and `..` dotdir entries.  The name is the last whitespace-separated
+        // token; trim a trailing `/` first (added by `-F`/`-p` flags).
+        if matches!(
+            line.split_whitespace()
+                .last()
+                .map(|s| s.trim_end_matches('/')),
+            Some(".") | Some("..")
+        ) {
+            continue;
+        }
```
dean0x, line 164: **FIX 5: Symlink and space-named entries silently mis-skipped by dotdir detection**

`ls -la` symlinks render as `name -> target`. For a symlink like `lrwxr-xr-x ... mylink -> ..`, the code does:
```rust
line.split_whitespace().last() // Returns "..", the TARGET, not the link name
```

**Impact:** A symlink pointing to `.` or `..` is silently dropped from counts and display — violates the project reliability rule "fail loud, never silently."

Example: a directory with only a symlink `link -> ..` incorrectly renders as an empty directory.

**Fix:** Strip the symlink target first (`split on " -> "` then take first part), then extract the name:
```rust
let name_field = line.split(" -> ").next().unwrap_or(line);
let name = name_field.split_whitespace().last().map(|s| s.trim_end_matches('/'));
if matches!(name, Some(".") | Some("..")) { continue; }
```

Also add a test fixture covering `lrwxr-xr-x ... link -> ..` to prevent regression.

---
_Review: reliability.md (82% MEDIUM confidence); corroborated by rust/regression/security_

crates/rskim/src/cmd/file/ls.rs
```diff
@@ -141,23 +170,27 @@ fn try_parse_ls_long(stdout: &str) -> Option<FileResult> {
...
     let shown_count = entries.len();
-    let footer = crate::output::elision_marker(shown_count, line_count, "entries");
 
-    let summary_entry = format!("LS: {line_count} entries ({dirs} dirs, {files} files)");
-    // Prepend summary as first entry
-    let mut all_entries = vec![summary_entry];
-    all_entries.extend(entries);
+    // Build footer: fold the dir/file breakdown into the elision marker (if any),
+    // or emit just "D dirs, F files" when everything fits.
+    let breakdown = format!("{dirs} dirs, {files} files");
```
dean0x, line 187: **FIX 5: Inconsistent footer wording — "dirs" vs "directories" in the same file**

This line (FIX 5's new footer) uses:
```rust
let breakdown = format!("{dirs} dirs, {files} files");
```

But the sibling tree footer at line 307 uses:
```rust
summary = Some(format!("{dirs} directories, {files} files"));
```

The project explicitly values output-string uniformity (see `output/mod.rs:400-410` canonical-unit table). Two footer builders in one file should use the same term.

**Fix:** Choose one term across both footers. Recommend `"directories"` for consistency with standard `ls` output, or update line 307 to match.

---
_Review: consistency.md (80% MEDIUM confidence)_

dean0x:
## PR Review Summary

**Branch:** fix/skim-issue-batch → main | **PR:** #367  
**Reviewed:** 6 reports (4021+ tests across rust/architecture/documentation/consistency/reliability/regression/testing)

---

### Inline Comments (≥80% confidence)

Created **6 blocking/should-fix comments** on the following findings:

1. **BLOCKING:** `cargo test nextest` misrouted to `cargo nextest` (FIX 2, F2c regression)
2. **FIX 1:** Stale inline comment about line_map order
3. **FIX 5:** Symlink/space-named entries silently mis-skipped  
4. **FIX 3:** Vestigial `visibility_modifier` dead code in `is_inline_modifier_kind`
5. **FIX 5:** Footer string inconsistency ("dirs" vs "directories")
6. **⚠️ Not in diff:** FIX 3 leaves `README.md:243`, `main.rs:367` (help text), and `docs/modes.md` claiming pseudo still strips visibility — this contradicts the code and should be updated per `documentation.md` (96% CRITICAL)

---

### Medium-Confidence Findings (60–79%, for follow-up)

1. **FIX 2 passing-nextest-run drops stderr summary** (~75%, regression.md): On success (exit 0), nextest writes its entire report to stderr, but `forward_stderr: false` drops it → near-empty output. Brushes #317 on the success path. Related to newly-created PF-003; candidate follow-up.

2. **FIX 1 setext/mixed-style markdown order untested** (~75%, testing.md): The sort specifically reasons about setext stability, but all 5 order tests use ATX only. Add position assertions to `test_markdown_mixed_header_styles` (already has the fixture).

3. **CHANGELOG leaks internal "A2/A4 contract" labels** (~72%, consistency.md): user-facing release notes cite unresolvable internal design-doc IDs alongside resolvable refs like `#317`/`#359`. Consider rephrasing for reference-style consistency.

4. **`is_nextest: bool` boolean-blindness** (~62%, rust.md/architecture.md): Call sites like `build_cargo_args(args, true)` are opaque. Consider enum `CargoRunner::{Test, Nextest}` for type safety.

5. **Kotlin/Ruby thin visibility coverage** (~65–70%, testing.md): Kotlin test asserts only `private`; Ruby test doesn't assert `protected`/`public`. Add fixture coverage.

---

### Assessment

This is a **tight, well-tested batch** (TDD discipline visible; 4021+ tests; behavior-focused assertions; no implementation coupling):
- **FIX 1 (markdown order):** Clean sort, full migration of tests, E2E coverage. Safe.
- **FIX 2 (nextest):** Routing is correct; exit codes match design; **one verified #317 regression on the positional-filter path**. Mitigation: thread `is_nextest` from dispatch site.
- **FIX 3 (pseudo visibility):** Surgical removal from all 8 languages; non-visibility modifiers still stripped; Kotlin-`open` vs Swift-`open` distinction correct. Safe (in-code docs accurate); **external docs need sync**.
- **FIX 5 (ls dotdir):** Integer math safe; ADR-001 preserved; **one symlink→`.`/`..` false-skip edge**.

---

### Recommendation

**CHANGES_REQUESTED** — pending:

**Blocking (must fix before merge):**
- [ ] FIX 2 F2c: thread `is_nextest` from dispatch rather than re-derive; add production-dispatch regression test

**Should-fix (blocks ship):**
- [ ] FIX 3: update `docs/modes.md`, `README.md:243`, and help text to reflect visibility preservation
- [ ] FIX 1: update stale line_map comment (line 978)
- [ ] FIX 5: fix symlink target extraction before dotdir skip; add fixture

**Nice-to-have (separate PR OK):**
- [ ] FIX 1: add setext-order test coverage
- [ ] FIX 2: verify passing-nextest output (forward stderr or accept as-is)
- [ ] Vestigial `visibility_modifier` dead-code cleanup
- [ ] Footer wording consistency
- [ ] CHANGELOG internal-label rephrasing

---

**Inline comments include detailed fix suggestions and code snippets.**

---
_Review: 10 reviewers; highest confidence item (BLOCKING) is 90% (rust.md, corroborated); lowest (nice-to-have) is 62% (boolean-blindness). The narrow trigger (bare `cargo test nextest` filter) and loud failure (non-zero exit) are why it's HIGH rather than CRITICAL._

PR #371, opened 2026-06-27
fix: output-fidelity batch — grep double-header, stdout-to-file rewrite bail (#370)

### LINKED ISSUE #370: Tracking: grep/rg double-header + rewrite stdout-redirect fidelity (output-fidelity batch)
## Summary

Two **validated** output-fidelity defects found during a verification session (2026-06-27), plus one environment caveat. Both defects are byte-/output-fidelity regressions in the spirit of #317 ("compress, never truncate; byte-faithful or bail"). They live in **different modules** and are each small, so the plan is to fix **both in one PR** (separate commits), then close this tracker.

- **D1** — `skim grep`/`rg`: redundant double-header + wrong pluralization (`1 files`).
- **D2** — rewrite engine rewrites commands whose **stdout is redirected to a file** (`… > out.json`), corrupting the file with a skim summary instead of raw output.

> Scope note: the existing tech-debt issues **#362 / #363 / #369** are **not** folded in here (independent, larger). They are listed under *Related* for visibility. The PATH-clone hazard (C1) is an environment caveat, not a Rust fix.

---

## D1 — `grep`/`rg` double-header + pluralization

**Validated** against this clone (`./target/release/skim`, 2.10.0):

```
$ ./target/release/skim grep -rn ' — ' crates/rskim/src/cmd/file/ls.rs
grep 9
 GREP: 9 matches in 1 files      ← (a) redundant 2nd header   (b) "1 files" should be "1 file"
 crates/rskim/src/cmd/file/ls.rs
   :8: ...
```

**Root cause** — two layers each emit a header:
- `FileResult::render` (`crates/rskim/src/output/canonical.rs:1011-1038`) emits the canonical `grep N` header (intended; Fix F).
- `build_file_result` (`crates/rskim/src/cmd/file/mod.rs:201-230`) **prepends** a second summary as `entries[0]`:
  `format!("{}: {total_matches} matches in {file_count} files", …)` — `files` hardcoded, no pluralization.

This is the **exact pattern FIX 5 removed for `ls`** (`ls N` + `LS: N entries`) — see `crates/rskim/src/cmd/file/ls.rs:128-199`. `grep`/`rg` were simply not included in that fold.

**Surface / blast radius:** output handler only. `build_file_result` is shared by `grep.rs:312` **and** `rg.rs:139`, so **both** tools are affected and **both** are fixed by one change. (`rg` couldn't be run live — ripgrep isn't installed locally — confirmed by code.)

**Fix (mirror ls FIX 5):** drop the prepended `TOOL: …` summary; fold the file count into the **footer** with correct pluralization. Target output:

```
grep 9
 crates/rskim/src/cmd/file/ls.rs
   :8: ...
 1 file
```
```
grep 25
 a.rs
   :3: ...
 b.rs
   :7: ...
 3 files
```

Spec: header stays `grep {total_matches}`; footer = `{file_count} file` when 1 else `{file_count} files`. (Open: whether to suppress the footer for the single-file case — implementer's call; ls always shows it.)

**Tests:** `crates/rskim/src/cmd/file/grep.rs:421 test_summary_line_present` asserts `rendered.contains("matches in")` — update to assert the new footer / single-header. `rg` tests assert on counts + match presence, so they don't break; add a footer/header assertion for parity.

### D1 checklist
- [ ] Rewrite `build_file_result` (`cmd/file/mod.rs`): no prepended summary; pluralized file-count footer
- [ ] Update `grep.rs::test_summary_line_present` + add an `rg` parity test
- [ ] Verify no other test asserts `"matches in"` / `"GREP:"` / `"RG:"`
- [ ] Confirm #317 honored (every match still printed; output only shrinks)

---

## D2 — rewrite engine must not rewrite stdout-redirected-to-file commands

**Validated** — reported case (`gh api … > dependabot-alerts.json` produced a 76-byte skim summary in the file instead of the JSON API response). Reproduced via `skim rewrite`:

| Command (input) | Current output | Correct? |
|---|---|---|
| `gh api repos/o/r/x > out.json` | `skim gh api … > out.json` | ❌ **BUG** |
| `gh api … >> out.json` | `skim … >> out.json` | ❌ **BUG** |
| `gh api … 1> out.json` | `skim … 1> out.json` | ❌ **BUG** |
| `grep -rn foo src/ > matches.txt` | `skim grep … > matches.txt` | ❌ **BUG** |
| `gh api … > /dev/null` | `skim … > /dev/null` | ⚠️ harmless (discarded) but pointless |
| `gh api … 2> err.txt` | `skim … 2> err.txt` | ✅ OK — stdout still reaches the agent |
| `gh api … 2>&1` | `skim … 2>&1` | ✅ OK — no stdout→file |
| `gh api … \| jq .` | *(passthrough)* | ✅ already excluded (pipe) |

**Root cause** — AD-RW-2 (`crates/rskim/src/cmd/rewrite/compound.rs:6-14`) **strips** redirects to match rules, then **splices them back** at their original position. That preserves the redirect *positionally* but **changes semantics**: `skim` is interposed, so the file now captures skim's compressed summary instead of the tool's raw bytes. There is a corruption guard (`rewrite_would_corrupt` / `command_needs_passthrough`, `compound.rs:39-85`) but it does **not** treat a clean stdout→file redirect as a bail condition.

This violates the CLAUDE.md MUST: *"rewrites must reconstruct the command byte-faithfully or bail (never emit a command that … changes semantics)."* It is the natural sibling of the already-implemented **pipe exclusion** (AD-RW-2): when the tool's stdout is consumed downstream (pipe) or captured (file), skim must not interpose.

**Fix:** add a bail/passthrough predicate for **stdout (fd 1) redirected to a file**. Precise rule (reuse existing redirect parsing — `is_single_redirect`, `strip_segment_redirects` in `compound.rs`):
- **Bail** on `>`, `>>`, `1>`, `1>>`, `&>`, `&>>` whose target is a **path** (incl. `/dev/null` — over-bail = a harmless missed optimization, consistent with "over-bailing only costs a missed optimization").
- **Do not bail** on `2>` / `2>>` (stderr only — stdout still reaches the agent) or fd-dups like `2>&1`, `>&1` (no file target).
- Wire it into the same passthrough decision as the pipe exclusion so compound segments (`a > f.json && b`) are handled per-segment.

**Surfaces (per CLAUDE.md "two interception surfaces"):**
- ✅ **Rewrite engine / PreToolUse hook** — the validated, reported surface. This fix covers it.
- ❓ **PATH wrappers** — if a wrapped tool (`gh`/`grep`/…) runs with `> file.json`, `skim`-as-tool would still write compressed output to the file. `is_terminal()` can't distinguish a pipe (keep compressing — skim's core use case) from a file redirect, so the wrapper fix needs `fstat`-style regular-file detection on fd 1, NOT tty detection. **Separate sub-task** — audit and fix or split to a follow-up.

### D2 checklist
- [ ] Add stdout→file detection (reuse `compound.rs` redirect parsing); bail/passthrough
- [ ] Wire into the pipe-exclusion passthrough path (per-segment for compound commands)
- [ ] Tests: full matrix above (bug rows bail; stderr/fd-dup/pipe rows unaffected); compound `a > f && b`
- [ ] Audit the **PATH-wrapper** surface (regular-file stdout) → fix here if cheap, else open a follow-up issue and link it
- [ ] Confirm rule families unaffected when stdout is NOT redirected (regression sweep on rewrite tests)

---

## C1 — PATH-clone hazard (environment caveat, no Rust fix in this batch)

During validation, `which -a skim` in the Bash tool resolved to `/Users/dean/Sandbox/skim/target/release/skim` (a **different, stale clone**) and `~/.cargo/bin/skim` — **not** this clone. Hook-rewritten bash commands therefore ran the wrong binary (observed: a stale `LS: N entries` double-header that FIX 5 already removed here). The working-memory assumption that `.zshrc:202` puts `skim-issues` first does **not** hold in the Bash tool's non-interactive env.

This is environmental, not a skim code bug. Mitigations (no code change required):
- [ ] Always validate via explicit `./target/release/skim` (never trust hook-rewritten `skim`/`grep`/`ls` output)
- [ ] Optionally pin the PreToolUse hook to an absolute path for the active clone, or sync `~/.cargo/bin/skim` to this clone before a verification session
- [ ] (Optional, stretch) consider a `skim`-side debug notice when the running binary's origin ≠ cwd project root

---

## Related — separately tracked tech debt (NOT in this batch)
- #362 — analytics: bound `flush_pending()` thread join with a deadline (currently unbounded) — `safety`
- #363 — analytics: `AnalyticsStore` abstracts reads but not writes (recording not injectable) — `foundation`
- #369 — `cargo nextest run` (exit 0) drops the stderr summary — `tech-debt`

---

## Execution plan (next session)
- Branch off `main`: `fix/output-rewrite-fidelity-batch` (do **not** commit to `main`).
- Commit 1 = D1 (output handler); Commit 2 = D2 (rewrite engine). Independent modules → no conflict.
- Tests (respect the per-crate build caps in CLAUDE.md): `cargo nextest run -p rskim --bins -j 4` + `cargo test -p rskim --doc`. Avoid `--all-features`/`--workspace` inside agents.
- `cargo clippy -p rskim -- -D warnings` + `cargo fmt -- --check`.
- One PR closing this tracker; Snyk red = known account-quota issue (not a code failure).

## Acceptance criteria
- `skim grep`/`skim rg` emit a single header and correct singular/plural (`1 file`).
- `skim rewrite '… > f.json'` (and `>>`, `1>`, `&>`) returns the original command unchanged; `2>`, `2>&1`, pipes, and non-redirected commands are unchanged from today.
- PATH-wrapper surface for D2 either fixed or split into a linked follow-up (no silent gap).
- Full `rskim` bins + doctests green; clippy/fmt clean.

dean0x:
Closes #370.

## Summary

Three output-fidelity defects fixed across both interception surfaces (rewrite engine and PATH wrappers):

- **D1** — `skim grep` / `skim rg` emitted a double-header (`GREP: N matches in M files` prepend + canonical `tool N` header) with a mis-pluralized file count
- **D2** — `skim rewrite` and the PreToolUse hook rewrote commands that redirect stdout to a file (e.g. `gh api repos/o/r/x > out.json`), causing skim's compressed summary to land in the file instead of the tool's raw bytes
- **D2b** — PATH-wrapper symlink invocations bypassed the rewrite engine entirely, so the same stdout-to-file corruption could occur via the wrapper surface

### D1 — grep/rg double-header + pluralized footer (`2ea9d87`)

**Before:**
```
GREP: 9 matches in 1 files     ← spurious prepend (wrong header, mis-pluralized)
grep 9                          ← canonical header
 src/foo.rs
   :8: // match
   ...
```

**After:**
```
grep 9                          ← canonical header only
 src/foo.rs
   :8: // match
   ...
 1 file                         ← pluralized footer (matches ls FIX 5)
```

`build_file_result` in `cmd/file/mod.rs` (shared by grep and rg) no longer prepends a summary entry. The file count moves to the footer with correct singular/plural form (`"1 file"` / `"N files"`).

### D2-A — Rewrite engine bails on stdout-to-file redirects (`c642ce4`)

**Before:**
```sh
$ printf 'gh api repos/o/r/x > out.json\n' | skim rewrite
skim gh api repos/o/r/x > out.json   # rewrite proceeds; out.json gets skim summary
$ echo exit=$?
exit=0
```

**After:**
```sh
$ printf 'gh api repos/o/r/x > out.json\n' | skim rewrite
                                      # empty stdout — bail
$ echo exit=$?
exit=1
```

New `stdout_redirected_to_file(cmd)` function in `cmd/rewrite/compound.rs`: quote-aware, single-pass O(n) detector wired into `rewrite_would_corrupt`. Catches `> f`, `>> f`, `>f`, `1> f`, `&> f`, `&>> f`, `>&f`. Skips stderr-only `2>`/`2>>`, fd-dups `>&1`/`>&2`/`>&-`/`1>&2`, quoted `>`, and pipes.

Safe commands still rewrite: `cargo test 2>&1`, `cargo test 2>/dev/null`, `cmd | jq`.

### D2b — PATH wrapper passes raw bytes when fd 1 is a regular file (`5c41a4a`)

**Before:**
```sh
# Shell does: fd 1 → out.json, then exec ~/.skim/bin/grep → skim binary
# skim compresses grep output and writes summary to out.json
grep -rn pattern src/ > out.json   # out.json contains skim header, not raw grep
```

**After:**
```sh
grep -rn pattern src/ > out.json   # out.json contains raw grep output
```

`stdout_is_regular_file()` in `main.rs` detects fd 1 pointing at a regular file using `ManuallyDrop<File::from_raw_fd(1)>` (borrows without closing). When true, argv0 dispatch calls `run_inherited_passthrough` instead of `cmd::dispatch`. Non-unix always returns false (no behaviour change).

**D2 covers BOTH interception surfaces — no follow-up needed.**

## Reviewer Focus Areas

- `stdout_redirected_to_file` CHECK ORDER (source-fd before target) is load-bearing — `2>&1` must be skipped before checking `>&1` would be a false bail
- `ManuallyDrop<File::from_raw_fd(1)>` safety: borrowing fd 1 without taking ownership; `// SAFETY:` comment explains this
- The `2>>` double-scan issue: when skipping `2>`, the inner branch advances `i` past the optional second `>` to prevent the `>>` case producing a spurious second iteration through the `>` character
- Test surface split: CLI bail = `.failure()` + empty stdout; hook bail = `.success()` + empty stdout — both surfaces tested independently

## Notes

The `security/snyk` check will appear red — this is a pre-existing account-quota issue (documented in team memory), not a code failure. Snyk local scan: 0 issues found.


dean0x:
Starting to post review findings...

crates/rskim/src/cmd/rewrite/compound.rs
```diff
@@ -80,11 +83,108 @@ pub(super) fn rewrite_would_corrupt(cmd: &str) -> bool {
...
+///
+/// This is a *visibility* bail, not a byte-faithfulness one — the command
+/// reconstructs fine; rewriting changes what reaches the file. Quote-aware: a
+/// `>` inside single or double quotes is ignored (no over-bail). Maximally
+/// strict: catches spaced/glued/append/fd-prefixed forms, `&>file`, `>&FILE`,
+/// and glued-middle `foo>out`. Skips `2>`/`2>>` (stderr — stdout still reaches
+/// the agent) and fd-dups (`>&1`, `>&2`, `1>&2`, `>&-`).
+///
+/// CHECK ORDER (source-fd before target) is load-bearing.
+fn stdout_redirected_to_file(cmd: &str) -> bool {
```
dean0x, line 107: **Security HIGH (85%) — false-negative paths re-open PF-004 corruption**

Two cases let real stdout→file redirects slip through:

**Case 1 — backslash-escaped quote outside quotes (`\'`):** The scanner only treats `\` as escape when `in_double` (line 118). Outside quotes, `\'` becomes a quote toggle (line 124), but bash treats `\'` as a *literal* single-quote that opens no context. The `>` in `grep x\' file > out z\'z` is skipped (scanner thinks inside quote), and file `out` receives skim's compressed summary instead of raw grep output.

**Case 2 — `>&<digit>filename`:** The fd-dup branch (line 333) checks only the first char after `&`, but bash treats `>&word` as dup only when `word` is *entirely* digits. So `cmd >&2x` (writes both to file `2x`) is misclassified as dup, returns false, file gets skim summary.

**Fix:**
- Case 1: Add `if !in_single && ch == '\\' { i += 2; continue; }` after line 275.
- Case 2: Check digit run is the whole target token (whitespace/end after). Otherwise bail.

**Verification:** Add to `test_corrupt_guard_stdout_to_file_bails`:
```
assert!(rewrite_would_corrupt(r"grep x\' file > out z\'z"), "backslash-escaped quote");
assert!(rewrite_would_corrupt("cmd >&2x"), ">&<digit>filename");
```

crates/rskim/src/main.rs
```diff
@@ -739,7 +777,15 @@ fn main() -> ExitCode {
     // parsing and route directly to the appropriate handler. PATH stripping
     // above ensures the handler won't find the symlink again (no recursion).
     let result: anyhow::Result<ExitCode> = if let Some((name, args)) = detect_argv0_dispatch() {
-        cmd::dispatch(&name, &args, &analytics)
+        // D2b (#370): when stdout is a regular file the shell has already
+        // redirected fd 1 to the file before exec-ing us. Run the real tool
+        // with inherited stdio so its raw bytes reach the file unmodified (#317).
+        // Pipes/ttys are not regular files → still compress (normal path).
+        if stdout_is_regular_file() {
```
dean0x, line 804: **Documentation HIGH (90%) — CLAUDE.md "Two interception surfaces" is now stale**

This code (D2b) adds stdout-to-file protection on the wrapper surface, but CLAUDE.md:85-93 contradicts the new behavior:

1. **Line 91 says wrapper routes "straight to `cmd::dispatch`"** — But your line 784-788 now intercepts `stdout_is_regular_file()` *before* dispatching. The wrapper branch no longer routes straight.

2. **Line 93 claims "fidelity guarantees do not apply to wrapper surface"** — But D2b adds fidelity: when fd 1 is a regular file, you run the tool with inherited stdio (raw bytes), not compressed output. The guarantee now exists on both surfaces via two mechanisms.

**Fix:** Update CLAUDE.md lines 91 and 93:
- Line 91: Note the wrapper branch checks `stdout_is_regular_file()` and bails to `run_inherited_passthrough` (raw bytes) when fd 1 is a regular file, *before* dispatching.
- Line 93: Clarify the *rewrite text-transformation* guarantees don't apply, but the stdout-to-file *output-fidelity* guarantee now exists on both surfaces (rewrite: `stdout_redirected_to_file`; wrapper: `stdout_is_regular_file`).

**Impact:** A future maintainer might trust the stale doc and "fix the gap" by removing this guard, regressing #370.

crates/rskim/src/cmd/file/mod.rs
```diff
@@ -193,6 +193,11 @@ pub(super) fn try_parse_file_line_content(
 /// elision footer. `shown_count == total_count` so the canonical header
 /// renders as `tool N` (no truncation ratio; Fix F). (#317)
 ///
+/// The file count is folded into the FOOTER (e.g. `"1 file"` / `"3 files"`)
```
dean0x: **Consistency MEDIUM (85%) — footer comment cites wrong precedent**

Your comment says pluralization mirrors "ls FIX 5", but ls does NOT pluralize. Check `ls.rs:187` — it emits `format!("{{dirs}} dirs, {{files}} files")` with "files" always literal (even for "0 files").

The real precedent is `diff.rs:215-216`, which uses the identical pluralization idiom you just added: `if file_count == 1 { "" } else { "s" }`.

**Fix:** Split the comment to attribute each pattern correctly:
```rust
/// The file count is folded into the FOOTER instead of a prepended
/// `TOOL: N matches in M files` double-header (the no-double-header pattern
/// from ls FIX 5). Pluralization ("1 file" / "3 files") follows the
/// `diff.rs` footer idiom — ls/tree intentionally do not pluralize. (#370, #317)
```

**Impact:** Maintainers reading "mirroring ls" will find non-pluralized output in ls.rs, contradicting the comment. The behavior itself is correct and consistent with diff.rs.

crates/rskim/src/cmd/rewrite/compound.rs
dean0x, line 107: **Complexity HIGH (85%) — function logic deserves refactoring**

The ~80-line loop body interleaves four concerns: quote-state tracking, unquoted-`>` detection, source-fd classification (stderr skip), and target classification (fd-dup vs file). Raw cyclomatic complexity ≈26; genuine branching ≈12-14 (HIGH). Three specific hazards:

1. **Three cursors with subtle relationships:** `i`, `j` (target scan), `k` (fd-dup digit probe). Asymmetric advancement: stderr branch advances `i` in-place; stdout branch uses `j` cursor but only writes back on fd-dup (line 334) or returns.

2. **Comment-enforced invariant:** Line 94 says "CHECK ORDER (source-fd before target) is load-bearing" — this ordering is guarded only by a comment, not by code structure. A future refactor could reorder the branches and silently break the guard.

3. **Quote representation divergence:** Uses ad-hoc `in_single`/`in_double` bools instead of the shared `QuoteState` enum (see companion rust+consistency findings).

**Fix:** Extract per-`>` classification into a named helper returning `enum RedirectVerdict { Bail, Resume(usize) }`, making the source-before-target ordering *structural* rather than comment-enforced. The loop body becomes cleaner: "scan respecting quotes; at each unquoted `>`, ask the classifier what to do."

**Impact:** Maintainability — existing tests still catch regressions, but decomposition reduces cognitive load and future edit risks.

crates/rskim/src/cmd/rewrite/compound.rs
```diff
@@ -80,11 +83,108 @@ pub(super) fn rewrite_would_corrupt(cmd: &str) -> bool {
...
+    let chars: Vec<char> = cmd.chars().collect();
```
dean0x: **Rust MEDIUM (80%) + Complexity MEDIUM (80%) — two deviations from module patterns**

**1. Vec<char> allocation:** Line 108 collects the command into `Vec<char>`, but CLAUDE.md Design Constraints say "prefer `&str` slices over allocation in the hot path". The sibling quote-scanners (`quoted_redirect_hazard`, `has_unmatched_quotes`) both stream via `cmd.chars()` without collecting. You need backward indexing (chars[i-1], chars[i-2]), so the fix is `cmd.as_bytes()` indexing instead — all matched operators (`>`, `'`, `"`, `\`, `2`, `&`, `-`, space) are single-byte ASCII; multibyte UTF-8 never matches, so behavior is identical and zero-alloc.

**2. QuoteState divergence:** Lines 111-112 use ad-hoc `in_single`/`in_double` bools, but three sibling scanners in this same file use the shared `QuoteState` enum. The illegal state `in_single && in_double` is representable and prevented only by hand-written guards. Fix: Reuse `QuoteState` enum, or add a one-line comment explaining why the bools are intentional.

**Impact:** Constraint alignment (hot-path allocation) and consistency risk (four different quote representations in one file invite divergence).

crates/rskim/src/cmd/rewrite/compound.rs
```diff
@@ -80,11 +83,108 @@ pub(super) fn rewrite_would_corrupt(cmd: &str) -> bool {
...
     }
+    if stdout_redirected_to_file(cmd) {
+        return true;
+    }
     // Whitespace round-trip guard: tokenization must be lossless.
     let rejoined = cmd.split_whitespace().collect::<Vec<_>>().join(" ");
     rejoined != cmd.trim()
 }
 
+/// Return `true` when `cmd` redirects **stdout (fd 1) or both streams** to a
```
dean0x, line 94: **Architecture MEDIUM (80%) — invariant drift risk between two surfaces**

One conceptual invariant ("stdout-to-file must receive raw tool output") is now encoded twice with *different definitions*:

- **Rewrite surface** (`stdout_redirected_to_file` here): syntactic — any unquoted `>` whose target is not an fd-dup. Deliberately bails on `> /dev/null`.
- **Wrapper surface** (`stdout_is_regular_file` in main.rs:717): semantic — `fstat(fd 1).is_file()`. `/dev/null` is a character device, returns false → compresses.

For real regular-file targets, the two AGREE (which is what matters). Divergences (`/dev/null`, FIFOs) are benign today. But there is NO shared doc anchor naming the invariant from both sites, and NO cross-surface test that feeds the same redirect-target taxonomy to both and asserts they agree. This mirrors **PF-002** (dual cache-dir resolvers drifted apart).

**Fix:** Make the shared invariant explicit: Add a doc-comment block to both functions, e.g., `INVARIANT: stdout→file receives raw tool bytes (enforced by stdout_redirected_to_file on rewrite surface AND stdout_is_regular_file on wrapper surface — keep these two in lockstep)`. Stronger: a cross-surface test listing real-file redirect forms, asserting `stdout_redirected_to_file(form)` is true for each.

**Impact:** MEDIUM — each surface has end-to-end tests (catches single-surface regressions); unguarded is a *semantic* divergence on new redirect forms.

crates/rskim/tests/cli_wrapper_argv0.rs
```diff
@@ -322,6 +322,160 @@ mod argv0_dispatch {
...
+        assert_eq!(
+            file_contents, raw_output,
+            "file must contain exactly the raw tool bytes"
+        );
+    }
+
+    /// Control test for D2b: when stdout is a pipe (assert_cmd's default),
+    /// the wrapper STILL compresses — the regular-file guard must not fire.
+    #[test]
+    fn argv0_wrapper_stdout_pipe_still_compresses() {
```
dean0x, line 403: **Testing MEDIUM (85%) — weak control test cannot detect its target regression**

The test `argv0_wrapper_stdout_pipe_still_compresses` claims to guard "regular-file guard must NOT fire on pipe", but both assertions `!stdout.trim().is_empty()` and `stdout.len() <= raw_output.len() + 50` pass whether output is compressed OR raw passthrough. A raw 5-line output (~50 B) satisfies both, so the test cannot distinguish. It duplicates `argv0_ls_wrapper_dispatches_and_does_not_expand` and adds no new protection.

**Fix:** Assert a compression-specific signal:
assert!(stdout.contains("grep "), "pipe stdout must show skim header (compressed, guard did not fire)");

Or for ls, assert the compressed form differs from raw_output exactly, so raw passthrough on pipe would fail the test.

**Impact:** The stated "silent compression regression" guard does not hold (though regression is caught elsewhere via argv0_grep_wrapper_single_header_and_footer).

dean0x:
## Summary of Review Findings

**Inline Comments Posted:** 7 comments on ≥80% confidence findings

**Remaining Findings (60-79% confidence — non-blocking):**

- **Testing (85%)** — `>&-` close-stdout case untested (compound.rs): The `chars[k] == '-'` clause has zero coverage. Recommendation: add `assert!(!rewrite_would_corrupt("cmd >&-"), ">&- close-fd dup")` to `test_corrupt_guard_stdout_to_file_no_bail`.

- **Complexity (80%) / Rust (80%)** — Quote-scanner divergence (compound.rs:111): The function uses ad-hoc `in_single`/`in_double` bools instead of the shared `QuoteState` enum like its siblings. This is addressed in the inline comment at line 108.

- **Reliability (65%)** — SAFETY comment overstates fd-1 validity (main.rs:721). Comment says "fd 1 is always valid" but a process can be started with fd 1 closed. The code is robust because a bad fd → `Err` → `false`, so consider rewording to note the graceful-fallback guarantee.

- **Rust (65%)** — Windows wrapper limitation undocumented (main.rs:727-730). The `#[cfg(not(unix))]` arm returns `false`, so on Windows a PATH-wrapper redirecting stdout to file would still compress (the rewrite/hook surface still protects). Worth a one-line doc note confirming this is an accepted platform limitation.

---

## Summary by Confidence & Severity

### ✅ INLINE COMMENTS (7) — All ≥80% Confidence

1. **Security HIGH (85%)** — `stdout_redirected_to_file` false negatives (compound.rs:107) — **BLOCKING**
2. **Documentation HIGH (90%)** — CLAUDE.md stale after D2b (main.rs:784) — **SHOULD-FIX**
3. **Consistency MEDIUM (85%)** — Footer comment wrong precedent (mod.rs:196) — **SHOULD-FIX**
4. **Complexity HIGH (85%)** — Recommend refactoring for maintainability (compound.rs:107)
5. **Rust MEDIUM (80%) + Complexity MEDIUM (80%)** — Vec<char> allocation + QuoteState divergence (compound.rs:108)
6. **Architecture MEDIUM (80%)** — Invariant drift risk between surfaces (compound.rs:94)
7. **Testing MEDIUM (85%)** — Weak control test (cli_wrapper_argv0.rs:391)

### 📋 SUMMARY COMMENT — Lower Confidence & Context

All ≥80% confidence findings have been posted as inline comments. The 60-79% findings above are noted for reference but do not require action for merge.

---

## Recommendation

All **BLOCKING** items (Security HIGH — false-negative paths) must be addressed. **SHOULD-FIX** items (Documentation, Consistency) should be fixed for maintainability. The remaining findings are optional non-blocking suggestions for future improvement.

**Status:** Ready for author review. No other findings block merge.

Issue #304, opened 2026-06-09
L3: Per-content-type compression of message content blocks (reuse existing engines)


dean0x:
Wave: 2 · Depends on: #300, #301, #302 · Part of: #309

## Summary

This ticket builds the compression engines for skim's Layer-3 message pipeline: given content blocks already classified by #302 as `{code, json, log, text, mixed/code-fence, unknown}`, route each block to the matching deterministic compressor — code through rskim-core's AST transform, logs/tracebacks through the existing `skim log` machinery, JSON through a new structural compressor whose output always parses as valid JSON (the one engine that is a build rather than a reuse: skim's existing JSON skim output is deliberately not valid JSON) — and substitute the result in place with a per-block never-inflate guarantee. Everything else (plain text, unknown) passes through byte-identical. The router is invoked from #303's transform seam on the request path — this ticket also owns flipping the proxy's shipped identity transform to the router, so `skim proxy` actually compresses once this lands — and it operates on **live-zone blocks only** (the not-yet-responded-to trailing turn, per #301's normative zone boundary); prior turns are cache hot zone, out of its reach by construction, and aged-content compaction there belongs to #307 under its typed waiver. This is where the user-visible token savings of the whole Layer-3 effort actually live: the proxy is plumbing, these engines are the product, and skim's deterministic engines already achieve 60–80% reduction on tool output. It is also where skim structurally leads the competition: PRISM shipped an AST code compressor and had to disable it by default after 40% invalid-syntax output on real files, while rskim-core is mature (5,213 tests), deterministic, and never emits invalid syntax. The ticket's job is to bring that proven engine to the in-message content pool that shell hooks cannot reach, without compromising the guarantees that make it trustworthy.

## Why now — evidence

All competitor evidence below was verified in the 2026-06-09 PRISM study.

- **The dominant token pool is in-message content, and nothing reaches it today.** PRISM #327 (35 comments, their most-discussed open issue) documents real sessions seeing only 0–15% savings because file Read/Glob outputs and aged messages go uncompressed — the maintainer concedes the gap. skim's shell hooks cannot reach this pool either: agent Read tools bypass the shell entirely. That pool is exactly what this ticket compresses once the message model exposes it.
[...]

PR #155, opened 2026-04-28
feat(line-numbers): add -n/--line-numbers flag for source line prefixes

dean0x:
## Summary

Adds `-n` / `--line-numbers` flag to skim CLI that prefixes each output line with its source line number. This feature helps users locate specific code sections in the original source file when using skim for code reading and context collection.

## Changes

- **New CLI flag**: `-n` / `--line-numbers` on all skim subcommands
- **Source line mapping**: Accurately tracks original source line numbers across all transformation modes
- **Mode-specific handling**: 
  - Structure mode: Map source lines through removed ranges
  - Signatures mode: Handle line-number gaps from stripped function bodies
  - Pseudo mode: Correct line mapping for modified lines (type annotations stripped)
  - Minimal mode: Preserve line numbers for trimmed comment content
  - Passthrough (full) mode: Identity mapping with last-lines reconciliation

## Bug Fixes Included

- **AC-6**: Pseudo mode line-number gaps for removed body lines (Python)
- **AC-7**: Minimal mode line-number gaps for stripped comment lines (Python)  
- **AC-9**: Correct source line numbers in passthrough + --last-lines (sequential-from-1 bug fixed)
- **AC-11**: Guardrail identity-map validation
- Line-number correctness with duplicate content and --last-lines
- Monotonic matching for accurate line reconciliation in truncation

## Testing

- 7 commits with incremental bug fixes and test coverage
- 8 new unit tests for line-map computation helpers
- 2 CLI regression tests validating prefix output and duplicate-content handling

## Related Issues

Depends on completed implementation of line-number tracking infrastructure.

---

_Created by Claude Code_

PR #385, opened 2026-07-01
fix: agent-reported batch — AST-cap passthrough, search-index disambiguation, grep grouping

dean0x:
## Summary

Batch fix for recently agent-reported skim issues. Of six reports, **three were genuine defects in current `main`** (fixed here) and **three were the stale-wrong-clone-on-PATH hazard** (already fixed on `main`; documented below, no code change).

Each fix lives in the **shared handler/transform layer**, so it applies to **both interception surfaces** (rewrite-hook *and* PATH-wrapper) — they route through the same `process_file` / grep handler / search dispatch, not the dispatch front-end.

### 1. Oversized files degrade to passthrough instead of erroring (`rskim-core`)
**Reported:** `head -40 crates/rskim-search/src/weights.rs` (hook-rewritten to `skim … --mode=pseudo --max-lines 40`) failed with `Error: Too many AST nodes: 100001 (max: 100000). Possible malicious input.`

A legitimate machine-generated file that exceeds an output-volume safety cap (AST node / signature / type-def / markdown-header count) now degrades to a **lossless raw passthrough**, honoring `--max-lines`/`--last-lines` so a `head`-style request still yields a window, not the whole 105K-line file.

- New typed `SkimError::ComplexityLimit`, distinct from `ParseError` (genuine failure) and from the **depth caps**, which stay hard errors (deep nesting signals pathological input, not size).
- The transform dispatcher catches it and falls back to `transform_passthrough_with_line_map`.

### 2. `search index` no longer shadows the search term "index" (`rskim`)
**Reported:** `skim search index --limit 5` errored (`unexpected argument '--limit'`); the literal word "index" was unsearchable.

`index` routes to the build subcommand **only** when trailing args fit the build grammar (`--force`/`--root`/`--max-files`/`--index-dir`). With a query flag or extra positional terms it is a search for "index". Added a POSIX `--` escape (`skim search -- index`). **Backward-compatible:** bare `skim search index` and `skim search index --force` still build.

### 3. grep/rg group-by-file consistently (`rskim`)
**Reported:** the same `grep -n` produced two different shapes (standard `file:line:content` for few matches, grouped for many) — a jarring cross-invocation inconsistency.

The net-savings guard flipped small result sets back to raw based on a token threshold. Set `skip_net_savings_guard` for grep/rg so the **grouped format is the single predictable output at any match volume**. Every matching line is still emitted exactly once; the only cost is tiny greps may render a few bytes larger than raw — a deliberate, accepted relaxation of the never-larger-than-raw guard for grep/rg (not the #317 *content* guarantee). Passthrough cases (`-c`/`-l`/`-L`, unparseable, over the line-bound) are unaffected.

## Not fixed here — already correct on `main` (stale-clone hazard)

`which skim` on the dev machine resolves to a **stale sibling clone** (`/Users/dean/Sandbox/skim`, a `#304` branch predating recent `main` fixes). These reports reproduce only on that binary:

- **git diff leading-hunk `+` lines duplicated** — fixed on `main` by `EmittedCursor` dedup (#350). Current binary: 0 dups; stale: 76 on the same diff.
- **`ls *.processing` "malformed"** — current `ls` handler dispatches and passes through cleanly on no-match.
- **grep `N matches in M files` double-header** — fixed on `main` (#370).

Remediation is operational: rebuild/repoint the deployed binary (or `~/.skim/bin` wrapper) to a current `main` build.

## Validation

- ✅ `cargo check`, `clippy -D warnings`, `cargo fmt --check`, Snyk Code (0 issues) — clean on both changed crates.
- ⚠️ Local test execution was blocked by an unrelated environment issue (a wedged macOS `syspolicyd` hanging freshly-built binaries at startup). Tests are included — core degradation/cap tests, search disambiguation tests, grep group-always e2e tests — and are relied on **CI** here to validate at runtime.

## Test plan
- [ ] CI green (build + full test suite across targets)
- [ ] `skim <large-generated-file> --mode=pseudo --max-lines 40` returns a 40-line window, exit 0
- [ ] `skim search index --limit 5` searches for "index"; `skim search index` still builds
- [ ] small and large `skim grep -n` produce the same grouped shape


dean0x:
# Code Review: Findings Summary

**PR #385** | **Branch**: fix/agent-reported-batch-ast-search-grep → main
**Review Date**: 2026-07-01 | **Status**: APPROVE WITH CONDITIONS

---

## Review Overview

10 specialist reviewers assessed this PR against security, architecture, performance, consistency, documentation, reliability, rust patterns, regression safety, testing, and complexity. The three batch fixes (oversized-file passthrough, search-index disambiguation, grep-group consistency) are well-designed. The merge requires addressing 0 CRITICAL, 0 HIGH hard blockers, and 3 MEDIUM quality-gate findings before merge.

---

## Findings by Confidence Level

### ≥80% Confidence — Inline/Structural Issues

These findings target lines in the PR diff and require fixes before merge:

#### [testing] rg's `skip_net_savings_guard` has no regression test — **90%**  
**Location**: `crates/rskim/src/cmd/file/rg.rs:31`  
**Issue**: This PR flips `skip_net_savings_guard` from `false` to `true` for both grep and rg identically. For grep, an e2e test (`test_grep_small_multifile_groups_consistently`) exercises the guard bypass; for rg, all tests call the renderer directly and never reach `execution.rs:732` where the flag is consumed. Reverting rg's flag would go undetected.  
**Fix**: Add an rg sibling to the grep test (gated on rg availability), or add a dispatch-level test reading `rg::CONFIG.skip_net_savings_guard`.

---

#### [reliability] Degrade-to-passthrough is silent — **80%**  
**Location**: `crates/rskim-core/src/types.rs:276`  
**Issue**: When a structural cap overflows, the dispatcher swallows the `ComplexityLimit` and returns passthrough with no notice—not even under `SKIM_DEBUG=1`. The user who asked for `--mode=pseudo` silently receives full content on the large files where compression matters most. This violates the design constraint "fail loud with actionable messages, never silently."  
**Fix**: Thread a degradation signal to the CLI (enum `Fidelity { Compressed, DegradedPassthrough { ... } }`) and emit `eprintln!` under `SKIM_DEBUG`/`--debug`. Alternatively, document the silence as intentional in CLAUDE.md.

---

#### [rust] SkimError missing `#[non_exhaustive]` — **85%**  
**Location**: `crates/rskim-core/src/types.rs:787` (enum declaration)  
**Issue**: This PR adds `ComplexityLimit` to public `SkimError` (rskim-core is published to crates.io). Every sibling error enum (`rskim-tokens`, `rskim-llm`, `rskim-contract`, `rskim-search`) carries `#[non_exhaustive]` with a doc note. `SkimError` is the lone exception. Adding a variant without the attribute is semver-breaking for downstream exhaustive matchers.  
**Fix**: Add `#[non_exhaustive]` to the enum (with doc note matching siblings). Downstream matchers then need a `_` arm—the intended contract.

---

#### [testing] ComplexityLimit↔ParseError discriminator unpinned — **85%**  
**Location**: `crates/rskim-core/src/types.rs:842`  
**Issue**: The design (stated in the `ComplexityLimit` doc comment) is that count overflows degrade to passthrough while depth overflows stay hard `ParseError`s. The positive side (degrade) is tested; the negative side is not. Nothing asserts that a depth-cap `ParseError` is NOT a complexity limit, so if `is_complexity_limit()` were ever widened, depth-cap failures would silently start degrading without triggering any test failure.  
**Fix**: Add a one-line test:  
```rust
#[test]
fn parse_error_is_not_a_complexity_limit() {
    assert!(!SkimError::ParseError("x".into()).is_complexity_limit());
}
```

---

#### [architecture] Build-grammar duplicated — **85%**  
**Location**: `crates/rskim/src/cmd/search/mod.rs:50`  
**Issue**: `index_args_are_build` hardcodes the build-flag whitelist (`--force`, `--help`, `-h`, `--root`, `--max-files`, `--index-dir`). The authoritative definition lives independently in `IndexCli` (`search/index.rs:102`). They agree today but are only coupled by convention. A future maintainer adding a flag to `IndexCli` and forgetting to mirror it into `index_args_are_build` would silently reintroduce the "index shadowed" bug.  
**Fix**: Add a synchronization test that fails when `IndexCli` gains a long-flag not in `index_args_are_build`'s accept-set. Or derive the grammar via `IndexCli::try_parse_from()`.

---

#### [documentation] Cap-site comments still framed as "SECURITY/malicious" — **85%**  
**Locations**: `crates/rskim-core/src/transform/minimal.rs:83`, `pseudo.rs:408`  
**Issue**: Both sites return `ComplexityLimit` (new logic: "legitimate but large file, degrade to passthrough") but the comment above still says `// SECURITY: Prevent memory exhaustion from excessive nodes`. Siblings (`structure.rs:99`, `signatures.rs:101`, `types.rs:102`) were updated to the new rationale; `structure.rs:531` has no comment at all. The six sites now carry three different comment styles for identical semantics.  
**Fix**: Align minimal.rs:83 and pseudo.rs:408 to the new comment (e.g., `// Node count over the cap: a legitimate but very large file, not an attack. Signal a complexity limit so the dispatcher degrades to lossless passthrough. (#317)`).

---

### 60-79% Confidence — Suggestions (Summary Only, No Inline)

Lower-confidence findings that should be considered:

- **Eager `source.to_string()` in passthrough** (85% [performance]): The degrade path opens with a full-file clone, then discards it in `--max-lines` branches. Pre-existing (traces #155), but newly exposed. Defer allocation to the no-truncation branch.
- **No e2e test for exit-0 + windowed passthrough** (80% [testing]): Add `skim <big> --mode=pseudo --max-lines N` asserting exit 0 + bounded stdout.
- **`--` escape doesn't shield `--help`/`-h`** (88% [rust]): `skim search -- --help` prints help instead of searching for `--help`. Only honor help flags before the first `--`.
- **Serde count-cap asymmetry** (85% HIGH [consistency]): JSON/YAML/TOML key-count caps still hard-error while tree-sitter counts degrade. Asymmetry contradicts documented "consistent protection across all parsing paths" comment.

---

## Pre-existing Issues (Noted, Not Blocking)

- `ParseError` remains overloaded (depth-cap, UTF-8, range, unsupported-transform)
- `Parser::transform` bypasses centralized `ComplexityLimit` degradation  
- Depth-cap untestability (tree-sitter self-limits well below the 500-node cap)

---

## Documentation Gaps (Off-Diff, Should Follow Up)

- **CHANGELOG.md** missing all three user-facing fixes (90% confidence)
- **CLAUDE.md** `search` entry lacks index build-vs-search disambiguation and `--` escape (85% confidence)

---

## Positive Findings

- **ComplexityLimit fallback correctly centralized, not duplicated** across the five transform modes
- **Core purity respected** — no I/O side effects introduced
- **grep/rg symmetry verified** — both tools apply guard consistently  
- **All tests are behavior-focused and deterministic**
- CI (clippy, tests) green

---

## Recommendation

**APPROVE WITH CONDITIONS**

**Conditions (blocking merge):**
1. Add rg regression test (testing, HIGH) — validates `skip_net_savings_guard` for rg
2. Add `is_complexity_limit()` negative-case test (testing, MEDIUM)
3. Update minimal.rs:83 and pseudo.rs:408 comment text (documentation, MEDIUM)
4. Decide on degrade-signal strategy (reliability, MEDIUM) — either emit notice or document silence

**Suggestions (post-merge acceptable):**
- Add sync test for build-grammar drift
- Defer source.to_string() allocation
- Add e2e exit-0 + windowed test
- Fix `--` escape hatch  
- Unify serde cap handling

**Review Summary Files:**  
All detailed reviews saved to `.devflow/docs/reviews/fix-agent-reported-batch-ast-search-grep/2026-07-01_1723/`:
- `security.md` — No security issues
- `architecture.md` — Build-grammar duplication finding
- `performance.md` — Eager allocation noted, net-perf win on grep/rg
- `complexity.md` — Clean complexity posture
- `consistency.md` — Serde asymmetry noted  
- `regression.md` — No regressions in touched code
- `testing.md` — rg regression-test gap and e2e exit-0 gap
- `reliability.md` — Degrade-to-passthrough silence finding
- `rust.md` — #[non_exhaustive] and -- escape findings
- `documentation.md` — Cap comments and CHANGELOG gaps

---

*Review conducted by Claude Code devflow agents. This is an automated comprehensive review for merge readiness.*

PR #421, opened 2026-07-04
fix: agent-reported batch — bash support, diff guardrail, mypy, ls sections, grep indent, hook pinning

dean0x:
## Summary

- **F4 — Bash language support**: Full tree-sitter-bash grammar integration with structure mode, shebang auto-detection (`#!/bin/bash`, `#!/usr/bin/env bash`, `env -S`, `zsh`/`ksh`/`dash`/`fish`), and ADR-002 unknown-file lossless passthrough degrade
- **F1 — git diff**: Hunk-scoped renderer replaces full-node-body walk; output ≤ raw in all cases; ADR-001 guardrail wired in
- **F2 — mypy**: `synthesize_success_line` knob emits "mypy OK 0 issues" on clean exit when injected `--output json` suppressed stdout
- **F3 — ls -R**: Per-directory section rendering; empty dirs visible; pre-header lines (ls -la <file> <dir>/) captured; tier-2 plain-ls sectioning
- **F5 — grep/rg**: `trim_end()` instead of `trim()` — leading whitespace preserved (Python indent, YAML)
- **F6 — hook pinning**: Generated hook scripts embed `SKIM_HOOK_BINARY` (absolute path) + `SKIM_HOOK_COMMIT` (SHA); stale hook detection updated

## Changes

- `Language::Bash` variant + `from_shebang()` + all 6 mode arms + stable lang ID 17
- `process.rs`: shebang detection in `run_transform` + `process_stdin`; `passthrough_with_truncation` for unknown-ext degrade; `ExitCode::from(3)` for `UnsupportedLanguage`
- 4 bash fixtures in `tests/fixtures/bash/`, 18 new tests in `crates/rskim-core/tests/bash_transform.rs`
- `test_cli_unsupported_extension` updated: unknown extension now succeeds with passthrough (ADR-002)
- `test_cli_stdin_without_language_fails` updated: new error message includes "shebang line"; stdin without shebang now degrades to lossless passthrough (exit 0)
- Non-snake-case test names from F7 commit fixed (`test_cat_guard_flag_A_*` → snake_case)
- `cargo fmt` applied to F6/F7 formatting drift in `hooks/mod.rs` and `rewrite/hook.rs`
- Scrutinizer lang_from_id fix: Bash (id 17) reverse mapping for search-index round-trip correctness
- Both-surfaces paired tests: 6 new tests in `cli_both_surfaces_paired.rs` covering F3/F5/F2 across hook-JSON and argv0-dispatch surfaces
- Bash benchmark: LARGE_BASH fixture (~917 lines, real-world deploy orchestration), measured 2.7ms for structure/signatures modes — well under 50ms/1000-line budget

## Breaking Changes

- `skim file.xyz` (unknown extension) now exits 0 with passthrough instead of exit 1 + error (ADR-002). `SKIM_DEBUG=1` emits a notice.
- `skim` exits 3 (not 1) for `UnsupportedLanguage` errors (explicit `--language=unsupported-lang` path).
- `skim <stdin>` without `--language`/`--filename`/shebang now degrades to lossless passthrough (exit 0) instead of error.

## Test Coverage

- **rskim-core**: 497 tests
- **rskim (binary + integration)**: 4099 tests
- **rskim-search**: 738 tests
- **Doctests**: 16 rskim-core + 3 rskim-llm = 19 tests
- **Pair tests (both surfaces)**: 6 new tests (hook + argv0 dispatch)
- **E2E hook regressions (F7)**: 8 tests pinning cat flag/extension guards

Total: **6199 tests**, all passing.

## Reviewer Focus Areas

- `crates/rskim/src/process.rs`: shebang path in `run_transform`, `process_stdin`, and the SKIM_DEBUG notice deduplication (single notice per file)
- `crates/rskim-core/src/types.rs`: `from_shebang()` — CRLF handling, `env -S` parsing, version-suffix stripping
- `crates/rskim-core/tests/bash_transform.rs`: AC-F4.1 zero-function deploy.sh test (line 139)
- `crates/rskim-core/src/transform/diff.rs`: `render_default_scoped` (hunk-scoped renderer), ADR-001 guardrail application
- `crates/rskim/src/cmd/file/mod.rs` + `grep.rs` + `rg.rs`: `trim_end()` leading-whitespace preservation
- `crates/rskim/src/cmd/ls.rs`: Per-directory section rendering + pre-header line capture + tier-2 detection
- `crates/rskim-search/src/ast.rs`: `lang_from_id(17)` reverse mapping for Bash
- `crates/rskim/tests/cli_both_surfaces_paired.rs`: Paired tests across hook and wrapper surfaces
- Exit-code mapping in `main.rs` — ensure only `UnsupportedLanguage` exits 3, not the degrade path

## Notes

- **Post-merge**: After merge, run `./target/release/skim init --agent claude-code` to regenerate the pinned hook script with the new binary path and commit SHA.
- **O1 decision reversal**: The git-diff raw-budget guardrail (ADR-001 application) was initially deferred as a phase-2 item but proved critical for preventing unexpected output expansion in hunk-scoped rendering. Implemented in F1; validates against regressions in the hook output-fidelity layer.


crates/rskim/src/cmd/git/diff/render.rs
```diff
@@ -303,6 +312,104 @@ fn render_changed_only(
...
+fn render_default_scoped(
+    output: &mut String,
+    changed_ranges: &[ChangedNodeRange],
+    hunks: &[DiffHunk<'_>],
+    source_lines: &[&str],
+    ln_width: usize,
+) {
+    // Track which hunks are attributed to at least one changed-node range.
+    // Unattributed hunks are orphans and rendered separately.
+    let mut hunk_attributed = vec![false; hunks.len()];
```
dean0x: **BLOCKING — Silent diff line-drop (#317 under-emission)**

When a hunk straddles a changed AST node AND has lines outside that node's range: the outside lines are clipped by `node-boundary clip`, but then the orphan fallback loop (line 400) checks `if hunk_attributed[hunk_idx]` and skips the entire hunk because it's marked as attributed, even if only partially. This silently drops the outside lines, violating #317 "compress, never truncate" MUST.

**Suggested fix:** Track attribution per-line using `EmittedCursor` instead of per-hunk bool, or explicitly check whether any lines fall outside the clipped range before skipping in the orphan loop.

crates/rskim/src/process.rs
```diff
@@ -327,18 +327,87 @@ fn run_transform(
...
+) -> String {
+    if let Some(n) = max_lines {
+        let lines: Vec<&str> = text.lines().collect();
+        if lines.len() <= n {
+            return text.to_string();
+        }
+        let keep = n.saturating_sub(1);
+        let omitted = lines.len() - n + 1;
+        let marker =
+            format!("# ... ({omitted} lines truncated; use SKIM_PASSTHROUGH=1 to see all)");
```
dean0x, line 411: **SHOULD-FIX — Inconsistent elision markers**

The `max_lines` marker (line 380) includes `"use SKIM_PASSTHROUGH=1 to see all"` hint, but the `last_lines` marker (line 395) only says `"... ({omitted} lines above)"` without the hint. Inconsistent with the shared `output::elision_marker` convention.

**Suggested fix:** Make both markers include the SKIM_PASSTHROUGH hint, or route both through the shared `elision_marker()` function.

CHANGELOG.md
```diff
@@ -7,7 +7,75 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  `skim` was updated or reinstalled** — Generated hook scripts now embed `SKIM_HOOK_BINARY`
+  (the canonicalized absolute install-time path) and `SKIM_HOOK_COMMIT` (short git SHA)
+  at generation time via a `build.rs` compile-time constant. At runtime the hook
+  executes via the pinned binary with a PATH fallback. `hook_is_current()` now checks
+  for the `SKIM_HOOK_BINARY` export — hooks generated before this fix are treated as
+  stale and `skim init` will prompt to reinstall. A version-mismatch helper warns when
+  the running binary differs from the pinned binary SHA.
+
+- **Manually repointing a pinned hook script's binary path triggers an integrity warning**
+  — This is expected behaviour. The hook-script integrity check (`hook_is_current()`)
```
dean0x: **SHOULD-FIX — Misattributes tamper/integrity behavior**

The bullet attributes the "tampered" verdict to `hook_is_current()` (a static currency check for versions/flags), but "tampered" actually comes from `check_hook_integrity()` (SHA-256 mismatch at runtime, line 275 of hook.rs). These are separate checks — `hook_is_current()` detects version skew, `check_hook_integrity()` detects tampering.

**Suggested fix:** Attribute the "tampered" verdict to `check_hook_integrity()` instead.

crates/rskim/src/cmd/file/ls.rs
```diff
@@ -62,6 +66,126 @@ const CONFIG_TREE: ToolRunConfig<'static> = ToolRunConfig {
...
+
+/// Split `ls -la` stdout into per-directory sections.
+///
+/// Returns one `LsSection` per section header found.  Lines before the first
+/// header are captured into a synthetic unlabeled leading section (label `""`),
+/// preserving them so agents see the full listing (ADR-001 — never emptier than
+/// raw).  `ls -la file dir/` places individual file entries before the first
+/// `dir:` header; discarding those lines violated the never-emptier-than-raw
+/// invariant.
+fn parse_ls_long_sections(stdout: &str) -> Vec<LsSection> {
```
dean0x, line 161: **SHOULD-FIX — Duplicated long-line accumulation + divergent section-header predicates**

Entry accumulation logic (skipping `.`/`..`, counting dirs/files) exists in both `try_parse_ls_long_flat()` (lines 352–390) and `parse_ls_long_sections()` (lines 113–180). Additionally, `is_ls_section_header()` (line 81) encodes the section-header check, but individual inline predicates also exist. Future changes must update all locations or they silently drift.

**Suggested fix:** Extract shared `accumulate_long_entry(line, &mut dirs, &mut files, &mut entries)` and ensure `is_ls_section_header()` is the single source of truth for header detection.

crates/rskim/src/cmd/rewrite/hook.rs
```diff
@@ -344,6 +349,63 @@ fn check_hook_version_mismatch(agent: AgentKind) {
...
+///
+/// Compares `SKIM_HOOK_BINARY` (embedded at install time) against
+/// `current_exe()`, and `SKIM_HOOK_COMMIT` against the compiled-in
+/// `SKIM_GIT_COMMIT`.  A mismatch means the hook is running a different binary
+/// than the one that installed the hook (e.g. the user upgraded skim but hasn't
+/// run `skim init` yet, or the binary path moved).
+///
+/// Both warnings are rate-limited per-agent to once per day, logged to
+/// hook.log only — NEVER to stderr.
+fn check_hook_binary_mismatch(agent: AgentKind) {
```
dean0x, line 386: **SHOULD-FIX — Triplicated rate-limited-warn idiom**

The daily rate-limit stamp + warn-once idiom is copy-pasted in at least 3 locations (lines 304, 344, 384, 399 across `check_hook_integrity`, `check_hook_version_mismatch`, and `check_hook_binary_mismatch`). Each uses:
```
let stamp = cache_dir.join(format!(".hook-...-warned-{agent_name}"));
if should_warn_today(&stamp) {
    crate::cmd::hook_log::log_hook_warning(&format!(...));
}
```

**Suggested fix:** Extract `fn warn_once_daily(cache_dir, kind, agent_name, msg)` helper to DRY this pattern.

tests/fixtures/bash/crlf_dialect.sh
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env bash
```
dean0x: **SHOULD-FIX — CRLF fixture has no CRLF bytes**

The file is named `crlf_dialect.sh` but contains pure LF line endings (no `\r\n` bytes). The shebang-line CR test passes via `Language::from_shebang()` unit tests, but the end-to-end full-file CRLF transform path is never exercised. This leaves the CRLF handling in `from_shebang()` untested in integration.

**Suggested fix:** Rewrite this fixture with actual `\r\n` line endings (or add a separate in-code CRLF test to `bash_transform.rs`).

dean0x:
## Summary: Remaining High-Confidence Issues + Lower-Confidence Suggestions

### High-Confidence Issues (Unable to anchor inline - review suggested fixes)

**[process.rs line 283] `--tokens` missing ADR-002 degrade + shebang detection**
- The `Some(budget)` branch doesn't apply shebang detection or passthrough degrade that the `None` branch received
- Result: `skim file.xyz --tokens=500` exits 3 while `skim file.xyz` exits 0 — inconsistent behavior
- Fix: Add shebang fallback (like line 331) + ADR-002 passthrough after the TypeScript default

**[process.rs line 416] process_stdin + bare-bool degrade reason**
- `run_transform()` returns bare `degraded: bool` without indicating the reason
- Forces `process_file()` to re-detect language/shebang 2–3 times
- Fix: Return `enum Degrade` + `Option<Language>` to eliminate re-detection

---

### Lower-Confidence Suggestions (60–79% confidence)

1. **render.rs migration residue** (architecture 75%) — `render_changed_only` is test-only; consider cleanup or document why it's kept

2. **build.rs relative paths** (architecture 60%) — Relative paths in `build.rs` don't resolve in linked worktrees; verify symlink behavior in CI

3. **CHANGELOG exit-code remap** (documentation 70%) — Exit code mapping to `ExitCode::from(3)` for `UnsupportedLanguage` not documented in the exit-code section

4. **PR-description stale paths** (documentation 95% but metadata) — "Reviewer Focus Areas" paths have shifted slightly (e.g. `crates/rskim-core/tests/bash_transform.rs` line references)

5. **process.rs render_changed_only + close_line** (architecture 75%) — `#[allow(dead_code)]` comment on `close_line` suggests incomplete migration residue

### Verification Notes

- **Dependencies review:** Fully clean — no material issues
- **Security review:** No material issues found
- **CONVERGENCE:** Findings 1, 4, 6, 8, 9 have ≥80% confidence and are single-source (no duplicate reviewer reports)

---

### Recommended Actions

1. **BLOCKING:** Fix Finding #1 (render.rs hunk_attributed) — violates #317 compress-never-truncate MUST
2. **High Priority:** Fix Finding #2 and #7 (process.rs consistency) — inconsistent exit behavior and coupling
3. **Medium Priority:** Address Findings #3, 4, 6, 8, 9 — documentation, consistency, and DRY improvements
4. **Low Priority:** Investigate lower-confidence suggestions for next sprint

CHANGELOG.md
```diff
@@ -152,6 +232,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- **Exit codes refined: parse errors → 2, unsupported language → 3** — Previously all
+  `skim` errors that prevented transformation exited 1. The CLI now maps known failure
+  classes to distinct codes: exit 2 for grammar/syntax parse failures
+  (`SkimError::ParseError`), exit 3 for unrecognised language when an explicit
+  `--language` flag was provided (`SkimError::UnsupportedLanguage`). Exit 1 is
+  preserved for all other errors (I/O, config, etc.). Shebang-only or
+  extension-based detection failures degrade to lossless passthrough (exit 0) rather
+  than exiting 3; exit 3 fires only on explicit `--language` values that skim does not
+  support. Scripts that tested for `exit != 0` are unaffected; scripts that branched on
+  the exact exit code may need updating.
```
dean0x, line 259: **SHOULD-FIX — Exit-3 trigger mis-attributed to `--language`** (documentation 90%, corroborated by regression 85%)

This entry says exit 3 fires "when an explicit `--language` flag was provided" and "only on explicit `--language` values that skim does not support." That's inaccurate: `--language` is a clap `value_enum`, so an unknown value is rejected by the parser with **exit 2** and never reaches `SkimError::UnsupportedLanguage`. The real (and only) exit-3 trigger is `process_stdin` returning `UnsupportedLanguage` for a `--filename` hint whose extension is unknown (`crates/rskim/src/process.rs:517-527`).

Fix: replace `--language` with `--filename` in both mentions in this bullet.

—
🤖 Generated with [Claude Code](https://claude.com/claude-code)

crates/rskim/src/main.rs
```diff
@@ -230,7 +238,7 @@ struct Args {
     /// Override language detection (required for stdin unless --filename is given)
```
dean0x: **SHOULD-FIX — Stale arg doc contradicts degrade-to-passthrough** (documentation 85%)

`--language` is no longer "required for stdin unless --filename is given". Per the F4/ADR-002 change, plain stdin with no detectable language now degrades to lossless passthrough (exit 0). The README was corrected for this; this inline `///` doc was missed — the same `#[arg]` block was edited to add `bash` at line 241.

Suggested: "Override language detection (optional for stdin; undetectable input degrades to lossless passthrough)."

—
🤖 Generated with [Claude Code](https://claude.com/claude-code)

crates/rskim/src/process.rs
```diff
@@ -364,23 +499,38 @@ pub(crate) fn process_stdin(
...
+    } else {
+        None
+    };
+
+    let language_or_none = options.explicit_lang.or(filename_lang).or(shebang_lang);
+
+    // ADR-002: when no language is detectable for plain stdin (no --language,
+    // no --filename, no shebang), degrade to a lossless passthrough rather than
+    // erroring — consistent with the file path behaviour in run_transform.
+    // --filename with an unrecognised extension is still an error: the user
```
dean0x: **SHOULD-FIX — Comment overstates when `--filename` errors** (regression 80%)

The comment says a `--filename` with an unrecognised extension "is still an error." But `shebang_lang` is still consulted — it's gated only on `explicit_lang.is_none() && filename_lang.is_none()`, and an unknown extension makes `filename_lang == None`. So `--filename bad.xyz` on content whose first line is a recognised `#!` shebang resolves via the shebang and transforms successfully (exit 0); it is not always an error.

Suggested: qualify to "…is an error **unless the content carries a recognised shebang**, which still wins."

—
🤖 Generated with [Claude Code](https://claude.com/claude-code)

crates/rskim/src/process.rs
```diff
@@ -364,23 +499,38 @@ pub(crate) fn process_stdin(
...
+            return Err(
+                rskim_core::SkimError::UnsupportedLanguage(Path::new(fname).to_path_buf()).into(),
+            );
         }
-    })?;
+        // No --filename, no --language, no shebang — degrade to lossless passthrough.
+        return Ok(stdin_passthrough_result(buffer, &options));
+    }
+
+    let language = language_or_none.expect("language_or_none is Some — checked above");
```
dean0x: **LOW — Reintroduces `.expect()` in app code; prefer `let…else`** (rust 82%)

The `if language_or_none.is_none() { … return … }` guard followed by `.expect("…checked above")` re-derives an invariant the compiler can enforce directly. Per the project Rust rule (no `.unwrap()`/`.expect()` outside tests), collapse the guard and the `.expect()` into one `let…else`:

```rust
let Some(language) = language_or_none else {
    if let Some(fname) = filename_hint {
        return Err(
            rskim_core::SkimError::UnsupportedLanguage(Path::new(fname).to_path_buf()).into(),
        );
    }
    return Ok(stdin_passthrough_result(buffer, &options));
};
```

Correct today — this is a robustness/clarity nit, not a bug.

—
🤖 Generated with [Claude Code](https://claude.com/claude-code)

crates/rskim/src/cmd/file/ls.rs
```diff
@@ -62,6 +66,148 @@ const CONFIG_TREE: ToolRunConfig<'static> = ToolRunConfig {
...
+/// - is non-empty
+/// - does NOT match the permission-line regex (so long-form entries with colon-
+///   suffixed names are not misidentified)
+/// - does NOT start with `"total "` (the disk-usage preamble)
+pub(crate) fn is_ls_section_header(line: &str) -> bool {
+    looks_like_section_header(line.trim(), true)
+}
+
+/// Accumulated data for one directory section in multi-dir `ls -la` output.
+struct LsSection {
```
dean0x, line 100: **SHOULD-FIX — `LsSection` zero-init spelled out at 3 sites** (complexity 85%, corroborated by rust 72%)

`LsSection { label: String::new(), entries: Vec::new()/with_capacity(...), …, files: 0 }` is hand-written at three construction sites (≈163, 179, 375). Adding or reordering a field forces a lockstep update at all three — a maintenance hazard. Derive `Default` on the struct and use `..Default::default()`, overriding only the fields that differ (e.g. `label`, and the `Vec::with_capacity` where pre-sizing matters):

```rust
#[derive(Default)]
struct LsSection { … }
```

—
🤖 Generated with [Claude Code](https://claude.com/claude-code)

crates/rskim/src/cmd/init/install.rs
```diff
@@ -469,17 +469,21 @@ fn find_git_root_from_cwd() -> Option<std::path::PathBuf> {
...
+///
+/// Old scripts that only have a bare `exec skim …` line (pre-F6) are treated as
+/// stale: they lack the pinned-binary exports and must be regenerated.
 fn is_hook_script_current(script_path: &std::path::Path, version: &str) -> bool {
     let Ok(contents) = std::fs::read_to_string(script_path) else {
         return false;
     };
     let version_line = format!("# skim-hook v{version}");
-    let has_bare_cmd = contents
+    let has_pinned = contents
```
dean0x: **SHOULD-FIX — Pinned-binary marker scan duplicated across two files** (architecture 85%)

`contents.lines().any(|l| l.trim_start().starts_with("export SKIM_HOOK_BINARY="))` here is byte-identical to `uses_pinned_binary` in `crates/rskim/src/cmd/init/state.rs:114`. The hook-binary-pinning feature KB itself flags this lockstep as an anti-pattern. Extract a single shared helper (e.g. `script_has_pinned_marker(contents: &str) -> bool`) and call it from both sites, so the marker string/format lives in exactly one place.

—
🤖 Generated with [Claude Code](https://claude.com/claude-code)

crates/rskim-core/tests/bash_transform.rs
```diff
@@ -0,0 +1,333 @@
...
+
+// ============================================================================
+// CRLF full-file transform (AC-F4.2: end-to-end, not just shebang line)
+// ============================================================================
+
+#[test]
+fn test_bash_structure_crlf_full_file() {
+    // AC-F4.2: a file whose EVERY line ends with \r\n (not just the shebang) must
+    // still produce valid structure output.  The fixture crlf_dialect.sh carries the
+    // CRLF name but ships with LF bytes (git normalisation); this in-code test is the
```
dean0x: **SHOULD-FIX — Comment now contradicts the fixture** (testing 85%)

This comment claims `crlf_dialect.sh` "carries the CRLF name but ships with LF bytes (git normalisation)." After the cycle-1 fix (the `.gitattributes` entry + fixture regeneration), `crlf_dialect.sh` now ships **37 real CRLF bytes**. Update the comment to match. Note the fixture is only `.is_ok()`-smoke-tested via `test_bash_structure_fixtures_parse`, so this in-code `CRLF_SH` test remains the real end-to-end CRLF coverage — keep that framing accurate.

—
🤖 Generated with [Claude Code](https://claude.com/claude-code)

crates/rskim/tests/cli_git.rs
```diff
@@ -923,3 +923,246 @@ fn test_skim_git_show_invalid_file_ref_exits_nonzero() {
...
+
+    // Output must be non-empty — diff must be visible.
+    assert!(
+        skim_len > 0,
+        "skim git diff must produce output; got empty stdout"
+    );
+
+    // ADR-001 invariant: skim must NEVER emit more bytes than the raw diff.
+    assert!(
+        skim_len <= raw_len,
```
dean0x, line 1065: **SHOULD-FIX — Size-only assertion is omission-blind (ADR-003)** (testing 80%)

`assert!(skim_len <= raw_len)` here — and in the argv0 sibling `test_skim_git_diff_argv0_surface_does_not_expand` (≈line 1162) — verifies only that output didn't grow. That's exactly the omission-blindness ADR-003 warns against: a renderer that silently drops changed lines still passes. Strengthen both tests to assert the changed body line(s) are actually present in `skim` output (e.g. `assert!(skim_stdout.contains("<a specific changed line>"))`), not just that the byte count shrank.

—
🤖 Generated with [Claude Code](https://claude.com/claude-code)

Cargo.toml
```diff
@@ -30,6 +30,7 @@ tree-sitter-ruby = "0.23"        # ABI 14
 tree-sitter-sequel = "0.3"       # SQL grammar (derekstride/tree-sitter-sql)
 tree-sitter-kotlin-ng = "1.1"    # Kotlin grammar (ABI 14)
 tree-sitter-swift = "0.7"       # Swift grammar (ABI 14)
+tree-sitter-bash = "0.25.1"     # Bash grammar (ABI 15)
```
dean0x: **MEDIUM — Only grammar pinned to a full patch version** (dependencies 92%)

`tree-sitter-bash = "0.25.1"` is the sole grammar carrying a patch component; all 15 siblings use bare-minor (`"0.25"`, `"0.23"`, …). Functionally equivalent today, but it's a consistency wart. Either align to `"0.25"`, or keep the `0.25.1` floor and add a one-line justification comment on this line (e.g. `# 0.25.1 floor: <reason>`). (Whether 0.25.0 is yanked was not verified under the no-network review constraint.)

—
🤖 Generated with [Claude Code](https://claude.com/claude-code)

CHANGELOG.md
```diff
@@ -7,7 +7,78 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
```
dean0x, line 8: **SHOULD-FIX — Breaking changes lack a dedicated heading** (documentation 80%)

The `[Unreleased]` section declares breaking changes (exit-code semantics, `hook_is_current` now requiring the `SKIM_HOOK_BINARY` export, and `--session-id` no longer injected) but scatters them under `### Added` / `### Changed` / `### Fixed`. Add a dedicated `### Breaking Changes` subsection here, matching the `[2.8.0]` precedent (line 385), so downstream users can find them in one place.

—
🤖 Generated with [Claude Code](https://claude.com/claude-code)

dean0x:
## Code Review — Cycle 2 summary

11 reviewers ran (security, architecture, performance, complexity, consistency, regression, testing, reliability, rust, dependencies, documentation). **10 inline comments** were posted for findings at ≥80% confidence (deduplicated across reviewers, and against cycle-1 comments). This comment collects the lower-confidence suggestions, one deferred design note, and the convergence status.

### Convergence status
- **No blocking correctness bugs found this cycle.** All inline items are SHOULD-FIX / LOW / MEDIUM polish.
- **Cycle 1 had 0 false positives** — every cycle-1 finding held up.
- **Re-review of cycle-1 Blocking #1 (`render_default_scoped` per-hunk clip-window dedup):** CONFIRMED CORRECT. The clip-window / `EmittedCursor` interaction drops no changed line and duplicates no changed line. This was verified independently by four reviewers this cycle — **regression, rust, performance, and reliability** — each reaching the same conclusion. The deviation flagged in cycle 1 is resolved.

### Deferred design note (architecture, ~80%)
- **`run_transform`'s bare `degraded: bool` grew in F4.** It now conflates two distinct degrade reasons (unknown-language passthrough vs structural-cap truncation), which forces `process_file` to re-derive the language-resolution chain a second time — including a **double shebang scan** — to recover the reason the bool discards. This is the cycle-1 deferred "Degrade enum" item; F4 grew the same wart rather than shrinking it. Recommend a `DegradeReason` enum (or `Option<DegradeReason>`) so the reason travels with the result and the language chain is resolved once. Non-blocking; tracked as design debt.

### Sub-80% suggestions (informational — not posted inline)
- **`render.rs:373` — prefer `saturating_add`** (reliability, 65%): arithmetic that is currently bounded by upstream invariants could use `saturating_add` for defense-in-depth. Correct today.
- **`bash_transform.rs:315` — tautological escape-hatch assertion** (testing, 75%): the test's escape-hatch branch asserts a condition that is trivially true; it does not meaningfully constrain behavior. Consider asserting the concrete expected output instead.
- **Per-agent `SKIM_HOOK_COMMIT` test coverage** (consistency, 68%): the `SKIM_HOOK_COMMIT` embedding is exercised for some agents but not uniformly across all six hook protocols; consider a parameterized test so every agent's generated script is checked.
- **`env!` vs `option_env!` for build-time SHA** (rust, 50%): stylistic preference around how the compile-time git SHA is read; current form is acceptable.
- **`render.rs` migration residue** (`#[cfg(test)]` / `#[allow(dead_code)]`): prior-known leftover scaffolding; noted for eventual cleanup, not introduced by this PR.

—
🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #426, opened 2026-07-08
fix: preserve tabs through gh/diff wrappers; document command wrapping in agent guidance

dean0x:
## Summary

- `skim gh pr checks` silently fell through to passthrough: `strip_ansi_escapes` dropped `\t` before `RE_GH_CHECK_TAB` could match, and gh's exit 8 (pending/failing checks) was classified as `UnexpectedFailure` and raw-forwarded before parsing.
- `skim diff` fused the file path with the mtime timestamp: `diff -u` emits `--- path\t<mtime>` headers, and the tab was dropped before `try_parse_standalone_unified` could split on it.
- Agent guidance had no explanation for command wrapping (`skim ls`, `skim gh`, …), leaving agents confused about "unexpected" `skim <tool>` invocations.

## Changes

**`gh` CONFIG** (`crates/rskim/src/cmd/infra/gh/mod.rs`)
- `skip_ansi_strip: false → true`: gh emits no ANSI when piped; tabs survive.
- `expected_exit_codes: &[] → &[8]`: gh exits 8 for pending/failing checks — a parseable result, not an error. Exit 8 is still propagated to the caller.

**`diff` CONFIG** (`crates/rskim/src/cmd/file/diff.rs`)
- `skip_ansi_strip: false → true`: diff emits no ANSI; the `\t` in `--- path\t<mtime>` headers now survives to the parser. ADR-001 net-savings guard undisturbed.

**Guidance template** (`crates/rskim/src/cmd/init/helpers.rs`)
- Added "### Command wrapping" section (user-reviewed wording) explaining that the rewrite hook may run `ls`, `grep`, `gh`, etc. as `skim <tool>`, and instructing agents to flag garbled compressed output to the user rather than working around it silently.

**Tests**
- Config-lock unit tests in `gh/mod.rs` and `file/diff.rs` pin the new CONFIG values so any regression fails immediately.
- Positive guidance asserts added to `test_guidance_content_has_version_markers`; all prior negative asserts (`!SKIM_PASSTHROUGH`, `!rskim`) are preserved.
- New `tests/cli_e2e_tab_preservation.rs` with three stub-driven e2e tests that drive the real strip→parse pipeline (per PF-004):
  - `test_gh_pr_checks_exit0_compressed_summary` — tabs survive, "5 checks" summary appears
  - `test_gh_pr_checks_exit8_summary_and_exit_code` — exit 8 compresses and propagates, URL in failing item (AD-INFRA-15)
  - `test_diff_tab_header_path_not_glued` — path not fused to mtime, exit 1 propagates

## Breaking Changes

None.

## Blast radius

- **gh**: `CONFIG` is shared by all `run_tool` gh routes (`pr checks`, `pr view`, `issue view`, `run view`, `release view`, `list`, `api`). `gh run watch` is unaffected (streaming path bypasses `run_tool`). A hypothetical gh exit-8 from a non-checks route falls to Tier-3 passthrough with exit code preserved — same observable behavior as before this fix.
- **diff**: `CONFIG` applies only to `skim diff`; `skim git diff` uses a separate config.
- **Guidance rollout**: version-gated per ADR-004 — existing installs pick up the new section on the next version bump + re-pin. Binary-side gh/diff fixes are effective immediately after rebuild.

## Reviewer Focus Areas

- Verify `expected_exit_codes: &[8]` blast radius note above is complete.
- Confirm the "Command wrapping" guidance wording does not mention `SKIM_PASSTHROUGH` (per existing negative assert) or `rskim`.
- Check that `guidance_content_mdc` wraps the same body and therefore picks up the new section automatically (no separate edit needed — verified in `test_guidance_content_mdc_has_frontmatter`).


dean0x:
## Code Review Findings — PR #426

High-confidence findings (≥80%) from specialized reviewers:

### 🔒 Security — ANSI Injection in Re-emitted Output (80%)

**Files:** `crates/rskim/src/cmd/infra/gh/mod.rs:75` + `crates/rskim/src/cmd/file/diff.rs:37`

When `skip_ansi_strip: true`, terminal escape sequences now survive into re-emitted summary fields:
- **gh**: check names from `build_checks_result()` 
- **diff**: header paths from `build_file_result()`

GitHub CI job names and third-party status apps are attacker-influenceable.

**Recommended fix** (preserves tab-preservation):
```rust
name: crate::output::strip_ansi(caps[1].trim()),
path: crate::output::strip_ansi(&path),
```

---

### ✅ Testing — Config-Lock Test Name/Assertion Mismatch (88%)

**File:** `crates/rskim/src/cmd/infra/gh/mod.rs:305`

Test name promises membership but asserts exact equality. If a future change adds another valid exit code, the test fails with misleading message.

**Fix:**
```rust
assert!(CONFIG.expected_exit_codes.contains(&8),
        "gh CONFIG.expected_exit_codes must include 8");
```

---

### ⚙️ Testing — Fixture-Coupled Regression (85%)

**File:** `crates/rskim/tests/cli_e2e_tab_preservation.rs:189`

Regression detection depends on fixture mtime year literal `2026`. Editing the fixture year disarms the test.

**Fix:** Derive forbidden patterns from fixture at runtime.

---

### 📝 Consistency — Volatile Line-Number Reference (85%)

**File:** `crates/rskim/src/cmd/file/diff.rs:247`

Comment uses absolute line numbers instead of symbol references. Goes stale on next edit.

**Fix:** Reference `try_parse_standalone_unified` by name.

---

**Summary:**
- Created: 0 inline comments (API constraints)
- Lower-confidence items (60-79%): See below

**Other Findings (60-79% confidence)**
- Testing (82%): gh-wide `skip_ansi_strip` but only `pr checks` tested
- Performance (65%): Stale baseline comment in execution.rs:744
- Documentation (65%): README drops `cat`/`head`/`tail` examples

---

*Claude Code Review | PR #426*

PR #446, opened 2026-07-12
fix: agent-reported batch — fileops -h passthrough, pseudo return-type preservation, truthful init summary

dean0x:
## Summary
Three validated bugs from the agent-reported issue batch (validated against v2.11.0; the other 9 claims were already fixed, intended behavior, or not reproducible).

1. **fileops dispatcher swallowed tool-level `-h`** (serious): `skim grep -h <pattern> <files>` printed skim's fileops help and exited 0 (masking failure) instead of running grep with --no-filename. Guard in `cmd/file/mod.rs` narrowed to `--help` only (mirrors the `db/mod.rs` hostname precedent) — affects all 12 fileops tools where `-h` has tool semantics (`grep`, `ls -lh`, `du -h`, `df -h`, `tree -h`, …). rg's rewrite rule now skips `-h`/`--help` (for rg those ARE help flags); `grep -h` intentionally still rewrites (safe post-fix, exercises rewrite→dispatch→parser).
2. **pseudo mode stripped function return types in Rust/Python/TypeScript** (design): the `cat` rewrite maps code files to `--mode=pseudo`, so agents reading source via `cat` silently lost function contracts — only 3 of 15 languages stripped them (accident of the original implementation). Return types are now preserved as API surface (A4 contract, mirroring c244a12/#367 visibility-modifier fix): Rust special-case deleted; Python/TS get a position-aware guard (`return_type` field child, preserved wholesale so nested type args survive). Parameter annotations/generics declarations remain stripped for Python/TS.
3. **`skim init` summary/dry-run printed wrong actions** (cosmetic): claimed it would patch settings.json for Copilot (which actually writes `<config>/hooks/skim.json` and never touches settings.json) and hardcoded "PreToolUse" for all agents (wrong for Gemini `BeforeTool`, Cursor `preToolUse`). Output is now protocol-driven via `uses_dedicated_hook_file()` + `hook_event_key()` with explicit `AgentKind` DI.

## New in This Revision

- **(a) Stderr transparency marker for hook-rewritten reads:** when the PreToolUse hook rewrites `cat`/`head`/`tail <code-file>` into `skim <file> --mode=pseudo`, the rewrite engine now injects `SKIM_REWRITTEN_FROM=<origin>` into the command token stream (ADR-004 out-of-band channel). On execution, when that env var is set and the served view differs from raw file bytes, skim emits a one-line stderr notice: `[skim] transformed view (cat → skim --mode=pseudo): not raw file bytes — SKIM_PASSTHROUGH=1 for raw output`. Multi-file aggregate path emits a single "N/M files not raw bytes" summary instead of one notice per file. Cache-hit path also emits the notice when applicable. Silent when output is byte-identical to raw, when the env var is absent, or when the fidelity guardrail forces raw output.
- **(b) Corrected agent guidance and README wording:** hook-guidance text now explicitly names `cat`/`head`/`tail` file-read rewrites and removes the false "does not change what the command did" claim. README pseudo-mode summary rows updated to reflect return-type preservation (60d0659).
- **(c) Review-condition fixes:** rewrite-engine regression test added for `grep -h` flag preservation (`test_grep_h_still_rewrites_and_preserves_flag` in `rewrite/rules.rs`); `cli_init.rs` dry-run assertion strengthened from vacuous `if let Some(...) { assert!(...) }` to unconditional `expect()+assert`; misleading "rewrite rule rewrites this" comment in `cli_e2e_file_h_flag.rs:79` corrected to name the handler surface (PF-004 two-surface distinction). All three fixed in 2574a6c.
- **(d) New E2E suite `cli_transparency.rs` (8 cases):** covers tagged-pseudo single-file, multi-file aggregate, cache-hit, mode-name, guardrail-suppressed, head-tag, and byte-identical (no-marker) paths.

## Verification
- `cargo nextest run -p rskim --all-targets -j 4`: 4346 passed · `-p rskim-core` 507 passed + 16 doctests
- Full `cargo test --all-features` regression: exit 0, all green
- `cargo clippy -p rskim -p rskim-core --all-targets -- -D warnings` + `--all-features` clippy (CI parity): clean · `cargo fmt --check`: clean
- Independent QA: 16/16 acceptance scenarios (both dispatch surfaces; sandboxed init dry-runs for copilot/gemini/cursor/claude — exit 0, zero files written)
- ~24 new/inverted tests: `cli_e2e_file_h_flag.rs` (4 tests), `cli_transparency.rs` (8 tests, added this revision), 8 return-type preservation tests, Copilot/Gemini dry-run assertions, 1 rewrite-engine `grep -h` coverage test (added this revision)

Note: local Snyk SAST scan pending (authentication unavailable in this session); the PR-level Snyk check will run in CI — a red `security/snyk` result with "limit of private tests" is the known pre-existing account-quota issue, not a code failure.

## Commits
- 167a331 fix: fileops dispatcher must not intercept tool-level -h as help
- 4613f59 fix: pseudo mode preserves function return types as API surface (A4)
- 0ae96f1 fix: skim init summary/dry-run output protocol-driven for all agents
- ec25d80 docs: modes.md pseudo return-type preservation
- b81f1f8 refactor: simplify (comment residue, matches!, let-chains)
- f208150 feat(transparency): origin tag in rewrite engine + marker emission
- 447370e fix(init): guidance wording for hook-rewritten file reads
- 2574a6c test: PR #446 review-condition fixes (grep-h, conditional, two-surface)
- 60d0659 docs: README pseudo mode rows + CHANGELOG
- e3e53ca refactor: simplify env-pair parsing and redundant test assertion
- b3ddd12 test: guardrail-no-marker and head-tag max-lines fixes (cli_transparency)
- 112bd0d docs(knowledge): update hook-binary-pinning feature knowledge base


dean0x:
**Misleading test comment conflates the two dispatch surfaces**

The comment claims "The rewrite rule rewrites this to `skim grep -hn …`" and the dispatcher must NOT intercept `-h`, but this test invokes `skim grep -hn …` **directly** via the handler surface, which never calls `try_rewrite`. The rewrite engine is only reached via `skim rewrite`/`--hook`.

The test is correct and valuable — it exercises the dispatcher surface — but the comment is misleading. Future maintainers might conclude the grep `-h` rewrite path is covered when it isn't.

**Suggested fix:** Reword to state the test verifies the **handler** surface only (`skim grep -h` dispatch), and drop the "rewrite rule rewrites this" claim. Or add a separate rewrite-engine assertion (see the should-fix item in the report for the missing test).

**Context:** This avoids the PF-004 anti-pattern of treating both dispatch surfaces as equivalent for rewrite tests.

dean0x:
**Vacuous conditional assertion in test**

At `crates/rskim/tests/cli_init.rs:1494`, the "must NOT say PreToolUse" check is guarded by `if let Some(line) = patch_line { assert!(...) }`. If the patch/summary line is ever absent (e.g., a future refactor drops it), the negative assertion is silently skipped rather than failing.

The unconditional `stdout.contains("BeforeTool")` above carries most of the intent, so impact is low, but the anti-regression guarantee for the specific `PreToolUse` wording is weakened.

**Suggested fix:** Assert the patch line exists **and** lacks `PreToolUse`:
```rust
let line = patch_line.expect("dry-run must emit a patch/summary line");
assert!(!line.contains("PreToolUse"), ...);
```

dean0x:
**Long function with 8 parameters exceeds readability threshold**

The `print_dry_run_actions` function at line 1276 now carries 8 parameters (`state, no_guidance, global, env, permissions, tier, perm_dir, agent`), triggering the newly added `#[allow(clippy::too_many_arguments)]`. The body is ~101 lines, with a trailing permissions-enumeration block (~55 lines) that is a distinct responsibility.

**Context:** The PR's let-chain refactor actually reduced nesting by one level and removed duplication, so the net complexity improvement is real. The one net-new complexity signal is the 8th parameter. The explicit `agent` DI is documented-intentional and reasonable (removes fallible internal derivations), but it pushes the signature past the readable limit.

**Suggested fix (optional, non-blocking):** Extract the permissions-enumeration block (lines 1317–1373) into a helper, e.g. `print_dry_run_permissions(agent, permissions, tier, perm_dir)`. This drops `print_dry_run_actions` to ~45 lines, removes the need for the `#[allow]`, and isolates the deepest-nested logic.

dean0x:
**Missing rewrite-engine test: `grep -h` still rewrites and preserves `-h`**

At `crates/rskim/src/cmd/rewrite/rules.rs:2382`, there is a test for `rg -h` being skipped by the rewrite rule (correct), but there is **no** test verifying that `grep -h` **still** rewrites and preserves the flag.

**The problem:** Fix 1 deliberately makes `rg` skip `-h`/`--help` (tested by `test_rg_help_skip`) while `grep -h` must **still** rewrite. The grep half is not guarded by a rewrite-engine test. Existing grep rewrite tests use `-i`, `-r`, `-rn` — none assert `grep -h`. The E2E tests in `cli_e2e_file_h_flag.rs` exercise the **handler** surface (direct `skim grep`), which never touches `try_rewrite`. So if someone symmetrically "fixed" grep by adding `-h` to its skip list, `grep -h` would silently stop being rewritten/compressed and **no test would fail**.

**This avoids the PF-004 anti-pattern:** Distinguishing between the two dispatch surfaces (rewrite-engine vs. handler) in test coverage.

**Suggested fix:** Add a rewrite-engine assertion mirroring `test_grep_r_preserves_flag`:
```rust
#[test]
fn test_grep_h_still_rewrites_and_preserves_flag() {
    use crate::cmd::rewrite::engine::try_rewrite;
    let r = try_rewrite(&["grep", "-h", "pat", "f1", "f2"]).expect("grep -h must rewrite");
    assert!(r.tokens.contains(&"skim".to_string()));
    assert!(r.tokens.contains(&"-h".to_string()), "-h must survive; got: {:?}", r.tokens);
}
```

dean0x:
## Code Review Summary — PR #446

**4 inline comments created above** for high-confidence (≥80%) findings.

These are valuable but non-blocking. Consider addressing in follow-up work or as future maintenance:

**Testing (72%)** — Variable/property annotation stripping not directly tested (`crates/rskim-core/src/transform/pseudo.rs:11`)
- The module doc claims "Parameter, variable, and property type annotations are still stripped," guarded by `is_return_field_child`, but only parameter stripping is tested.
- No test covers a Python variable annotation (`x: int = 5`), TS class property (`count: number = 0`), or TS class-method return type (distinct from interface `method_signature`).
- Impact: Low. A one-line negative test would harden the doc's claim.

**Testing (72%)** — Copilot dry-run test lacks positive "Register hook" assertion (`crates/rskim/tests/cli_init_copilot_migrate.rs:149`)
- The test asserts absence of "Patch settings"/"Would patch" and presence of `skim.json`, but never positively asserts "Register hook … (BeforeTool entry)" wording.
- A regression that dropped the register line but left `skim.json` elsewhere could pass unnoticed.

**Testing (68%)** — Weak positive signal in `test_grep_h_flag_not_intercepted_as_help` (`crates/rskim/tests/cli_e2e_file_h_flag.rs:24`)
- Only asserts *absence* of help text; `grep -h pat /dev/null` yields empty stdout, so the test can't distinguish "reached grep" from "errored before grep."
- The multifile test compensates, but adding a positive signal would strengthen it.

**Complexity (65%)** — Magic offset in `adjust_type_start` (`crates/rskim-core/src/transform/pseudo.rs:526`)
- `start.saturating_sub(2)` hard-codes the byte length of the longest separator (`": "`); coupling stays in sync with `SEPARATORS` only by hand.
- Deriving the window from `SEPARATORS.iter().map(|s| s.len()).max()` would make the coupling explicit. Minor; previous code had the same shape.

**Documentation (72%)** — CHANGELOG + `pseudo.rs` module-doc over-generalize Rust (`CHANGELOG.md:56`, `crates/rskim-core/src/transform/pseudo.rs:13`)
- Blanket "parameter … annotations are still stripped" holds only for Python/TypeScript. Rust pseudo mode does **not** strip parameter type annotations (its `strip_kinds` contains no `"type"` node).
- The authoritative `docs/modes.md` per-language table is correct; only the CHANGELOG summary and module doc over-generalize.
- Consider: "for Python/TypeScript, parameter … annotations are still stripped."

**Documentation (68%)** — `docs/modes.md` lists "generics" as removed while return types preserve nested generics (`docs/modes.md:11`, `docs/modes.md:275`)
- The comparison table and "What's Removed" list generics/type-params as stripped, yet return-type examples preserve them (`-> Result<T, E>`, `Promise<User>`, `tuple[int, str]`).
- Clarified in surrounding prose, but summary line reads slightly contradictory. A parenthetical "(except inside preserved return types)" would remove the tension.

---

### Non-Diff Finding (Out of Scope for Inline Comments)

**Documentation (95% confidence) — README.md pseudo-mode description is stale** (`README.md:236`, `README.md:243`)
- This PR changed pseudo mode to preserve function return types (updated `docs/modes.md`), but the parallel description in README was not updated.
- `README.md:243`: `"strips types"` is now misleading — pseudo strips *parameter* types but *preserves* return types.
- `README.md:236`: "What's Kept" reads `Logic flow, names, values` instead of updated `Logic flow, names, values, visibility, return types`.
- **Not commentable inline** (README.md is not in this PR's diff).
- **Suggested fix:**
  - Line 243: `# Pseudocode (strips param types & decorators; preserves visibility & return types)`
  - Line 236 "What's Kept": `Logic flow, names, values, visibility, return types`

---

### Summary

| Category | Count | Status |
|----------|-------|--------|
| Blocking (≥80% confidence) | 4 | Inline comments created |
| Should-Fix (≥80% confidence, no test gap) | 1 | Covered by inline comments |
| Lower-Confidence Suggestions (60-79%) | 5 | Summary only |
| Pre-existing / Non-blocking | 0 | — |
| Non-diff findings (≥80%) | 1 | Summary only (README not in PR diff) |

**Overall Assessment:** ✅ APPROVED_WITH_CONDITIONS
- No correctness defects; tests are behavior-focused and deterministic.
- Address the two testing gaps (dispatch-surface clarity + `grep -h` rewrite coverage) and the README documentation drift.
- Optional: extract the permissions-enumeration helper and the magic offset suggestion.

---

**Reviewer:** Claude Code (v0.1) — Code Review Agent  
**Reports reviewed:** architecture, complexity, consistency, documentation, performance, regression, reliability, rust, security, testing  
**Confidence threshold:** ≥80% for inline comments

dean0x:
Fixed in **2574a6c** — the doc comment at line 79 now explicitly names the **handler surface** (direct `skim grep` dispatch) and drops the `try_rewrite` / rewrite-engine claim. The companion rewrite-engine assertion (`test_grep_h_still_rewrites_and_preserves_flag` in `rewrite/rules.rs`) was added in the same commit to ensure both surfaces are independently covered (PF-004 two-surface distinction).

dean0x:
Fixed in **2574a6c** — the conditional `if let Some(line) = patch_line { assert!(...) }` was replaced with:
```rust
let line = patch_line.expect("dry-run must emit a patch/summary line");
assert!(!line.contains("PreToolUse"), ...);
```
The `expect` forces a test failure if the patch line is absent, closing the silent-skip gap.

dean0x:
Declined by explicit user decision — the `print_dry_run_permissions` helper extraction was considered and rejected. The `#[allow(clippy::too_many_arguments)]` annotation is retained; the 8th parameter (`agent: AgentKind`) is documented-intentional DI that removes fallible internal derivations, and the function body is within acceptable bounds given the call-site context. The suggestion is appreciated and noted as a future maintenance option.

dean0x:
Added in **2574a6c** — `test_grep_h_still_rewrites_and_preserves_flag` in `crates/rskim/src/cmd/rewrite/rules.rs` asserts that `try_rewrite(["grep", "-h", "pat", "f1", "f2"])` returns a rewritten command that contains both `skim` and `-h`, preventing silent regression if `grep -h` is ever symmetrically added to the skip list. This is the rewrite-engine coverage that complements the handler-surface tests in `cli_e2e_file_h_flag.rs`.

dean0x:
Thanks for the thorough review. Here is the resolution status for each finding:

**High-confidence (inline) findings — all resolved:**
- **Misleading test comment (cli_e2e_file_h_flag.rs:79):** doc comment reworded to name the handler surface and drop the `try_rewrite` claim — 2574a6c.
- **Vacuous conditional assertion (cli_init.rs:~1489):** replaced `if let Some` guard with `expect()+assert` — 2574a6c.
- **Missing rewrite-engine grep `-h` test:** `test_grep_h_still_rewrites_and_preserves_flag` added to `rewrite/rules.rs` — 2574a6c.
- **Stale README pseudo-mode rows (README.md:236/243):** both rows updated for return-type preservation — 60d0659.

**Optional refactor (print_dry_run_actions 8 params):** declined by explicit user decision; `#[allow(clippy::too_many_arguments)]` retained. The `agent: AgentKind` DI is intentional and removing it would reintroduce fallible internal derivation.

**Non-diff README finding:** covered by 60d0659 (README pseudo-mode rows updated as part of this revision).

All four blocking findings are addressed. The lower-confidence suggestions (annotation-stripping test coverage, Copilot positive assertion, magic offset, CHANGELOG/module-doc Rust over-generalization) are noted for follow-up.

Issue #477, opened 2026-08-09
skim init cannot re-point an existing hook at a different binary — hook_binary_pin is read but never validated


dean0x:
Same shape as #466 (commit field not compared), which PR #470 fixed. The pin field is the remaining uncovered case.

## What happens

`skim doctor` parses `SKIM_HOOK_BINARY_PIN` out of every installed hook script and displays it under "pin:". The value is informational only: no component in `skim init` reads it back and compares it against the binary that `init` would write.

Consequence: when two clones exist on the same machine (the exact hazard `CLAUDE.md` warns about — "a bare `skim` on `$PATH` may resolve to a *different* local clone"), `skim init --yes` on the second clone reports "Already up to date. Nothing to do." and exits without touching the hook. The hook continues to point at the first clone's binary.

## Observed

After `skim init` had written a hook pinned to `~/.cargo/bin/skim`, the binary at that path was replaced by a different build (a feature-branch clone). Running `skim init --yes` in the first clone's directory:

```
Already up to date. Nothing to do.
```

`skim doctor` reported `✓ installed … pin: ~/.cargo/bin/skim` (correct path, wrong binary). The hook-version and commit fields both matched because the two builds came from the same commit — the only difference was which filesystem path the binary lived at.

The "already up to date" fast path at `crates/rskim/src/cmd/init/install.rs` line 474 checks `hook_is_current()` (version + commit) and `guidance_current`, but has no corresponding `pin_is_current()` predicate.

## Why this matters

CLAUDE.md explicitly documents the parallel-clone hazard:

> "A bare `skim` on `$PATH` may resolve to a *different* local clone (this machine keeps parallel clones to avoid worktree churn), so it can silently exercise the wrong code."

The binary pin exists precisely to detect and correct this. Displaying it in `doctor` without acting on it in `init` gives a false sense of safety.

## Relationship to prior work

#466 identified that `SKIM_HOOK_COMMIT` was captured but not compared; #470 added the commit comparison gate. `SKIM_HOOK_BINARY_PIN` is the same pattern: stored in the hook script, surfaced by `doctor`, but absent from the `init` validation loop.

This is also tracked as the next instance of **PF-015** (display-without-gate) in the project's learning ledger.

## Suggested fix

[...]

Issue #471, opened 2026-08-09
skim doctor reports hooks healthy while integrity/tamper detection is firing — it never checks the .sha256 manifest


dean0x:
Found immediately after \`skim doctor\` shipped (#468), while repairing hand-edited hook scripts on a dev machine.

\`skim init\` writes a SHA-256 manifest alongside each generated hook script at \`{config_dir}/hooks/skim-{agent}.sha256\`. \`verify_script_integrity()\` in \`crates/rskim/src/cmd/integrity.rs\` compares that manifest against the live file, and \`check_hook_integrity()\` (\`crates/rskim/src/cmd/rewrite/hook.rs\`) logs \`hook script tampered: …\` when they diverge.

**\`skim doctor\` does not consult the manifest at all.** Its hook section parses only \`SKIM_HOOK_VERSION\` and \`SKIM_HOOK_COMMIT\` out of the script text. So a script whose text *looks* right but whose hash no longer matches its manifest is reported as healthy.

**Observed.** Four agent hook scripts were edited in place with \`sed\` (changing only \`SKIM_HOOK_COMMIT\`). All four manifests went stale:

| Agent | Manifest hash | Actual script hash |
|---|---|---|
| claude-code | \`de45448c…\` | \`5c65c1c0…\` |
| codex | \`d09d5b94…\` | \`f378dc05…\` |
| gemini | \`f76991fe…\` | \`61fa9d55…\` |
| cursor | \`d222a1e8…\` | \`9296f961…\` |

\`skim doctor\` reported all four as \`✓ installed … pin: …\` with no warning, and exited on staleness alone. Meanwhile \`hook.log\` carried \`hook script tampered: /Users/dean/.claude/hooks/skim-rewrite.sh (run \`skim init --yes\` to reinstall)\`, and — more seriously — \`check_hook_integrity()\` returning true causes the drift-detection block to \`return vec![]\`, so **integrity failure silently suppresses all drift detection** for that agent.

**Why this matters.** \`doctor\` exists to be the authoritative answer to "is my skim installation sane?" A tampered or corrupted hook script is precisely the condition it should surface, and it is the one condition that also disables the other detection path. Reporting \`✓\` there is a false clean bill of health.

[...]

Issue #466, opened 2026-08-07
skim init --yes refuses to re-pin SKIM_HOOK_COMMIT after an in-place rebuild (and the drift warning tells you to run it)


dean0x:
Found during QA of PR #456.

**Reproduction (exact, observed):**
Hook at `~/.claude/hooks/skim-rewrite.sh` pins `SKIM_HOOK_COMMIT=ca6e756`. Binary rebuilt in place; `skim --version` now reports `2.11.0 (d5695c1)`. Then:

```
$ ./target/release/skim init --yes --agent claude-code
  skim init -- Claude Code integration setup

  Checking current state...
  + skim binary: /Users/dean/Sandbox/skim-issues/target/release/skim (v2.11.0)
  + Config: /Users/dean/.claude/settings.json (exists)
  + Hook: installed (v2.11.0)

  Already up to date. Nothing to do.
```

The hook file is left **byte-identical** (md5 `f491aa011c55e0d2e02c26be471f12ba` before and after), still recording `ca6e756`.

**Root cause** — `crates/rskim/src/cmd/init/state.rs:39`:

```rust
pub(super) fn hook_is_current(&self) -> bool {
    self.hook_version.as_deref() == Some(&self.skim_version) && self.hook_uses_pinned_binary
}
```

It compares `hook_version` against `skim_version` (`2.11.0 == 2.11.0`) and checks the pinned-binary flag. It never compares the recorded `SKIM_HOOK_COMMIT` against the compiled-in `SKIM_GIT_COMMIT`. An in-place rebuild at the same semver is therefore structurally undetectable by `init` — which is the normal state of affairs during development, where the version only changes at release time.

**Why this is more than cosmetic.** `crates/rskim/src/cmd/rewrite/hook.rs:428-443` detects exactly this drift and emits:

```
binary rebuilt in-place: hook commit {hook_commit}, running commit {compiled_commit} (run `skim init --yes` to update)
```

The remediation it names is precisely the command that refuses to act. `--force` is uninstall-only (confirmed via `skim init --help`), so there is currently **no first-class way to re-pin**. A user following skim's own advice loops.

**Scope limit (keeping this accurate):** the hook execs `SKIM_HOOK_BINARY` by absolute path, so it does run the newly built binary. Only the recorded provenance string is wrong. No output is mis-compressed as a result — but the commit field is exactly the field a drift check relies on, so a stale value defeats provenance reporting.

[...]

PR #488, opened 2026-08-17
fix: hook provenance, pipe fidelity, and transform linearization (#478, #479, #487)

### LINKED ISSUE #476: pseudo/minimal modes drop top-level header comments (license, SPDX, provenance lines)
Found while tracing why `--mode=pseudo` stripped a `# skim-hook vX.Y.Z` provenance line from a hook script, and why the economics of that transform were inverted.

## Root cause

`is_doc_comment()` in `crates/rskim-core/src/transform/minimal.rs` returns `false` unconditionally for four languages:

| Language | Line | Comment |
|----------|------|---------|
| Python | 176 | `// All Python comment nodes … are regular comments.` |
| Ruby | 205 | `// Ruby doesn't have doc comments` |
| SQL | 217 | `// SQL -- comments have no doc comment convention` |
| Bash | 222 | `// Bash # comments have no doc comment convention` |

The reasoning is correct for *interior* comments (Ruby YARD, Python `#` inline annotations), but wrong at the module-level boundary. A `# SPDX-License-Identifier: MIT` at the top of a `.rb` file, a `#!/usr/bin/env python3\n# Copyright …` block at the top of a `.py` file, or the `# skim-hook vX.Y.Z` pinned version line that `skim init` writes into every agent hook script — all of these are stripped by `is_removable_comment()` because `is_doc_comment` returns `false` and no other path rescues them.

This is called from both `collect_removable_comments` (minimal mode) and the pseudo-mode `collect_noise_ranges` path (see the doc-comment on `is_removable_comment`, line 141–147).

## Why the economics matter

For a tiny file dominated by a header comment, the transform can cost more than it saves. A measured case:

- Raw: 186 B
- Compressed stdout: 166 B (−20 B)
- ADR-008 stderr marker: +277 B (unavoidable transparency cost)
- **Net: +257 B** — compression made the agent's total context larger

`output::guardrail` in `crates/rskim/src/output/guardrail.rs` is the backstop, but line 42 sets `MIN_RAW_SIZE_FOR_GUARDRAIL = 256`. A 186 B file is below the cutoff, so the guardrail never fires, and the worst-ratio cases — where compression provably cannot pay for itself — are the ones that go unprotected.

## Two distinct problems

1. **Correctness / compliance**: dropping SPDX license identifiers, copyright headers, and installer-written provenance lines is wrong regardless of size. A reader who skims a Ruby file should not lose `# frozen_string_literal: true` or `# Copyright 2024 …`.

2. **Guardrail lower-bound blind spot**: `MIN_RAW_SIZE_FOR_GUARDRAIL = 256` is a one-sided cutoff. Below it the check is silently skipped, which is where the negative-ROI cases cluster. The guardrail comment acknowledges "overhead is expected for tiny files" — but this presumes the tiny file has *some* content left after stripping, which is not true when the entire file is a comment block.

## Suggested fix

For languages with no syntactic doc-comment marker, fall back to a **positional heuristic**: preserve `comment` nodes that are the first non-whitespace content in the file (i.e. `node.start_byte()` is near 0, or the node precedes the first non-comment declaration). This preserves shebangs (already handled separately), license blocks, and provenance headers without changing the treatment of inline or interior comments.

Separately, reconsider whether `MIN_RAW_SIZE_FOR_GUARDRAIL` should apply when the *compressed* output is smaller than a threshold, or be removed in favour of always running the token-count slow path — it is cheap and the "expected overhead for tiny files" assumption breaks down for comment-heavy inputs.

## References

`crates/rskim-core/src/transform/minimal.rs` — `is_doc_comment` (line 162), `is_removable_comment` (line 148).
`crates/rskim/src/output/guardrail.rs` — `MIN_RAW_SIZE_FOR_GUARDRAIL` (line 42).
ADR-008 (lossy-view transparency marker), ADR-001 (net-savings guard).

### LINKED ISSUE #478: skim init --wrappers silently no-ops when hook is already current — fast path swallows the flag
`--permissions` has an escape hatch for exactly this case; `--wrappers` does not.

## What happens

`skim init --yes --wrappers` silently does nothing when the hook is already installed and current.

The fast path at `crates/rskim/src/cmd/init/install.rs` line 474:

```rust
if state.hook_installed && state.hook_is_current() && guidance_current && !permissions_blocked {
    print_already_up_to_date();
    return Ok(std::process::ExitCode::SUCCESS);  // line 476
}
```

returns before reaching `maybe_install_wrappers` at line 531. No wrappers are installed or updated, no message is printed, and the exit code is 0 — the caller has no way to tell that the flag was ignored.

## The asymmetry

`--permissions` is covered: `permissions_blocks_fast_path()` (defined at line 163, evaluated at line 472) returns `true` when permissions work is pending, which prevents the early return. There is no equivalent `wrappers_blocks_fast_path()`.

The dry-run path (lines 486–498) calls `maybe_install_wrappers` before returning, so `--dry-run --wrappers` correctly previews what would be installed. Only the live path is affected.

## Impact

A user or agent running `skim init --yes --wrappers` to add PATH-level interception after an initial install will get a clean exit and no wrappers. The correct incantation to force installation is `skim init --yes --wrappers --force` (or uninstall and reinstall), but the silent no-op gives no hint that anything was skipped.

This is particularly sharp for agents following the `CLAUDE.md` pattern of running `skim init --yes --wrappers` as a setup step — the step appears to succeed on repeat runs.

## Suggested fix

Add a `wrappers_blocks_fast_path(flags)` predicate symmetric with `permissions_blocks_fast_path`. The simplest form: return `true` when `flags.wrappers == Some(true)` (explicit opt-in). Integrate it into the fast-path condition at line 474:

```rust
let wrappers_blocked = wrappers_blocks_fast_path(flags);
if state.hook_installed && state.hook_is_current() && guidance_current
    && !permissions_blocked && !wrappers_blocked
{
```

## References

`crates/rskim/src/cmd/init/install.rs` — fast path (lines 474–476), `permissions_blocks_fast_path` (line 163), `maybe_install_wrappers` (line 531, line 688).

### LINKED ISSUE #479: decide: should integrity failure keep suppressing drift detection?
This is a design question, not a bug report. The current behavior is deliberate; this issue asks for an explicit call on whether to keep it.

## The code

`crates/rskim/src/cmd/rewrite/hook.rs`, line 359:

```rust
// Only detect drift if integrity is intact; integrity failure subsumes drift.
let drifts = if !integrity_failed {
    detect_drift(&de)
} else {
    vec![]
};
```

When `check_hook_integrity()` fires — the SHA-256 manifest no longer matches the live script — `detect_drift` is skipped entirely and `drifts` is empty.

## The rationale

The comment states it: integrity failure *subsumes* drift. If the script has been tampered with, whatever version/commit/pin fields it declares are unreliable anyway, so comparing them is noise.

## Why it is worth revisiting

Integrity failure and drift are logged through the same channel (`hook.log`). When integrity fails, drift is silenced on both channels at once. In practice, the two conditions are independent:

- **Integrity failure** means the SHA-256 manifest no longer matches the live hook file (e.g. the file was patched by hand, or a package manager overwrote it).
- **Drift** means the running binary has advanced to a new commit or version while the hook still references the old one.

A machine can have both: a hand-patched hook *and* a stale commit reference. Under the current logic, only the integrity problem surfaces; the drift is invisible until the integrity issue is repaired.

The condition also ran silently until #471 added `skim doctor` integrity reporting. Now that `doctor` loudly surfaces integrity failure (and — after #471 ships — notes that it suppresses drift detection), the coupling is at least visible on demand.

## Options

**Option A — keep as-is.** The "integrity subsumes drift" logic is correct: drift comparisons against a tampered script are unreliable. The #471 doctor change makes the suppression visible. The one-line cost of the current approach is low; the noise of false drift reports on a tampered script is real.

**Option B — decouple.** Run `detect_drift` regardless of integrity state. The drift env (`DriftEnv::from_process()`) reads live process state (`--version`, `argv[0]`), not script text, so it is not contaminated by a tampered script. Log both findings independently. The caller can filter or prioritize.

**Option C — run drift but suppress the log.** Compute drifts unconditionally (for the return value / future callers), but gate the `log_drift_warnings` call on `!integrity_failed`. This keeps the log quiet while making the data available.

## Cross-reference

#471 — `skim doctor` does not check the SHA-256 manifest; that fix makes integrity failure visible in `doctor` and notes the suppression coupling.

### LINKED ISSUE #487: --mode=pseudo -n reports line numbers shifted down by min(K,2) after leading comments are stripped
## Bug

`skim <file> --mode=pseudo -n` reports incorrect source line numbers for any file
whose pseudo output starts with K ≥ 1 blank lines.

## Root cause

`trim_and_normalize` (in `minimal.rs`) drops blank lines before the first non-blank
content (`result.is_empty()` guard). `normalize_line_map_blanks` (in `mod.rs`),
which mirrors `trim_and_normalize` for pseudo mode's line map, previously did NOT
mirror this rule — it only capped 3+ consecutive blanks. Result: the map has
`min(K, 2)` extra leading entries. `format_with_line_numbers` is a positional zip,
so every output line's reported source number is shifted down by `min(K, 2)`.

## Minimal repro

```python
# blank first line produces K=1 leading blank in pre-normalised pseudo text

# Copyright 2024 Acme Corp.
# 18 more header lines...

import os

def compute(x: int, y: int) -> int:   # source line 25
    return x + y
```

```bash
skim demo.py -n --mode=pseudo
# Before fix: def at line 24 (off by 1)
# After fix:  def at line 25 (correct)
```

## Fix

Add `if result.is_empty() { continue; }` guard in `normalize_line_map_blanks` for
blank lines, mirroring `trim_and_normalize`'s leading-blank drop. Also added:
- Invariant test: `normalize_line_map_blanks(t, m).len() == trim_and_normalize(t).lines().count()`
- K=1 and K=5 unit tests
- E2E test in `cli_line_numbers.rs` using `parse_annotated` harness

Fixed in this PR as part of the five-bug batch.

## Affected modes

Only `pseudo` mode. `minimal` uses `compute_line_map_by_text_matching` (text-content
matching, immune to this class of bug). `structure`/`signatures`/`types` never call
`trim_and_normalize`.

dean0x:
## Summary

24 commits in two batches on `fix/init-pin-wrappers-header-comments`.

**Batch 1** hardens hook provenance — `skim init` binary-pin resolution, the `--wrappers`
and `--force` fast-path gates, the tampered-script repair loop — and stops minimal/pseudo
from silently stripping module-header comments (SPDX, `frozen_string_literal`, provenance
lines) in Python/Ruby/SQL/Bash.

**Batch 2** makes every wrapper output tier *stream*, so a closed downstream pipe behaves
like the raw tool instead of erroring or panicking, and linearizes two super-linear
transform paths.

The full ledger — what shipped, what is only *partially* addressed, and what was raised
during the work and is tracked nowhere else — lives in the tracking issue: **#505**.

### Batch 1 — hook provenance / init / doctor (12 commits)

| SHA | What | Issues |
|---|---|---|
| `f5c8327` | init pin gate, `--wrappers` fast-path, module-header comments, pseudo line-map | #477 #478 #476 #479 + line-map (#487) |
| `0ca08ac` | doctor: binary-pin-mismatch path, drop dead stale branch, tarball-build misreport | — |
| `9ebeb02` | `skim_sandboxed_with_bin` — shuts the PF-017 env-leak gap | — |
| `a012058` | KB: hook-binary-pinning | — |
| `f00e37a` | KB: analytics | — |
| `7bb8cac` | **perf: O(N³) module-header comment detection → linear** | — |
| `37c87d6` | init `--force` wired; doctor out-of-repo drift false-positive | — |
| `088c431` | quadratic-scaling guard always asserts | — |
| `9be0d67` | documentation drift sweep | — |
| `d2f5e12` | shut the tampered-script repair loop; wrapper pin invariant | — |
| `5909b06` | review artefacts (later reverted) | — |
| `c09f7cf` | Revert the above — `.devflow/docs/` is ignore-by-default per #341 | — |

### Batch 2 — pipe fidelity + transform linearization (12 commits)

| SHA | What | Issues |
|---|---|---|
| `9f2cdb5` | tolerate a shut downstream pipe; **never exit 1 on EPIPE** | (#495) |
| `e12502c` | stream the raw-passthrough tier | #495 |
| `8f4a37c` | stream the `SKIM_PASSTHROUGH=1` escape hatch | #495 #317 |
| `33a5e98` | stop `println!` panicking on a shut pipe (101 → 141) | (#495) |
| `4e6eca7` | stream `run_raw_passthrough`, the missed sibling surface | #370 |
| `e98c450` | **Go doc-comment detection Θ(M³/3) → O(N)** | #494 (item 3 only) |
| `85fbcdc` | rustfmt | — |
| `56239f6` | thread walk position through the pseudo walker | #494 |
| `9633c84` | abandon the stderr drain on early close — joining it is unbounded | — |
| `3e7372a` | bound test reads, kill three vacuous asserts, retract a scope over-claim | #370 |
| `4c85233` | pin the anonymous-sibling trap in the Go differential | — |
| `0bdda12` | correct the streamed-passthrough analytics parity claim | — |

### Measured outcomes

| | before | after | raw tool |
|---|---|---|---|
| `skim grep … \| head -20` exit code | **1** (= "no matches") | **141** | 141 |
| stderr on early close | `Error: Broken pipe` | 0 bytes | 0 bytes |
| `find` time-to-first-byte | 0.304 s | **0.013 s** | 0.009 s |
| slow producer, reader closes at 2 s | **1 line** | **801 lines** | 701 |
| 70 MiB via `SKIM_PASSTHROUGH=1` | **0 bytes, exit 1** | **73,400,320 bytes, exit 0** | same |
| `skim git diff \| head -2` | **exit 101 + panic** | 141, silent | 141 |
| non-UTF-8 bytes | U+FFFD corruption | verbatim | — |
| Go doc comments, N=1000 | **41,615 ms** (α=2.98) | **1.48 ms** (α=0.98) | — |
| grandchild holding fd 2, `\| head -3` | **26.50 s** | 1.39 s | 0.33 s |

Gates: `rskim` 4454 passed / 0 failed / 2 skipped · `rskim-core` 555 + 16 doctests ·
clippy `--all-features` clean · fmt clean. Byte-identity sweep run per transform commit:
76 fixtures × 6 modes × (stdout + stderr + exit) = **1368 compared outputs, empty diff**.

## Breaking Changes

None at the CLI surface. Two intentional behavioural changes:

- **minimal/pseudo retain module header comments** for Python, Ruby, SQL, and Bash.
  Slightly lower token reduction on comment-heavy files, in exchange for not silently
  dropping legally required or provenance-critical headers.
- **Wrapper tiers now exit 141 on a shut downstream pipe** (previously `1`, which is
  indistinguishable from "no matches", or `101` + a panic message on the `println!`
  paths). This matches the raw tool.

## High-risk areas for review

The five-fix walkthrough that used to be here is gone: it described `f5c8327` only, and
pointed at `wrappers_blocks_fast_path` and `is_hook_script_current`, both deleted by
`37c87d6`. The genuinely load-bearing parts of the current diff are:

**1. `crates/rskim/src/cmd/stream_pump.rs` — the concurrency ordering.** Three orderings
are correctness-critical and none of them is enforced by the type system:

- **`ChildGuard` wraps the child *at spawn*** (`stream_child`, before any read), so
  kill-on-drop covers every early-return path. Moving the wrap later reintroduces a tool
  that keeps working after its reader is gone.
- **The stderr drain thread starts *before the first stdout read*.** Reversed, a child
  that fills the stderr pipe blocks forever while skim waits on stdout.
- **`drain_capped` keeps reading after it hits `MAX_OUTPUT_BYTES`** — it stops *retaining*
  and sets `discarded`, but must keep consuming, for the same reason.
- **On `PumpStop::PipeClosed` the drain thread is abandoned, never joined.** This is the
  subtle one: `child.kill()` closes only the child's *own* copy of the stderr write end.
  A grandchild that inherited fd 2 (`cargo` → `rustc`, `yarn` → `node`) keeps the pipe
  open, so `drain_capped` may never reach EOF and a join there has **no bound** — the
  26.50 s → 1.39 s row in the table above is exactly that. Please sanity-check the claim
  that dropping the handle is safe here (the outcome carries no stderr payload).

This hazard is cited in-tree as `PF-023`, which **does not exist in the ledger** (highest
is PF-021) — see #505 §5.

**2. The ADR-011 marker-vs-banner split.** Every stderr notice added or moved in batch 2
had to be classified: *loss-bearing markers* (elision markers with exact counts and the
`SKIM_PASSTHROUGH=1` hint, the ADR-008 transparency marker) are **unconditional**;
*no-loss raw-fallback banners* are **gated behind `SKIM_DEBUG`/`--debug`**. Worth checking
that nothing which can fire while the reader sees less-than-raw ended up debug-gated —
that is the failure direction that loses data silently.

**3. Go doc-comment classification equivalence** (`e98c450`, `4c85233`). `is_go_doc_comment`
went from a repeated backwards sibling walk to a single cursor scan. The risk is not
performance, it is *classification drift*: the differential-equivalence tests and the
1999-line `large_doc_blocks.go` fixture exist to pin old-vs-new agreement, including the
anonymous-sibling trap `4c85233` addresses. Confirm the differential actually covers the
anonymous-node case rather than only named siblings.

## Closes

Closes #478
Closes #479
Closes #487

## Refs — does NOT close

The issues below are referenced for context only. Every one of them stays **open** after
this merges — please do not close them out of the commit chain, which reads more
"complete" than the delivered semantics warrant. Breakdown and sequencing in #505.

- **#476** — Problem 1 (module-header comments) shipped. Problem 2 is untouched:
  `MIN_RAW_SIZE_FOR_GUARDRAIL = 256` at `crates/rskim/src/output/guardrail.rs:42` is not in
  this diff at all, and the issue's 186 B worked example still nets **+257 B**.
- **#477** — the fix reversed direction mid-branch. The issue asks for an install-time
  **gate**; `37c87d6` + ADR-014 demoted it to a **doctor advisory** with `--force` as the
  override. Maintainer decision, not a merge side effect.
- **#493** — the line-map desync is gone and an invariant test pins it, but the
  *structural* complaint stands: `trim_and_normalize` and `normalize_line_map_blanks` are
  still two functions hand-mirroring the same two rules (PF-019).
- **#494** — item **3** only (`is_go_doc_comment` linearized, 28,118× at N=1000). Item 1
  (`compute_header_end_byte` still unbounded), item 2 (SQL `marginalia`, zero hits in
  tree), and item 4 (the criterion cold-cache benchmark — `crates/rskim-bench/benches/`
  does not exist) all remain.
- **#495** — the SIGPIPE/buffering half shipped across six commits. The compound-command
  **rewrite-engine** item did not: every test here drives `cmd::dispatch`, and per
  CLAUDE.md the two interception surfaces share handlers but not the dispatch front-end.
  `3e7372a` exists to retract an earlier over-claim that it was covered.
- **#503** — Finding 7 (rust-10/testing-11) was addressed incidentally by `e98c450`.
  Findings 1–6 remain.

---

### Fix 4 (#479) decision rationale

`detect_drift` is correctly skipped when the hook script is `Tampered`, because the
issue's premise for decoupling is wrong. `DriftEnv::from_process()` reads six fields;
three are binary-derived and tamper-proof (`current_exe()` — the kernel's exe path,
not `argv[0]`; `env!`/`option_env!` compile-time constants). The other three —
`hook_version`, `hook_binary`, `hook_commit` — are read from environment variables
the hook script itself exports (`hooks/mod.rs:573-585`). A tampered script is
therefore one entire side of every drift comparison; editing `SKIM_HOOK_COMMIT` erases
`CommitMismatch`, and deleting the exports makes `detect_drift` return empty outright
via its scoping rule (`hook.rs:126-130`). Option C (compute but don't log) is dead:
`drifts` is constructed, tested, logged, and dies at the same call-site — "compute
but don't log" is behaviorally identical to option A.


crates/rskim-core/src/transform/minimal.rs
```diff
@@ -271,6 +272,64 @@ fn is_go_declaration(kind: &str) -> bool {
...
+/// `frozen_string_literal: true`, provenance markers, and FIXTURE/TESTS headers.
+/// No doc-comment convention exists for these languages (`is_doc_comment` returns
+/// `false`), so without this guard minimal/pseudo would strip them.
+///
+/// # Blank-line break
+///
+/// A single newline between two consecutive comment lines means no blank line
+/// (adjacent). Two or more newlines means at least one blank line exists between
+/// them, which ends the header block — comments after the break are strippable.
+fn is_module_header_comment(node: Node, source: &str, language: Language) -> bool {
```
dean0x: **[BLOCKING] Unbounded O(N³) backwards walk — defeats the existing `MAX_AST_NODES` bound**

Converging reviewers: security (90%), performance (90%), reliability (92%), rust (85%), complexity (90%), regression (80%) — 6 independent reviewers.

For a contiguous run of N root-level comment nodes with no blank-line break, comment at index `k` walks back `k` siblings. Total backward steps = N(N-1)/2. `prev_named_sibling()` in tree-sitter 0.25 is not O(1) — it re-scans the parent child list from the front, so one call from sibling index `k` costs O(k). Realized complexity: **O(N³)** for the run, not O(N²). At N=1000 (a large SQL dump banner or commented-out prologue) that is ~500k backward steps × O(k) per step. Measured: ~25ms at N=1000, ~14s at N=20000 — far past the stated `<50ms / 1000 lines` budget.

This is also on the `pseudo` hot path — `pseudo.rs:446` calls the same `is_removable_comment`, and pseudo is what `mode_for_files` selects for every `cat`/`head`/`tail` rewrite. `MAX_AST_NODES = 100_000` does **not** help: the counter increments at `minimal.rs:86`, before `is_removable_comment` at `:95` — it bounds nodes visited, not work per node. The loop at `minimal.rs:303` also has no iteration cap, violating the project rule ("All loops and retries must have a fixed upper bound").

**Fix (preferred — O(N) total):** compute the header block extent once per file with a single forward pass, then make the per-node predicate an O(1) byte compare:

```rust
pub(crate) fn module_header_end_byte(root: Node, source: &str, language: Language) -> usize {
    if !matches!(language, Language::Python | Language::Ruby | Language::Sql | Language::Bash) {
        return 0;
    }
    let mut end = 0usize;
    let mut prev_end: Option<usize> = None;
    let mut cursor = root.walk();
    for child in root.named_children(&mut cursor) {
        if !is_comment_node(child.kind(), language) { break; }
        if let Some(pe) = prev_end
            && source.get(pe..child.start_byte())
                .is_some_and(|g| g.bytes().filter(|&b| b == bn).take(2).count() > 1)
        { break; }
        end = child.end_byte();
        prev_end = Some(end);
    }
    end
}
// is_module_header_comment becomes: node.end_byte() <= header_end  (O(1))
```

Thread `header_end` through `CommentWalkContext` and the pseudo `ctx`. **Minimum acceptable stopgap:** `const MAX_HEADER_WALK: usize = 512;` mirroring `MAX_PARENT_WALK` — but this caps to O(N²), not O(N), and semantics degrade for headers longer than the bound. Also add a regression fixture with 2000 contiguous comment lines asserting wall-clock completion; the current five unit tests are all 2–4 lines and cannot observe this.

---
*Posted by Claude Code*

crates/rskim/src/cmd/init/install.rs
```diff
@@ -833,6 +861,19 @@ fn is_hook_script_current(script_path: &std::path::Path, version: &str) -> bool
             _ => return false,
         }
     }
+    // #477: also require the binary pin to match the running binary.
+    // When two clones share the same version and commit the version+commit checks
+    // above both pass, but the hook still pins the wrong clone. Comparing the
+    // canonical paths catches this case and triggers a rewrite.
+    if let Some(pin) = parse_binary_pin_from_script(&contents)
```
dean0x: **[BLOCKING] `is_hook_script_current` fails OPEN on absent/empty pin — `skim init` never converges, `skim doctor` stays at exit 1 forever**

Converging reviewers: security (85%), consistency (90%), regression (85%), rust (85%), reliability (88%), complexity (80%), architecture (88%) — 7 independent reviewers. See also counterpart at `state.rs:59`.

The `if let Some(pin) = parse_binary_pin_from_script(&contents) && let Ok(running) = ...` chain **silently skips the pin check** when the pin is absent or unparseable, returning `true` ("already current"). `DetectedState::pin_is_current()` (`state.rs:59`) treats the identical condition as `false` ("stale"). These two gates sit **in series** on the same CLI path and produce opposite verdicts.

Concrete non-convergence loop for a script with `export SKIM_HOOK_BINARY=` (the state that pre-B5b builds produced — documented at `install.rs:930–934`):
1. `pin_is_current()` = false → fast path bypassed at `:506` → falls through to `create_hook_script`
2. `is_hook_script_current()` → pin check skipped → returns `true` → prints "Skipped: … (already vX)" → returns without rewriting
3. `skim doctor` → `!pin_is_current` → "binary pin mismatch (hook: ?, running: …)" → **exit 1** with advice "run `skim init`"
4. Go to 1. No escape — `--force` does not exist, no documented operator path

Also, `script_has_pinned_marker` already passed at `:857`, so `None` from `parse_binary_pin_from_script` means the marker exists but the value is malformed — a state that **must** force a rewrite, not skip one. This is PF-015 instance (3) re-introduced: a second gate between the CLI entry point and the predicate that was hardened.

**Fix — make failure direction match `pin_is_current` (fail closed):**

```rust
// install.rs:864-876  →  replace the if-let chain
let Some(pin) = parse_binary_pin_from_script(&contents) else {
    return false; // marker present but pin absent/malformed → force rewrite
};
let Ok(running) = super::helpers::resolve_skim_binary() else {
    return false; // cannot resolve → treat as stale (mirrors pin_is_current Err arm)
};
let pin_path = std::path::Path::new(pin.as_str());
let canon_pin = std::fs::canonicalize(pin_path).unwrap_or_else(|_| pin_path.to_owned());
if running != canon_pin { return false; }
```

Also add a unit test asserting `is_hook_script_current` returns `false` for a script with `export SKIM_HOOK_BINARY=''` and a matching version+commit — that case is currently unreachable from the test suite.

---
*Posted by Claude Code*

crates/rskim/src/cmd/init/state.rs
```diff
@@ -40,6 +40,41 @@ pub(super) struct DetectedState {
...
+    /// SHA), all three of those fields match — yet the hook still points to the
+    /// wrong binary. `pin_is_current` catches that remaining case by comparing
+    /// the actual file-system path, not just the version/commit metadata.
+    ///
+    /// `skim doctor` can display the `hook_binary_pin` field independently
+    /// (PF-015: display-without-gate pattern), which is why this predicate is
+    /// separate: it lets the doctor report a pin-mismatch cause with
+    /// `hook_is_current=true, pin_is_current=false` before the fix would have
+    /// been taken by the former fast-path check.
+    pub(super) fn pin_is_current(&self) -> bool {
```
dean0x, line 59: **[BLOCKING] `pin_is_current` vs `is_hook_script_current` (install.rs:868): two gates, one invariant, opposite failure directions**

Converging reviewers: security (85%), consistency (90%), regression (85%), rust (85%), reliability (88%), complexity (80%), architecture (88%) — 7 independent reviewers. See counterpart comment at `install.rs:868`.

`pin_is_current` correctly fails **closed** when the pin is absent (`None` → `false` at `:60–61`) or when `resolve_skim_binary()` errors (`Err(_)` → `false` at `:74`). Its counterpart `is_hook_script_current` (`install.rs:868`) fails **open** in both cases. The two gates sit in series: `pin_is_current` blocks the fast path at `install.rs:506`; `is_hook_script_current` decides whether to actually rewrite at `install.rs:896`.

Two additional issues in this function:

1. **`pin_is_current` ignores `self.skim_binary`** even though `detect_state` (`state.rs:121`) already populates that field from `resolve_skim_binary()`. The method re-calls `resolve_skim_binary()` at predicate-evaluation time, making it non-pure, non-hermetic, and redoing I/O on every call. `hook_is_current()` is a pure query over struct fields; `pin_is_current()` should be too. Fix: `self.skim_binary == canonicalize(pinned_path).unwrap_or(pinned_path)`.

2. **Re-canonicalizing the pin loosens the comparison** (security, 82%): the pin was **written** as the canonical path by `resolve_skim_binary()` → `create_hook_script`. For any legitimately-installed script `canonicalize(pin) == pin` and the extra call is a no-op. The only inputs whose value changes are externally-modified pins — precisely the case the check exists to detect. A pin of `/tmp/x` where `/tmp/x → <running binary>` now compares equal (TOCTOU gap on that symlink), for zero benefit on clean installs. Compare the pin as-recorded: `self.skim_binary.as_path() == Path::new(pinned.as_str())`.

---
*Posted by Claude Code*

crates/rskim/src/cmd/doctor/mod.rs
```diff
@@ -457,17 +456,36 @@ fn hook_status_line(
...
         let reason = if !commit_ok {
             format!("commit mismatch (hook: {hook_commit_str}, binary: {compiled_commit})")
         } else if !version_ok {
             format!("version mismatch (hook: {hook_version}, binary: {compiled_version})")
         } else {
-            "stale".to_string()
+            // Version and commit match but the pin path differs: two clones at
+            // the same commit.  Show both paths so the user can identify which
+            // binary is running and which the hook points to.
+            let running = std::env::current_exe()
```
dean0x: **[BLOCKING] Fourth inline derivation of the canonical binary path — breaks `hook_status_line`'s purity contract and is untestable**

Converging reviewers: consistency (96%), rust (90%), reliability (90%), complexity (95%), architecture (92%) — 5 independent reviewers.

This PR's stated goal is unifying binary-path derivation behind `resolve_skim_binary()`. This hunk adds a **fourth** independent computation of the exact same value, in the same file that already has `current_exe_canonical()` at `:96` (called at `:37`, ~450 lines above).

Three distinct defects in five lines:

1. **`current_exe_canonical()` already exists here.** The `and_then(|p| canonicalize(&p).ok().or(Some(p)))` form is `current_exe_canonical()` written with an `and_then` that can never return `None` (`.ok().or(Some(p))` is total). A reader must prove that before concluding the `None` arm is unreachable — dead reasoning for a value already computed by a named local helper.

2. **DI violation — breaks the function's documented purity.** `hook_status_line` takes `compiled_version` and `compiled_commit` as parameters *because* it is documented as a pure, fully-assertable function. Every other environment-derived value arrives as a parameter; this one is pulled from process state. The three new unit tests can only assert `line.contains("running:")` — they cannot fix the expected path, cannot construct a pin-equals-running case, and cannot assert the branch does **not** fire when paths match. PF-015 defect (2) verbatim: "tests constructing already-populated values exercise the DISPLAY layer only."

3. **Display path diverges from comparison path.** The comparison in `pin_is_current()` uses `resolve_skim_binary()`; this display uses `current_exe_canonical()`. If they ever diverge (e.g. `helpers.rs:33`'s silent fallback is changed here but not there), doctor prints `binary pin mismatch (hook: X, running: X)` — two identical paths, a mismatch verdict, and no actionable diagnosis.

**Fix (closes all three):** add `running_binary: Option<PathBuf>` to `HookFacts`, populated by `helpers::resolve_skim_binary().ok()` in `hook_facts()`. Then:

```rust
// doctor/mod.rs:483-487 — replace the inline derivation
let running = facts.running_binary.as_ref()
    .map(|p| p.to_string_lossy().into_owned())
    .unwrap_or_else(|| "?".to_string());
```

This kills the fourth derivation site, restores purity, and makes the full path assertable in the existing unit tests. A cheaper interim: `let running = current_exe_canonical().map(|p| p.to_string_lossy().into_owned()).unwrap_or_else(|| "?".to_string());` — at minimum call the existing local helper rather than re-inlining it.

---
*Posted by Claude Code*

crates/rskim/src/cmd/init/install.rs
```diff
@@ -153,6 +154,26 @@ fn print_completion_message(agent_name: &str) {
...
+/// Rule (mirrors `permissions_blocks_fast_path`):
+/// - `Some(true)`  → always block (explicit `--wrappers` flag).
+/// - `Some(false)` → never block (explicit `--no-wrappers` flag).
+/// - `None`        → never block — this is load-bearing.  When no wrapper flag is
+///   given, `maybe_install_wrappers` prompts the user interactively (on a TTY) or
+///   skips silently (non-TTY).  If `None` were to block the fast path, every
+///   non-TTY `skim init` invocation (CI, test harness) would fall through and
+///   reinstall the hook on each run, breaking idempotence tests such as
+///   `test_init_skips_when_version_and_commit_are_current`.
+fn wrappers_blocks_fast_path(flags: &InitFlags) -> bool {
```
dean0x: **[BLOCKING] `wrappers_blocks_fast_path` ignores `flags.project` — permanent no-op reinstall churn for `skim init --project --wrappers`**

Converging reviewers: regression (90%), rust (92%).

`wrappers_blocks_fast_path` returns `true` for `Some(true)` unconditionally, but `maybe_install_wrappers` is gated on `!flags.project` at **both** call sites (`install.rs:532` dry-run, `install.rs:567` real). `--wrappers` is not rejected alongside `--project` in `flags.rs` (unlike `--permissions`, which has an explicit mutual-exclusion guard at `:389–395`).

Result for `skim init --project --wrappers --yes` on a fully-current install: the fast path is permanently blocked → full `execute_install` runs on every invocation → `settings.json.bak` clobbered, guidance reinjected, SHA-256 manifest rewritten, search git hooks installed into cwd, a detached `skim search --build` spawned — **and zero wrappers installed**, because `install.rs:567` skips `maybe_install_wrappers` for project scope. PF-018 verbatim: "an idempotence fast path is a predicate over the UNION of effects the command performs."

The three new unit tests (`test_wrappers_blocks_fast_path_{some_true,some_false,none}_*`) all build flags with `project: false`, so this case is unreachable from CI.

**Fix:**

```rust
fn wrappers_blocks_fast_path(flags: &InitFlags) -> bool {
    match flags.wrappers {
        // Only block when wrapper installation will actually run:
        // maybe_install_wrappers is guarded by !flags.project at :532 and :567.
        Some(true) => !flags.project,
        Some(false) => false,
        None => false, // load-bearing — see doc comment
    }
}
```

Or add the mutual-exclusion guard in `flags.rs` mirroring the `--permissions` check. Either way, add a test with `project: true` to the existing suite.

---
*Posted by Claude Code*

crates/rskim/tests/cli_init.rs
```diff
@@ -1719,3 +1719,100 @@ fn test_init_skips_when_version_and_commit_are_current() {
...
+// ============================================================================
+// Fix 1: pin-path mismatch triggers reinstall (#477)
+// ============================================================================
+
+/// When the hook script's `SKIM_HOOK_BINARY` pin points to a different path
+/// than the running binary (same version, same commit — two-clone scenario),
+/// `skim init --yes` must NOT report "Already up to date" and must rewrite
+/// the script so the pin points to the correct binary.
+#[test]
+fn test_init_rewrites_hook_when_pin_path_differs() {
```
dean0x: **[BLOCKING] New test bypasses `skim_sandboxed` — the sandbox helper this same PR extracted to close PF-017**

Converging reviewers: consistency (92%), testing (90%), rust (85%), compliance (90%) — 4 reviewers at >=80%. Also flagged by security (75%).

`skim_init_cmd(config)` sets only `CLAUDE_CONFIG_DIR`. It leaves `HOME`, `SKIM_CACHE_DIR`, `SKIM_WRAPPERS_DIR`, `GEMINI_CONFIG_DIR`, `CODEX_HOME`, and `CRUSH_CONFIG_DIR` pointing at the developer's real home. This is the exact pattern PF-017 names as the root cause ("sandbox `$HOME` itself rather than enumerating app-specific overrides — an enumeration is always incomplete"), and this PR's own `skim_sandboxed_with_bin` extraction exists to stop it.

The sibling test added ~46 lines later (`test_init_wrappers_bypasses_fast_path`, `:1785`) correctly routes through `common::skim_sandboxed`. The blast radius today is narrow (override mode limits `detect_installed_agents` to claude-code; non-TTY skips wrapper install). **The hazard is latent**: adding `--wrappers` to this test re-arms real `~/.skim/bin` writes with no test failure to warn you.

**Fix — route through the authoritative helper matching its sibling:**

```rust
fn test_init_rewrites_hook_when_pin_path_differs() {
    let home = TempDir::new().unwrap();
    let home_path = home.path();
    // detect_installed_agents() in override mode requires the dir to exist.
    fs::create_dir_all(home_path.join(".claude")).unwrap();

    common::skim_sandboxed(home_path)
        .args(["init", "--yes"])
        .assert()
        .success();

    let hook_path = home_path.join(".claude/hooks/skim-rewrite.sh");
    // ... remaining steps unchanged, reading from hook_path ...
}
```

---
*Posted by Claude Code*

crates/rskim/src/cmd/init/helpers.rs
```diff
@@ -11,6 +11,28 @@ use std::path::{Path, PathBuf};
...
+/// raw path from `current_exe()` rather than failing, preserving the behaviour
+/// of the pre-unification code at `install.rs:895-906`.
+pub(super) fn resolve_skim_binary() -> anyhow::Result<PathBuf> {
+    let p = std::env::current_exe().map_err(|e| {
+        anyhow::anyhow!(
+            "cannot determine the skim binary path: {e}\n\
+             hint: re-run `skim init` from the skim binary directly"
+        )
+    })?;
+    Ok(std::fs::canonicalize(&p).unwrap_or(p))
```
dean0x, line 32: **[SHOULD FIX] `resolve_skim_binary()` swallows `canonicalize` failure silently — applying opposite policy to its own `current_exe` half**

Reviewer: reliability (84%).

`current_exe()` failure is handled exemplarily — `map_err` with an actionable hint at `:27–32`. `canonicalize()` failure at `:33` uses `unwrap_or(p)` with no diagnostic on any channel. The two halves of one function apply opposite fail-loud policies.

When the binary was replaced or unlinked while running, `canonicalize` fails and the raw (possibly `(deleted)`) path flows unchecked into all three unified consumers:
- `create_hook_script` embeds it as `SKIM_HOOK_BINARY` — the `if [ -x "$_SKIM_BIN" ]` guard always fails, so every hook invocation silently falls through to bare `exec skim` on `$PATH` — **precisely the wrong-clone hazard ADR-004 exists to eliminate**, re-armed silently.
- `maybe_install_wrappers` creates 8 symlinks pointing at a non-existent target.
- `detect_state` stores it as the comparison basis, so `pin_is_current()` then misreports.

The comment at `helpers.rs:16–25` asserts this is the "single source of truth" — unifying three sites behind one silent fallback unified three silent failures.

**Fix:** keep the fallback (correct for exotic layouts) but log the degrade per ADR-011 (no-loss fallback → `SKIM_DEBUG`-gated):

```rust
match std::fs::canonicalize(&p) {
    Ok(c) => Ok(c),
    Err(e) => {
        crate::debug_log!("resolve_skim_binary: canonicalize({}) failed: {e}", p.display());
        Ok(p)
    }
}
```

And in `create_hook_script`, before embedding the path, assert it is executable — a broken pin should be refused rather than installed:

```rust
anyhow::ensure!(
    binary_path.exists(),
    "resolved skim binary {} does not exist — refusing to write an unusable hook pin (ADR-004); \
     hint: reinstall skim, then re-run `skim init`",
    binary_path.display()
);
```

---
*Posted by Claude Code*

crates/rskim/src/cmd/doctor/mod.rs
```diff
@@ -457,17 +456,36 @@ fn hook_status_line(
...
+    if !facts.hook_is_current || !facts.pin_is_current {
         let pin = facts.hook_binary_pin.as_deref().unwrap_or("?");
         let version_ok = facts.hook_version.as_deref() == Some(compiled_version);
-        let commit_ok = facts.hook_commit.as_deref() == Some(compiled_commit);
-        // Report the most specific cause first.
+        // Mirror hook_is_current()'s treatment of tarball/non-git builds: when
+        // the compiled commit is "unknown" the comparison is not determinable and
+        // must not be reported as a mismatch — doing so would misattribute a
+        // genuine pin mismatch as a "commit mismatch" (Defect 3).
+        let commit_ok = if compiled_commit == "unknown" {
```
dean0x: **[BLOCKING] "unknown"-commit rule now triplicated across three independent sites with no shared predicate and no test asserting equivalence**

Reviewer: complexity (90%).

The same special case — "compiled commit == `\"unknown\"` (tarball/non-git build) ⇒ commit comparison is indeterminate ⇒ treat as OK" — is now implemented independently in three places:

1. `state.rs:97–108` — `hook_is_current()`
2. `install.rs:856–863` — `is_hook_script_current()`
3. `doctor/mod.rs:466–470` — `hook_status_line()` ← **new in this PR**

The dead-code removal of the `"stale"` terminal (`doctor/mod.rs:474`) is correct **only because** these three copies of the rule currently agree. There is no shared predicate and no test asserting the equivalence. If any one copy drifts — say `hook_is_current()` gains a fourth condition — the `else` at `:479` stops being a proof and becomes a fabricated diagnostic: it prints `binary pin mismatch (hook: X, running: Y)` unconditionally, including when `X == Y`. The removed `"stale"` safety net previously absorbed exactly that class of drift.

**Fix:** extract one shared function:

```rust
// init/state.rs — callable from all three sites
pub(crate) fn commit_matches(hook_commit: Option<&str>, compiled_commit: &str) -> bool {
    if compiled_commit == "unknown" { return true; } // tarball/non-git build — indeterminate
    hook_commit == Some(compiled_commit)
}
```

Then replace `let commit_ok = if compiled_commit == "unknown" { true } else { ... }` here with `commit_matches(facts.hook_commit.as_deref(), compiled_commit)`. Add one test asserting `hook_is_current() == (version_ok && commit_matches(..))` so a future divergence fails loudly rather than silently invalidating the `else` removal.

---
*Posted by Claude Code*

crates/rskim/src/cmd/init/install.rs
```diff
@@ -479,10 +500,13 @@ fn run_install_single(
     let manifest_present =
         crate::cmd::integrity::read_hash_manifest(&state.hook_config_dir, state.agent_cli_name)
             .is_some();
+    let wrappers_blocked = wrappers_blocks_fast_path(flags);
     if state.hook_installed
```
dean0x, line 496: **[SHOULD FIX] Seven-term mixed-polarity fast-path condition with hidden I/O; `--wrappers` re-run triggers full install side-effects including clobbering `settings.json.bak`**

Reviewers: complexity (90%), regression (85%).

**Complexity concern (complexity, 90%):** the condition grew from 5 to 7 terms with this PR. Current shape: five positively-named terms and two negated `*_blocks_fast_path` results, mixed polarity that requires mental inversion mid-expression. Three terms (`hook_is_current()`, `pin_is_current()`, `wrappers_blocks_fast_path`) perform or contain syscalls despite appearing as simple boolean accessors. Every prior bug in this area (#466, #471, #477, #478) was a mis-gated condition.

Consider naming it with a reason type:

```rust
enum FastPathBlocker { NotInstalled, HookStale, PinStale, GuidanceStale, Permissions, Wrappers, ManifestMissing }
fn fast_path_blocker(state: &DetectedState, ...) -> Option<FastPathBlocker> { ... }
if fast_path_blocker(...).is_none() { print_already_up_to_date(); return ...; }
```

**Side-effects concern (regression, 85%):** `--wrappers` on an otherwise-current install now falls through the **entire** `execute_install` pipeline just to reach `maybe_install_wrappers` at `:567`. Along the way: `backup_settings` (`install.rs:1271–1286`) unconditionally `fs::copy`s the current `settings.json` over `settings.json.bak` — **destroying the user's pre-skim backup** on a routine `skim init --wrappers` re-run. Also fires `install_search_integration` (git hooks into cwd, detached `skim search --build`). Previously this invocation printed "Already up to date" and wrote nothing.

Surgical fix that avoids both issues: keep wrappers out of the fast-path conjunction and run them inside the fast path before returning:

```rust
if state.hook_installed && state.hook_is_current() && state.pin_is_current()
    && guidance_current && !permissions_blocked && manifest_present
{
    print_already_up_to_date();
    if !flags.project {
        maybe_install_wrappers(flags.wrappers, flags.dry_run)?;
    }
    return Ok(std::process::ExitCode::SUCCESS);
}
```

This also resolves the `--project --wrappers` infinite-loop finding at `:169` as a side effect.

---
*Posted by Claude Code*

.devflow/features/analytics/KNOWLEDGE.md
```diff
@@ -0,0 +1,154 @@
...
+  - crates/rskim/tests/common/mod.rs
+created: 2026-06-25
+updated: 2026-06-25
+---
+
+# File-Op Analytics Recording & Cache-Dir Resolution
+
+## Overview
+
+This subsystem records per-invocation token-savings data to a local SQLite database (`~/.cache/skim/analytics.db`) so `skim stats` can show dashboards. The core challenge — which produced PF-001 — is that recording must happen AFTER stdout is flushed (streaming contract), token counting is expensive, and short-lived processes exit before background threads finish writing. The subsystem also resolves where all cache state lives; a prior drift (PF-002) caused `SKIM_CACHE_DIR` to relocate only the parser cache but not analytics.db.
```
dean0x: **[BLOCKING] PF-002 is cited four times in this KB but does not exist**

Reviewer: documentation (95%).

`grep -rn "PF-002" .devflow/` returns zero matches — not in `pitfalls.md`, not in the decisions ledger, nowhere. The KB builds an entire narrative on it (`:24`, `:126`, `:144`, `:154`). The citation rule in `devflow:apply-decisions` treats IDs as verbatim-only: an agent reading this KB will try to look up PF-002, find nothing, and cannot tell whether the pitfall was retired, mis-numbered, or invented. Fabricated citations destroy trust in the KB's other citations (ADR-001, PF-001 — both of which do exist and check out).

**Fix:** either (a) file the real pitfall through `assign-anchor` and cite the returned ID, or (b) describe the drift narratively without an ID — the mechanism is already fully explained in Rule 6, so no context is lost. Suggested replacement for this line:

> "The core challenge — which produced PF-001 — is that recording must happen AFTER stdout is flushed... The subsystem also resolves where all cache state lives; an earlier drift caused `SKIM_CACHE_DIR` to relocate only the parser cache but not analytics.db (see Rule 6)."

---
*Posted by Claude Code*

.devflow/features/analytics/KNOWLEDGE.md
```diff
@@ -0,0 +1,154 @@
...
+
+`FileCounts::Known { raw, compressed }` — counts were already computed (e.g. `--show-stats` was on, or a count-carrying cache hit). No re-read, no double tokenization work.
+
+`FileCounts::Tokenize { raw: RawSource, compressed }` — plain run / cold cache. For single files the background thread re-reads the file via `read_source`. For stdin the raw buffer is captured in `ProcessResult.stdin_raw` (stdin cannot be re-read) and moved as `RawSource::Inline`. TOCTOU: if the file is deleted or grown past the 50 MB guard between the main-thread run and the background re-read, `read_source` returns `Err` and the row is silently dropped via `filter_map`. Sibling rows in the same batch are unaffected.
+
+### Rule 5: Parallel tokenization, serial persist
+
+Inside `record_file_ops`'s background thread: `rows.into_par_iter().filter_map(...)` resolves all counts in parallel (rayon), then `for rec in records { persist_record(&rec); }` writes serially. This is intentional: rayon is safe for read-only BPE tokenization; SQLite's WAL mode allows concurrent readers but a single writer is simpler and avoids contention errors.
```
dean0x: **[BLOCKING] Rule 5 misdescribes the persist loop — conflates the two recording paths its own Rule 3 forbids conflating**

Reviewer: documentation (95%).

The text says `for rec in records { persist_record(&rec); }` writes serially. The actual `record_file_ops` background thread (`analytics/mod.rs:1191–1199`) does the opposite:

```rust
if let Ok(db) = AnalyticsDb::open_default() {
    for rec in &records { let _ = db.record(rec); }
    if !records.is_empty() { db.maybe_prune(); }
}
```

One connection for the whole batch; `persist_record` is never called. `persist_record` is a **separate** per-row helper used only by the **subcommand path** — it opens a fresh connection and prunes on every call. The consequential claim at `:111` ("`maybe_prune()` runs after each `persist_record` call") is therefore wrong for the file-op path, where prune runs **once per batch**.

This is the KB's own Rule 3 violation ("Two recording paths — never conflate them") committed by the KB itself. An agent trusting the description would replicate the per-row-connection pattern in a new recorder, undoing a real optimization, or "discover" the batch design as a bug and revert it.

**Fix:** rewrite `:69` to match the code:

> "Inside `record_file_ops`'s background thread: `rows.into_par_iter().filter_map(...)` resolves all counts in parallel (rayon), then a **single** `AnalyticsDb::open_default()` connection writes every record serially via `db.record(rec)`, followed by one `db.maybe_prune()` for the whole batch. Note this does NOT go through `persist_record` — that helper opens a fresh connection per row and is used only by the subcommand path."

Also correct `:111` to scope the per-call prune claim to `persist_record`/the subcommand path.

---
*Posted by Claude Code*

.devflow/features/index.md
```diff
@@ -3,7 +3,7 @@
...
...ok_status_line, or adding test harness sandboxing (skim_sandboxed / skim_sandboxed_with_bin / hermetic_path). Keywords: hook install, binary pinning, SKIM_HOOK_BINARY, SKIM_HOOK_COMMIT, resolve_skim_binary, generate_hook_script, is_hook_script_current, uses_pinned_binary, pin_is_current, script_has_pinned_marker, AwarenessOnly, codex, wrapper symlinks, wrappers_blocks_fast_path, guidance_content, guidance_content_mdc, Command wrapping, PermissionsProtocol, confirm_grant, READ_ONLY_SUBCOMMANDS, hook_config_dir, modifiedArgs, seed tier, sidecar manifest, proxy feature gate, cfg-gated registry, routing guard, cli_proxy_gating, KNOWN_SUBCOMMANDS, META_SUBCOMMANDS, wrapper_targets, uses_dedicated_hook_file, hook_event_key, print_install_summary, print_dry_run_actions, print_dry_run_permissions, SKIM_JSON_NAME, SKIM_REWRITTEN_FROM, rewrite_origin, rewrite_transparency_marker, view_differs, transparency marker, origin tag, ScriptIntegrity, classify_script_integrity, verify_script_integrity, hook_status_line, HookFacts, NoManifest, Tampered, Verified, Unreadable, binary pin mismatch, skim_sandboxed, skim_sandboxed_with_bin, hermetic_path, ADR-004, ADR-005, ADR-006, ADR-008, PF-015, PF-017.
```
dean0x: **[BLOCKING] New `analytics` feature KB is not registered here — invisible to every agent**

Reviewer: documentation (98%).

`.devflow/features/analytics/KNOWLEDGE.md` was added in this PR, but no `analytics` entry was added to `index.md`. The index still has exactly 9 entries; `grep analytics .devflow/features/index.md` returns nothing.

`index.md` is the relevance-matching cache orchestrators consume to decide which KB to pass as `FEATURE_KNOWLEDGE`. A KB with no index row is real to a human browsing the directory and **nonexistent to every agent**. 154 lines of researched context deliver zero value.

**Fix:** add an alphabetically-first row (before `ast-index`) derived from the KB's frontmatter:

```markdown
- **analytics** — crates/rskim/src/analytics, crates/rskim/src — Use when adding new recording paths, changing cache-directory resolution, debugging silent analytics drops, writing tests that invoke skim, or tracing where session_id flows. Keywords: analytics, token savings, record_file_ops, flush_pending, register_thread, cache_root, SKIM_CACHE_DIR, SKIM_ANALYTICS_DB, SKIM_DISABLE_ANALYTICS, session_id, fire-and-forget.
```

---
*Posted by Claude Code*

crates/rskim/tests/cli_doctor.rs
```diff
@@ -141,6 +144,69 @@ fn test_doctor_exits_1_and_names_tamper_after_hook_modification() {
...
+        .success();
+
+    // Run doctor from the original binary: same version/commit, different path
+    // → Verified integrity, pin_is_current == false → exit 1 + "binary pin mismatch".
+    common::skim_sandboxed(home)
+        .arg("doctor")
+        .current_dir(home)
+        .env("PATH", hermetic_path())
+        .assert()
+        .failure() // exit 1 (pin drift)
```
dean0x: **[SHOULD FIX] `.failure()` does not verify exit code 1 — any non-zero exit keeps this test green**

Reviewer: testing (88%).

`assert_cmd`'s `.failure()` succeeds for **any** non-zero status: a panic (101), an arg-parse error, or a future exit-code change to 2 would all keep the test green while the documented contract ("Exit codes: `0` success / `1` general error", CLAUDE.md; "`skim doctor` exit 0 healthy / 1 on any drift — works as a CI pre-flight") silently regresses.

The pre-existing `test_doctor_exits_1_and_names_tamper_after_hook_modification` at `:143` has the same weakness — but this is a new line and the PR description explicitly claims exit-code coverage for the binary-pin-mismatch branch.

**Fix (one line):**

```rust
.assert()
.code(1)  // exit 1 (pin drift) — not merely any non-zero
.stdout(predicates::prelude::predicate::str::contains("binary pin mismatch"));
```

---
*Posted by Claude Code*

.devflow/features/hook-binary-pinning/KNOWLEDGE.md
```diff
@@ -35,13 +35,40 @@ v2.11.0 added the **permissions seeding subsystem** (`skim init --permissions`):
...
+
+**`DetectedState::hook_is_current()`** combines version match AND pinned format: `version matches && hook_uses_pinned_binary`. It does NOT check path — that is `pin_is_current()`'s job. The separation is intentional (see below).
+
+**`DetectedState::pin_is_current()` (new in PR #488)** compares the canonical binary path stored in `hook_binary_pin` against the result of `resolve_skim_binary()`. It is **deliberately separate from `hook_is_current()`** for two reasons: (1) `hook_is_current()` + commit checks are what skim doctor uses to derive the *cause* of a mismatch — folding pin into `hook_is_current()` would collapse every pin mismatch into the generic `[stale]` bucket; (2) doctor can display the `hook_binary_pin` field independently via `HookFacts.pin_is_current` (PF-015 display-without-gate pattern). Absent pin → `false`.
+
+**Fast-path condition in `run_install_single`** now requires all five conditions:
```
dean0x: **[BLOCKING] "all five conditions" describes a seven-condition block**

Reviewer: documentation (98%).

The prose says "now requires all five conditions" then immediately shows a code block with **seven**: `hook_installed`, `hook_is_current()`, `pin_is_current()`, `guidance_current`, `!permissions_blocked`, `!wrappers_blocked`, `manifest_present`. Seven is correct — it matches `install.rs:503–511` exactly. The State Transitions section at `:150` correctly says "all 7 fast-path conditions true."

A reader who trusts the prose count over the block will believe two conditions are optional. The two most likely to be dropped are precisely the ones this PR added (`pin_is_current` and `!wrappers_blocked`) — both load-bearing for the two-clone fix and the #478 fix respectively.

**Fix:** change "five" to "seven" at this line.

---
*Posted by Claude Code*

.devflow/features/hook-binary-pinning/KNOWLEDGE.md
```diff
@@ -35,13 +35,40 @@ v2.11.0 added the **permissions seeding subsystem** (`skim init --permissions`):
...
+
+### The Currency Predicates
+
+Three predicates determine script staleness. Together they form a hierarchy; the fast-path idempotence check requires all three to be true.
+
+**`uses_pinned_binary` (`init/state.rs`)** delegates to `script_has_pinned_marker` to check whether the installed script uses the F6 pinned format. This is the ONLY place the marker string is scanned; any change to the marker must go here.
+
+**`is_hook_script_current` (`init/install.rs`)** ANDs its own version-line check with the pinned-marker scan, and **also validates the binary pin path** by calling `parse_binary_pin_from_script()` and comparing to `resolve_skim_binary()`. If the pin path in the script differs from the current binary's canonical path (two-clone same-commit scenario), it returns `false` and triggers a rewrite. This three-part check means a script that is correct in version, format, and path will not be needlessly regenerated.
+
+**`DetectedState::hook_is_current()`** combines version match AND pinned format: `version matches && hook_uses_pinned_binary`. It does NOT check path — that is `pin_is_current()`'s job. The separation is intentional (see below).
```
dean0x: **[SHOULD FIX] `hook_is_current()` description omits the commit check — breaks the KB's own central argument**

Reviewer: documentation (95%).

This line states `hook_is_current()` combines "version match AND pinned format: `version matches && hook_uses_pinned_binary`." The actual predicate (`state.rs:88–113`) is version + pinned format **+ commit** (`hook_commit` vs `option_env!("SKIM_GIT_COMMIT")`), and its own rustdoc says so. The same truncated definition repeats at `:150` and `:252`.

This is not a harmless omission — it breaks the KB's own central argument. Lines `:99` and `:230` justify deleting the `"stale"` terminal with "`commit_ok ∧ version_ok ⇒ hook_is_current`." That implication holds **only because `hook_is_current` checks commit**. Under this line's stated definition the implication does not hold, so a future agent auditing the dead-code removal derives a contradiction from the KB alone and may "restore" the terminal this PR correctly proved dead. The KB also contradicts the new `pin_is_current` rustdoc added in this same diff (`state.rs:47`: "`hook_is_current` checks version + pinned format + commit").

**Fix:** at `:56`, `:150`, and `:252` change to "version + pinned format + commit" and keep the "does NOT check path — that is `pin_is_current()`'s job" clarification (that part is correct and is the actual reason `pin_is_current()` exists).

---
*Posted by Claude Code*

crates/rskim-core/tests/ruby_transform.rs
```diff
@@ -153,11 +156,15 @@ fn test_ruby_minimal_strips_comments() {
```
dean0x, line 156: **[SHOULD FIX] Assertion inversions deleted the negative-case coverage of the blank-line-break rule for Ruby, SQL, and Bash**

Reviewer: regression (88%).

`is_module_header_comment` terminates at exactly two boundaries: a blank-line break and a non-comment preceding sibling. If either boundary silently stopped working (e.g. a tree-sitter grammar bump changed a comment node to include its trailing newline), the whole test suite would stay green while `skim --mode=minimal` quietly stopped compressing those languages.

After this PR's inversions, **only Python** has a live guard on the negative case. Ruby, SQL, and Bash do not:
- **Ruby:** this test moved its strippable comment from root level **into a class body**, so it now exercises `is_inside_function_body`/`is_root_child` — a different code path. There is no Ruby test asserting a root-level comment after a blank-line break is stripped.
- **SQL:** `test_sql_minimal_strips_comments` was inverted to assert preservation. Its replacement (`test_sql_minimal_reduces_tokens`) asserts only `result.len() < source.len()` — and the comments in `sql/comments.sql` are stripped by the preceding-code rule, so this assertion passes even if the blank-line break stopped working entirely for SQL.
- **Bash:** `bash_transform.rs:290` only asserts `result.is_ok()`. Zero behavioural coverage. (Separately flagged by testing reviewer at 90%.)

**Fix:** add one negative-case assertion per affected language. For Ruby:

```rust
// test that a root-level comment AFTER a blank line is still stripped
let src = "# module header\n\n# strippable\ndef foo; end\n";
let result = transform(src, Language::Ruby, Mode::Minimal).unwrap();
assert!(result.contains("# module header"), "header preserved");
assert!(!result.contains("# strippable"), "post-blank comment stripped");
```

Mirror the same pattern for SQL (`-- header / blank / -- strippable / SELECT 1;`) and Bash (`#!/usr/bin/env bash / # header / blank / # strippable / foo()`).

---
*Posted by Claude Code*

crates/rskim/src/cmd/init/helpers.rs
```diff
@@ -11,6 +11,28 @@ use std::path::{Path, PathBuf};
...
+/// any symlinks. This is the single source of truth for the binary path used
+/// in hook scripts (`SKIM_HOOK_BINARY`) and wrapper installation so that all
+/// three sites agree — avoiding the two-clone mismatch where both clones share
+/// the same version string but `hook_is_current()` cannot detect they are
+/// different binaries.
+///
+/// Canonicalize failure (e.g., binary deleted while running) falls back to the
+/// raw path from `current_exe()` rather than failing, preserving the behaviour
+/// of the pre-unification code at `install.rs:895-906`.
+pub(super) fn resolve_skim_binary() -> anyhow::Result<PathBuf> {
```
dean0x: **[SHOULD FIX] `resolve_skim_binary()` — the PR's central unification — has zero direct tests and no symlinked-binary coverage**

Reviewer: testing (90%).

This function is now the single source of truth for the canonical binary path consumed by four sites (`install.rs:747`, `:869`, `:938`, `state.rs:65`, `:121`). `helpers.rs` has a `mod tests` block but adds no test for it.

More critically: **nothing tests the symlink-resolution behaviour that is the entire reason the helper exists.** The feature KB states explicitly: "A test environment where the binary has no symlinks will pass even if the three-site invariant is broken — the failure only appears on symlinked-path machines." `test_pin_is_current_matching_path_returns_true` (`state.rs:1050`) is a tautology on non-symlinked paths — it derives its expected value with the same `canonicalize(current_exe())` algorithm the implementation uses.

**Fix:** add a Unix-gated E2E that installs **through a symlink** and asserts the hook pins the resolved canonical target, not the symlink:

```rust
#[cfg(unix)]
#[test]
fn test_init_pins_canonical_path_when_invoked_via_symlink() {
    let tmp = TempDir::new().unwrap();
    let home = tmp.path();
    std::fs::create_dir_all(home.join(".claude")).unwrap();

    let real_bin = common::skim_bin();
    let link = home.join("skim-link");
    std::os::unix::fs::symlink(&real_bin, &link).unwrap();

    common::skim_sandboxed_with_bin(home, &link)
        .args(["init", "--agent", "claude-code", "--no-guidance", "--no-wrappers"])
        .assert()
        .success();

    let script = std::fs::read_to_string(home.join(".claude/hooks/skim-rewrite.sh")).unwrap();
    let canonical = std::fs::canonicalize(&real_bin).unwrap();
    assert!(
        script.contains(&*canonical.to_string_lossy()),
        "hook must pin the canonical target, not the symlink:\n{script}"
    );
    assert!(!script.contains("skim-link"), "hook must not pin the symlink path:\n{script}");
}
```

A second re-run asserting "Already up to date" would additionally lock in the no-reinstall-loop property.

---
*Posted by Claude Code*

dean0x:
## Review Summary — PR #488

Inline comments above cover the highest-confidence findings (>=80%). This comment consolidates everything that is either lower-confidence (60–79%) or whose anchor line is not in the diff.

---

### Blocking items not coverable as inline comments

**`crates/rskim-core/src/types.rs:567, :642` — `Mode::Minimal` rustdoc contradicts the new behaviour on a published crate** (documentation 92%, consistency 88%)
`types.rs` is not in this diff, but `docs/modes.md` was updated while the in-code API docs were not. Both sites still say Minimal mode strips "non-doc comments" with no header-comment exception. `rskim-core` is published to crates.io, so this renders on docs.rs as the authoritative contract. Fix: `/// Minimal cleanup - strip non-doc comments (module header comments preserved for Python/Ruby/SQL/Bash), normalize blank lines`.

**`CLAUDE.md:83` — doctor's pin state described as SHA-only; it is now also path-based** (documentation 88%)
The bullet says doctor reports "hook pin state per agent (**pinned SHA vs running SHA**)". This PR adds a path comparison and the new "binary pin mismatch" terminal reports paths, not SHAs. An agent debugging a `binary pin mismatch` line will consult CLAUDE.md, read that pin state is a SHA comparison, and look for a commit divergence that by construction does not exist (this terminal is only reachable when version AND commit both match). Fix: "hook pin state per agent (pinned commit SHA **and** pinned absolute binary path vs the running binary)".

---

### Should-fix items not coverable as inline comments

**`crates/rskim/src/cmd/doctor/mod.rs:371-372` — `hook_status_line` docblock still says `Unreadable` names the suppression coupling** (consistency 95%, documentation 90%)
The docblock says `Unreadable → drift (✗), names the suppression coupling`. This PR removed that claim from the branch body (`mod.rs:412-418`) and added a test at `mod.rs:1002-1008` that *forbids* the string "silences drift detection". The docblock 40 lines above contradicts both the code and the test — the exact false claim fix #479 was written to eliminate survives here. Fix: update `:371-372` to: `Unreadable → drift (✗). Does NOT name the suppression coupling — Unreadable maps to integrity_failed = false, so drift detection still runs at hook-exec time.`

**`crates/rskim/tests/cli_init.rs:1786` — fast-path-bypass test has no same-harness control** (testing 82%)
`test_init_wrappers_bypasses_fast_path` asserts only the *absence* of "Already up to date", without first proving the fast path was reachable in this harness. Fix: add a control run between steps 1 and 2 asserting `stdout.contains("Already up to date")`.

**`crates/rskim/src/cmd/init/install.rs:778-789` — `print_wrapper_install_result` PATH-blurb gating is untested** (testing 88%)
New gating on `result.created + result.updated > 0` is a user-visible behaviour change with no test on either side. Fix: add a step 3 to `test_init_wrappers_bypasses_fast_path` running `--wrappers` again and asserting the blurb does NOT appear.

**`crates/rskim/src/cmd/init/install.rs:169` doc comment** (consistency 88%)
The one-liner "Rule (mirrors `permissions_blocks_fast_path`)" overstates the similarity — arm ordering is reversed and the `None` arm deliberately diverges. Fix the doc comment to call out the divergence explicitly.

---

### Lower-confidence suggestions (60-79%)

- **`doctor/mod.rs:488`** (consistency 90%) — label `running:` breaks the `binary:` vocabulary used in the two adjacent reason strings. Pick one label for the running-binary slot across all three reasons.
- **`doctor/mod.rs:500`** (complexity 85%) — `let pin = facts.hook_binary_pin.as_deref().unwrap_or("?")` computed twice (at `:460` and `:500`). Hoist above the branch.
- **`state.rs:1054`** (testing 85%) — `let Some(running_path) = running else { return; }` is a silent green pass when `current_exe()` fails. Use `.expect("current_exe() must resolve in the test environment")`.
- **`install.rs:503`** (consistency 88%) — `skim init --dry-run` on a fully-current install exits the fast path before the `if flags.dry_run` block, previewing nothing. PF-018 listed hoisting dry-run as one of three items; two of three landed.
- **`helpers.rs:24-25`** (complexity 90%) — tombstone comment cites `install.rs:895-906` which this commit replaced. Drop the archaeological pointer; keep the behavioural contract (canonicalize failure falls back to raw path).
- **`minimal.rs:300`** (consistency 65%, rust 70%) — `node.parent().map(|p| p.parent().is_none()).unwrap_or(false)` → `node.parent().is_some_and(|p| p.parent().is_none())`.
- **`minimal.rs:495`** (rust 75%, testing 70%) — `nth_root_comment(tree, _source, n)` never reads `_source`. Drop the unused parameter.
- **`hook.rs:86`** (architecture 85% — pre-existing) — `current_exe().and_then(canonicalize).ok()` discards the raw path on canonicalize failure, diverging from `resolve_skim_binary()`. Not in this diff; worth folding into the path-unification follow-up.
- **`docs/modes.md:10`** (regression 85%, documentation 85%) — "What's Removed" cell reads `Non-doc comments (except headers)` without qualifying language scope. C/Rust/Go/TS licence headers are still stripped. Fix: `Non-doc comments (Python/Ruby/SQL/Bash module headers preserved)`.
- **`analytics/KNOWLEDGE.md` Anti-Patterns section** (compliance 88%) — no entry for credential scrubbing, though `original_cmd` persists raw command text. Code is correct; KB omits the invariant. Fix: add one Anti-Pattern bullet pointing to `build_analytics_label` in `cmd/git/mod.rs` as the reference pattern.

---

### Stats

| Category | Count |
|----------|-------|
| Inline comments posted | 14 |
| Skipped (already existed) | 0 |
| Consolidated into this summary | 12 |
| Lines not in diff (types.rs:567, CLAUDE.md:83, doctor:371) | 3 |
| Failed to post | 0 |

---
*Posted by Claude Code*

PR #351, opened 2026-06-22
feat(rskim-proxy): HTTP proxy foundation — TransformStage seam + identity transform (#303)

dean0x:
## Summary

- Implements the full `rskim-proxy` crate: HTTP/1.1 reverse proxy with a `TransformStage` seam, identity transform, provider detection, auth classification, analytics, health endpoints, and graceful shutdown
- Adds the conformance harness (AC19a/AC19b), statelessness/determinism property test (AC11), criterion latency bench (AC14/AC15), and a scoped `windows-latest` CI job (AC17)
- Gate-2 alignment: streaming passthrough (AC5/AC7), per-stage `catch_unwind` (AC9), `init_logging()` wiring (AC13), wire integration tests for 17 ACs, AC25 dispatch assertion

### Phase 1 — Crate skeleton, config, seam, CLI wiring
- `crates/rskim-proxy/` — new optional workspace crate (hyper/tokio/rustls, no system-CA, webpki-roots; `publish = false`; deny `unwrap`/`expect`/`panic` via `[lints.clippy]`)
- `ProxyConfig` + `ProxyConfigBuilder` with evidence-based constants: `DEFAULT_PROXY_PORT = 41322`, port range 41000-49000 (D8/AD-PXY-03, not Windows dynamic range), `DEFAULT_MAX_BODY_BYTES = 64 MiB`
- `TransformStage` trait + `IdentityStage` + `IdentityStageContractAdapter` + `TransformPipeline` (ordered composition, AD-PXY-06)
- `ProxyProvider` + `detect_provider` (path-suffix → shallow JSON shape → Unknown; 8 KiB sniff limit)
- `AuthMode` + `classify_auth` (header shape only; value never stored/logged — AC13)
- `AnalyticsHook` trait + `ChannelAnalyticsHook` (bounded crossbeam, lossy on overflow — AC15)
- CLI `skim proxy` wiring (`cmd/proxy.rs`; CLEARTEXT_WARNING for non-loopback bind)

### Phase 2 — Provider detection corpus + seam integration tests
- Fixture corpus: 16 provider/shape combinations covering Anthropic/OpenAI/unknown + SSE/JSON/empty
- Seam integration tests: AC2 (identity passthrough), AC4 (never-inflate guardrail), AC8 (malformed JSON byte-identity), AC9 (panicking stage catch_unwind)
- Per-stage `catch_unwind`: each `stage.apply()` individually wrapped — panicking stage falls open to pre-stage bytes, remaining stages still execute

[...]

Issue #306, opened 2026-06-09
L3: KV-cache alignment — stable prefixes, deterministic ordering, cache_control breakpoints


dean0x:
`Wave: 3 · Depends on: #302, #303 · Part of: #309`

## Summary

Make every request that flows through `skim proxy` byte-deterministic in its cacheable prefix, so the provider-side prompt cache actually hits across turns. Anthropic prompt caching is an exact-bytes prefix match (render order: `tools` → `system` → `messages`); cached reads cost ~0.1x the base input price — roughly a 90% discount — which makes this the largest per-token price lever in the Layer-3 slate, complementary to the stale-compaction/CCR savings pools (#307, #308), which deliver the most user-visible savings. The work is pure arithmetic, not heuristics: deterministically sort tool arrays and JSON-Schema keys (metadata only — never semantically visible content), keep the system prefix byte-stable, and inject up to 4 `cache_control` breakpoints at deterministic, structurally defined positions (v1 policy below: tools/system zone only; never inside `messages`). The deeper win is determinism itself: agent clients that already place their own breakpoints get hits they were losing to unstable serialization, and clients that place none get the tools+system prefix cached for free. Users get a materially smaller bill with zero risk to message content: the only bytes that may change are tool ordering, schema key ordering, and injected cache metadata. Trust is the co-equal deliverable — a parse failure, an unsupported body shape, or any internal error forwards the original request byte-for-byte.

## Why now — evidence

**The proxy paradigm does not pay for itself through content compression alone.** The independent benchmark PRISM #645 (publishing ~2026-06-17) measured the message-boundary-proxy paradigm at ~10.2% average savings costing +0.9s/turn (PRISM's own vendor-claimed fleet median: 4.8%). Cache economics — the ~90% cached-read discount multiplied across every turn of an agentic session — is where the proxy's per-token cost story lives. That is why this ticket is Wave 3's priority despite being "just sorting and markers."

[...]

Issue #305, opened 2026-06-09
L3: Proxy observability — analytics schema v4, per-provider/model savings accounting


dean0x:
Wave: 3 · Depends on: #300, #303, #304 · Part of: #309

## Summary

Extend skim's existing analytics pipeline (`~/.cache/skim/analytics.db`, SQLite WAL, fire-and-forget recording) so every request compressed by the Layer-3 proxy is recorded with provider, model, and turn attribution, and surface per-provider/per-model savings breakdowns in `skim stats`. Savings are always real measured token deltas — `raw_tokens − compressed_tokens`, both counted with the provider-appropriate tokenizer from #300 — never estimates ("measured" is defined precisely in design constraint 4). This is the trust layer of the L3 slate: users justify adopting a compression proxy with its savings numbers, and competitor history shows that opaque or estimated accounting (mixed scopes, ledger bugs, vendor-claimed figures that diverge from independent measurement) destroys that trust. It also serves as the regression tripwire: per-provider breakdowns are how a silent per-provider compression failure gets noticed at all. Recording must add zero latency to the request forwarding path (defined in constraint 1) and must never affect a forwarded request.

## Why now — evidence

PRISM's savings-accounting record (all from the 2026-06-09 competitive study) is a sustained bug-and-trust problem:

- **PRISM #83** (closed 2026-03-31): dashboard total savings were non-monotonic and mixed three different scopes (shell-output rewriting via the bundled GRANITE binary, cache savings, compression savings) into one number. The trust problem recurred months later (further dashboard/stats corrections through June 2026, including a hero-tile rework).
- **PRISM #491** (closed 2026-05-25): the stats `compression_cache.total_tokens_saved` counter was always 0 due to a key-name bug — savings were derived from a cache ledger rather than measured directly, so a bookkeeping bug silently zeroed the headline metric.
- **PRISM #454** (closed 2026-05-12): compression on Azure/OpenAI backend traffic regressed to near-zero (from 30–40%) **silently** — a per-provider loss of the product's entire value that no aggregate metric exposed. This is the strongest single argument for per-provider breakdowns.
[...]

PR #504, opened 2026-08-25
fix(#306): close out KV-cache alignment — AC7 duplication removed, detector hardened, CI reach extended

dean0x:
Closes out #306 for merge into `wave/l3-wave3`. Six commits on top of the prior 29.

## What this does

**1. AC7 passes by removal of duplication, not by exemption.** The `ac7_model_encoding_mapping_is_sole_in_workspace` guard (#300) was failing on #305's `select_encoding` — a delegating wrapper holding no model table, flagged purely because its signature contained `&str` and returned `Encoding`. Rather than exempt it, the provider family defaults (`AnthropicOffline`, `O200k`) were collapsed into `rskim-tokens` next to the model table, as `Provider` + `encoding_for_provider_model`. `select_encoding` is deleted; call sites in `analytics/mod.rs` (×2) and `proxy_analytics.rs` route through `impl From<RecordingProvider> for rskim_tokens::Provider`.

All nine (provider × model) cells were verified equivalent to the deleted function by direct comparison against `git show 8b6b3eb5:crates/rskim/src/analytics/mod.rs`. The `Provider::Unknown` early return happens *before* the model string is read — a "compute default then consult model" ordering would return `Cl100k` for `(Unknown, Some("gpt-4"))` and break an AC21 cell.

**2. The AC7 detector now detects constructs, not signatures.** The old rule read function headers and never bodies, making it simultaneously over-inclusive (the delegating wrapper above) and under-inclusive (owned params, module-level tables). It is replaced by three body-based signals — literal match arm, `&str`-keyed table type, prefix-family fn body — each reporting *which* construct fired, behind a comment/string-literal sanitizer that preserves byte offsets.

This also removed the blanket `integration.rs`-by-name exclusion that had been exempting all five `integration.rs` files in the workspace from a workspace-wide invariant. Only `crates/rskim-tokens/src/encoding.rs` remains excluded.

Review then found five false-negative classes the first pass missed, one of them live in this repo (`use Encoding::*;` + bare variants — that exact import exists in the two files most likely to grow a parallel table). Closed: glob imports, `==` equality ladders, alias imports (`use Encoding as Enc`, `type Enc = Encoding`), and `Self::` inside `impl Encoding`. Cow-keyed pair slices remain a documented accepted gap. An anti-rot test (`ac7_encoding_variant_list_is_complete`) fails in both directions if the variant list drifts from `Encoding`.

**3. Stale documentation corrected.** `8b6b3eb5` re-baselined the AC17 gate 5.0 ms → 9.0 ms but missed two `align_bench.rs` assertions and a CHANGELOG work-note quoting superseded 2.5 ms / 5 ms / pre-R6 7.09 ms figures. Review additionally caught that the bench header claimed CI asserts **p99** across **three groups** — CI's own comment says it uses the 95% CI of the *mean* ("NOT a true p99") and gates exactly one group. Also corrected: a rustdoc claim that CI forbids `rskim-tokens` depending on `rskim-contract`/`rskim-llm` (the isolation step only greps for HTTP/TLS crates, so it would not).

**4. CI now reaches this code.** Triggers were `push: [main, feature/*]` and `pull_request: [main]`, so no `wave/*` or `ticket/*` push fired and a PR based on `wave/l3-wave3` did not either. **This 11k-line wave, including a brand-new crate, has never had a single CI run.** Combined with mandatory per-crate local scoping, a cross-crate invariant was structurally invisible for the life of the wave — which is why the AC7 violation sat. Push now covers `wave/**` and `ticket/**`, and `pull_request` is unfiltered. A `concurrency` block was added so a PR branch push does not run the whole matrix twice.

## AC17 gate amendment

The issue body still states `< 5 ms` (in AC17 and again in constraint 12). That figure was an unmeasured halving of #303's proxy bar. The shipped gate is **9.0 ms**, derived in `ADR-CA-001` § "AC17 gate history" from three measured runs: worst post-R6 upper bound 6.9696 ms × 1.25 headroom = 8.712 → 9.0 ms, against a 4.03% noise floor and inside #303's 10 ms ceiling. Amending the issue rather than shipping against an unmet stated criterion, per ADR-003.

## Verification

Full workspace, run locally with the machine idle:

| Gate | Result |
|---|---|
| `cargo fmt -- --check` | exit 0 |
| `cargo clippy --all-features --all-targets -- -D warnings` | exit 0 |
| `cargo test --all-features --no-fail-fast` | all targets green |
| `cargo test --doc` | `9 passed` |
| AC7 (post-widening, whole-workspace walk) | `7 tests run: 7 passed` |
| AC21 nine cells (both crates) | pass |
| AC19 — #301 conformance harness | `197 passed` / `4 passed` / `10 passed` |
| AC19 — align conformance bridge | `8 passed` |

`test_skim_test_cargo_in_this_repo` fails under a cold cargo cache: it wraps a real nested `cargo test` in a hard 120s timeout (`cli_test_cargo.rs:22`) and the compile alone exceeds it. Warm, it passes at 85.57s. Worth watching on CI, which starts cold.

The AC7 guard's non-vacuity was proven by production mutation (PF-014), not by reading: injecting a real duplicate table into `analytics/mod.rs` made the gate fail naming the file *and* the construct; restoring left the tree clean.

## Not in this PR

- **AC20 live validation** — needs a real Anthropic key; gated at the `wave/l3-wave3` → `main` merge, not this one.
- Stale open PRs #329 / #351 target `main` and their content already reached the wave via squash merges `7b60be5f` (#335) and `d642c914` (#354).

## Incidental cleanup

`.devflow/docs/handoff-ticket-305-proxy-observability.md` was committed during #305 (the devflow implement workflow writes a handoff file for sequential Code agents and is meant to delete it on completion; this one survived). ADR-009 tracks only `.devflow/features/index.md` and `.devflow/features/*/KNOWLEDGE.md` — removed here so it does not reach `main` with the wave.


dean0x:
Merged into `wave/l3-wave3` as merge commit 650f2909 (two parents: bc825fd1 + 088a57b3).

Performed as a pushed `git merge --no-ff` rather than via the PR button — this repository is configured squash-only, and a squash would have flattened the branch history. This matches how `bc825fd1 merge(wave/l3-wave3): integrate #305` was created. GitHub therefore does not auto-detect the merge, hence closing manually.

CI was green on this PR at merge time (7/7). AC20 remains an open gate on the `wave/l3-wave3` → `main` merge, not on this one.

PR #508, opened 2026-08-27
fix: skim fidelity & parity overhaul — guard baseline, transparency, surface parity, render + literal correctness

### LINKED ISSUE #510: bug: skim git diff --json drops the rendered diff and bypasses the ADR-001 guardrail (#317 disclosure gap)
## Problem

`skim git diff --json` emits only file metadata. The rendered diff — hunks, markers, line
numbers — is never present in the JSON envelope, in any mode (default, structure, full).

`DiffResult` (`crates/rskim/src/output/canonical.rs:901-909`):

```rust
/// Complete diff result with file entries and pre-rendered display
#[derive(Debug, Clone, Serialize, Deserialize)]
pub(crate) struct DiffResult {
    #[serde(default)]
    pub(crate) files_changed: usize,
    pub(crate) files: Vec<DiffFileEntry>,
    #[serde(default, skip_serializing)]
    rendered: String,
}
```

`skip_serializing` on `rendered` means the JSON output is
`{files_changed, files:[{path, status, changed_regions}]}` and nothing else.

### 1. The render work is performed, then discarded

`render_and_format` (`crates/rskim/src/cmd/git/diff/mod.rs:184-249`) renders every file —
in parallel above `PARALLEL_THRESHOLD` — builds `DiffResult::new(diff_file_entries,
rendered_output)` at `:230`, and then throws the render away:

```rust
let result = DiffResult::new(diff_file_entries, rendered_output);

let out = match output_format {
    OutputFormat::Json => {
        let json = serde_json::to_string_pretty(&result)   // :234 — drops `rendered`
            .map_err(|e| anyhow::anyhow!("failed to serialize diff result: {e}"))?;
        let status = exec::write_line_to_stdout(&json)?;
        return Ok((json, status));                          // :237
    }
```

Every `--json` invocation pays a full parse + AST render per file for output that is
discarded at the serializer. This is pure waste on the JSON path.

### 2. The JSON arm bypasses the ADR-001 guardrail entirely — this is a #317 disclosure gap

The guardrail runs only on the text arm (`mod.rs:394-413`):

```rust
let (final_str, tier) = match output_format {
    OutputFormat::Text => {
        let outcome = crate::output::guardrail::apply_to_stderr(raw_diff, result_str)?;  // :399
        ...
    }
    OutputFormat::Json => {
        // JSON was already printed inside render_and_format; raw_diff is no
        // longer needed.  Analytics uses raw_for_record (clone taken above).
        drop(raw_diff);                                                                  // :417
        (result_str, "full")
    }
};
```

`render_and_format` already returned at `:237`, so `apply_to_stderr` never sees the JSON
body. The consequence matters more than the wasted work:

**A summary-only JSON view is strictly less than raw, with no class-1 marker.** Per the
`compress, never truncate` constraint (#317) and the ADR-011 stderr taxonomy, a view that
shows the reader less than the raw tool is loss-bearing and must carry an unconditional
elision marker. `--json` shows dramatically less than `git diff` and says nothing — and the
one mechanism that would have caught the shortfall (the net-savings guard) is structurally
unreachable on this path. Note the analytics tier is recorded as `"full"` at `:418`, which
is the opposite of what the user receives.

This reframes the issue: it is not merely "the docs don't say `--json` is summary-only", it
is a byte-fidelity disclosure gap on a documented output surface.

## Suggested resolution

Not mutually exclusive; (a) is required either way.

**(a) Disclose the shortfall.** Whatever the envelope ends up carrying, `--json` must not
silently serve less than raw. Either emit a class-1 marker/field noting the render is
omitted and how to get it (`SKIM_PASSTHROUGH=1`, or the text mode), or route the JSON body
through the same fidelity gate the text path uses.

**(b) Carry the render.** Re-enable serialization for `DiffResult::rendered` — as
`rendered: Option<String>` behind a flag (`--json-full`), or unconditionally — so the
envelope is a superset of the summary rather than a strict subset of raw.

**(c) Skip the work when it is discarded.** If the envelope stays summary-only, do not
render at all on the JSON path; build `DiffFileEntry` values directly. This removes the
wasted parse + AST per file.

## Severity

**P2** — a documented output surface serves strictly less than the raw tool with no
disclosure, and the guardrail that exists to catch exactly that cannot run on it.

## Affected files

- `crates/rskim/src/output/canonical.rs:901-909` — `DiffResult::rendered`, `skip_serializing`
- `crates/rskim/src/cmd/git/diff/mod.rs:230-237` — render built, then discarded by the serializer
- `crates/rskim/src/cmd/git/diff/mod.rs:394-419` — guardrail on the text arm only; JSON arm bypasses it

---

**PR**: #508 (branch `fix/fidelity-parity-overhaul`, 25 commits, HEAD `b6fc565`; `main` untouched)
**Related**: #505 (PR #488 consolidation hub) tracks the sibling guardrail work under #476.

### LINKED ISSUE #511: transform: truncate.rs cut is not literal-boundary aware — elision marker can land inside an unterminated literal
## Problem

`crates/rskim-core/src/transform/truncate.rs` cuts by counting source lines and never
consults literal boundaries. When a multi-line string, template literal, or heredoc spans
the cut point, the output ends mid-literal and the elision marker is appended **after** an
unterminated literal — so the marker itself lands inside what a downstream parser reads as
an open string.

Example: `--mode full --max-lines 4` on a TypeScript file whose 4th line falls inside a
template literal yields an unterminated backtick string followed by
`// ... (N lines truncated) — SKIM_PASSTHROUGH=1 for full output`, where the `//` comment is
not a comment at all — it is literal content.

**That mis-scoped marker is the entire issue.** The cut is not literal-boundary aware.

## Corrections to the original report — the marker machinery is present and working

An earlier draft of this claimed the truncation happens "with no marker", "no `...` hint",
and "no `SKIM_PASSTHROUGH=1` suggestion". **All three are false at HEAD**, and the issue
should not be read as a missing-marker report:

- `append_hint` (`truncate.rs:12-23`) applies to every marker this module emits.
- `truncate_to_lines` builds `// ... (truncated) — <hint>` at `:204` and pushes it at
  `:211`, `:220`, and `:239`.
- `simple_line_truncate` (`:260-292`) always appends
  `// ... (K lines truncated) — <hint>` at `:277-284`.
- The CLI wires the hint unconditionally — `crates/rskim/src/cascade.rs:48-50`:
  ```rust
  let mut config = TransformConfig::with_mode(mode)
      .with_line_numbers(line_numbers)
      .with_elision_hint("SKIM_PASSTHROUGH=1 for full output");
  ```

The ADR-011 class-1 remedy clause is therefore already satisfied in content. What is wrong
is *where the cut lands*, which can put that correct marker in a syntactically invalid
position.

## Real exposure is narrow

The `simple_line_truncate` path is reached for **passthrough** transforms —
`crates/rskim-core/src/types.rs:457-467`:

```rust
// Apply max_lines truncation for passthrough paths. Tree-sitter languages
// handle max_lines inside transform_tree via AST-aware priority selection,
// but for passthrough (Full mode or serde/Markdown Minimal/Pseudo) we must
// apply it here as a simple line truncation (B5: pass elision_hint).
if let Some(max_lines) = config.max_lines {
    let truncated = crate::transform::truncate::simple_line_truncate(...)
```

So the exposed combinations are:

- `--mode full --max-lines N` (any language)
- serde-backed formats (JSON/YAML/TOML) and Markdown in Minimal/Pseudo

Structure mode has already stripped bodies before truncation runs, so it is largely
unaffected — most multi-line literals are gone by then. `truncate_to_lines`' AST-aware
priority selection also cuts on node spans, which makes a mid-literal landing much less
likely there, though it is not a guarantee.

## Suggested fix

Consult `collect_literal_ranges` (`crates/rskim-core/src/transform/literals.rs:130`) and
either:

- extend the cut to the end of the enclosing literal, or
- pull the cut back to the literal's start, and account for the elided literal bytes in the
  existing marker (the count is already part of the marker text).

Either way the marker must end up outside literal scope.

## Prior art on this branch

- `1ad93a8` (literal-aware pseudo/minimal passes) deferred this explicitly. From its commit
  message: *"Follow-up: truncate.rs mid-literal-cut (out of scope, filed separately)."*
  This is that filing.
- `3856671` (`fix(E1-E4): pseudo param annotations, gh field drops, truncation off-by-one`)
  did touch `truncate.rs`, but for the **E4.2 off-by-one only** — `:274-276` and `:315-317`
  changed the marker from consuming one of the N requested lines to being line N+1. It did
  not add literal awareness and does not close this.

## Severity

**P3** — narrow trigger surface, and the output is a *view*, not something skim writes back.
The reader can see the truncation happened; the failure is that the output no longer parses
and the marker is mis-scoped.

## Affected files

- `crates/rskim-core/src/transform/truncate.rs` — `simple_line_truncate` (`:260-292`),
  `truncate_to_lines` marker sites (`:204`, `:211`, `:220`, `:239`)
- `crates/rskim-core/src/transform/literals.rs:130` — `collect_literal_ranges` (the fix input)
- `crates/rskim-core/src/types.rs:457-467` — passthrough routing into `simple_line_truncate`

---

**PR**: #508 (branch `fix/fidelity-parity-overhaul`, 25 commits, HEAD `b6fc565`; `main` untouched)

### LINKED ISSUE #513: tests: seven scaling guards assert a single-sample wall-clock ratio — use minimum of k repetitions
## Problem

Seven tests in `rskim-core` assert a wall-clock ratio `t2 / t1` from **one timing sample per
size point**. A single descheduling event on the `t2` run — and nothing else — flips them
red. The tests can fail with no O(N²) regression present.

## Full inventory (the original report named 2; there are 7)

All share the identical shape. Filing them together, because fixing a subset leaves the
flake surface open.

| Test | Location | Ratio limit | Noise floor | Measured (per doc comment) |
|---|---|---|---|---|
| `test_typescript_pseudo_semicolon_walk_scaling_guard` | `transform/pseudo.rs:2741` | `< 2.8` | `t1 >= 8.0` | 2.07× |
| `test_python_pseudo_return_type_walk_scaling_guard` | `transform/pseudo.rs:2778` | `< 2.8` | `t1 >= 6.0` | 2.04× |
| `test_cpp_pseudo_template_walk_scaling_guard` | `transform/pseudo.rs:2810` | `< 2.8` | `t1 >= 5.0` | 2.01× |
| `test_quadratic_scaling_guard` | `transform/minimal.rs:1424` | `< 2.5` | — | — |
| `test_go_leading_comment_run_scaling_guard` | `transform/minimal.rs:2062` | `< 2.8` | `t1 >= 1.5` | — |
| `test_go_doc_block_scaling_guard` | `transform/minimal.rs:2090` | `< 2.8` | `t1 >= 3.0` | — |
| `test_go_pseudo_leading_comment_run_scaling_guard` | `transform/minimal.rs:2116` | `< 2.8` | `t1 >= 1.5` | 2.02× |

(All paths under `crates/rskim-core/src/`. Note `test_quadratic_scaling_guard` uses a
tighter `2.5` limit, so it has *less* headroom than the rest.)

## Root cause

`time_pseudo` (`crates/rskim-core/src/transform/pseudo.rs:2697-2706`) takes exactly one
`Instant::now()` sample and returns it:

```rust
fn time_pseudo(source: &str, language: Language) -> f64 {
    let mut parser = Parser::new(language).unwrap();
    let tree = parser.parse(source).unwrap();
    let config = TransformConfig::with_mode(Mode::Pseudo);
    let start = std::time::Instant::now();
    let r = transform_pseudo_with_spans_and_line_map(source, &tree, language, &config);
    let ms = start.elapsed().as_secs_f64() * 1000.0;
    assert!(r.is_ok(), "pseudo transform must succeed: {:?}", r.err());
    ms
}
```

The `minimal.rs` guards inline the same single-sample pattern (`:1458`, `:1468`, `:1998`,
`:2009`).

With one sample, scheduler noise is *additive and one-sided* — it can only make a sample
slower. Whether the test passes depends on which of the two runs got unlucky.

## Headroom is thinner than it looks

The guards' own `THRESHOLD RATIONALE` note (`pseudo.rs:2693-2695`) says:

> THRESHOLD RATIONALE — 2.8 ≈ 2^1.5 is the exponent-space midpoint between
> linear (2.0×) and quadratic (4.0×), matching the Go guards in minimal.rs.
> These shapes measure 2.01×–2.07× at the guard sizes, so 2.8 leaves ~35 % headroom.

That is the right *threshold* reasoning — 2.8 genuinely sits between linear and quadratic.
The problem is not the threshold, it is that a single sample of `t2` needs only to stretch
~35–39% relative to `t1` to cross it, and `t1` values of 1.5–8 ms are small enough that one
preemption does that.

## Honest note on reproduction

**The "fails intermittently" claim from a prior phase did NOT reproduce.** 10 runs on this
branch — 5 isolated, 5 under full-suite parallel load — all passed. This issue is filed on
the *shape* of the assertion, not on an observed failure. The shape is wrong regardless: a
test that can fail from scheduler noise erodes trust in the suite the first time it does,
and CI runners are noisier than this workstation.

## Suggested fix

Replace the single sample with the **minimum of k repetitions** (k = 3–5) per size point.
The minimum is robust to one-sided noise: a spike makes one sample worse, not the minimum.
This is the standard shape for wall-clock ratio guards.

- Change `time_pseudo` to loop k times and return the min.
- Apply the same to the four inlined `minimal.rs` timing sites.
- Keep the existing `2.8` / `2.5` thresholds and the noise floors — they are sound once the
  sampling is robust. The floor assertions (`DO NOT convert this to a skip`) should stay
  exactly as they are.

Cost: k× the transform time at the guard sizes. At 1.5–8 ms per sample and k=3 this is
single-digit milliseconds per test.

## Severity

**P3** — test hygiene. No production code affected.

## Affected files

- `crates/rskim-core/src/transform/pseudo.rs` — `time_pseudo` (`:2697-2706`); guards at
  `:2741`, `:2778`, `:2810`
- `crates/rskim-core/src/transform/minimal.rs` — inlined timing at `:1458`, `:1468`,
  `:1998`, `:2009`; guards at `:1424`, `:2062`, `:2090`, `:2116`

---

**PR**: #508 (branch `fix/fidelity-parity-overhaul`, 25 commits, HEAD `b6fc565`; `main` untouched)
**Related**: #505 records `088c431` ("quadratic-scaling guard always asserts") from PR #488 —
that commit hardened `test_quadratic_scaling_guard` against silent skipping; this issue is
about the sampling underneath it and covers the same test plus six siblings.

dean0x:
## Output-fidelity overhaul — validated, finished, and described as it actually happened

Base `main` (`a7d12b5`). Every claim below was measured on this branch on 2026-09-02; where a premise of the original plan was refuted, the refutation is recorded rather than smoothed over. Raw controls are always `/usr/bin/git` by absolute path with `SKIM_PASSTHROUGH` scrubbed (PF-026).

### Fixed and verified
- **B1** raw output when the diff parser yields no files (`--raw` 0→549 B, `--dirstat` 0→122 B vs `/usr/bin/git`).
- **C1/C2** `SKIM_PASSTHROUGH=1` / `--passthrough` strip every skim-only flag and exec real git. Verified byte-identical for 8 flags × {diff, show, log} = 24 cells; pinned by a table-driven test.
- **D3** `--json` carries patch bodies (#510). Show side now has containment tests (RED 0/3 at `ae94d3f^`, GREEN 3/3 at HEAD). Both JSON-content tests re-pinned from `08a7892` (branch-only, gone after squash-merge) to `b79f6e3` (on `main`).
- **D1′** the disclosure split, finished as the plan specified: `emit_json_envelope(json, completeness, tool, elided, terminate)` is the only way to write a `--json` envelope, so a handler that omits its `Completeness` does not compile (fn-pointer signature pin — the crate is bin-only, so `compile_fail` doctests and trybuild cannot express it). Eight exits declare a value (diff/show/raw-passthrough `Reencoded`; status/push/fetch/commit and `git log` `Lossy`; two paths — the generic one and `skim log` — derive it from the parse tier); that counts logical exits, and a strict per-call-site count is nine, the ninth being the empty-parse `Reencoded` envelope inside the diff handler (`cmd/git/diff/mod.rs:421`, the B1 fix). A `Lossy` envelope emits one unconditional class-1 stderr line; `Reencoded` stays silent, pinned both ways. `remedy_for` derives the narrowest true remedy. Stdout bytes unchanged on every exit.
- **E2/E4/E5** markers on every truncation path with accurate counts; `--max-lines N` = at most N lines total including the marker, N=1 excepted (ADR-016). E1's 12 RED tests re-derived by checkout: 12/12 fail at `5cc284e^`, 12/12 pass at HEAD.
- **E′2** `--tokens` served 0 bytes, exit 0, when the cascade escalated into an empty mode (`types` on a fn-only Rust file). Empty escalated output no longer satisfies the budget; all-empty views fall back to the raw source; empty input still yields empty output.
- **E′1** the post-guardrail bound enforcement hard-coded a `#` marker with a third phrasing (`# ... (6 lines truncated; use SKIM_PASSTHROUGH=1 to see all)` on a `.ts` file). The marker is spelled once in `rskim_core::elision_marker_line`; every site uses the language prefix; Markdown keeps the hint inside `<!-- -->`; tight `--tokens` budgets put the count on stdout and the remedy on stderr (ADR-016).
- **E6/E7** literal- and fence-aware cuts (#511): a one-pass per-language scanner (`literal_scan`, 18 languages) lets `--max-lines` move a cut back to before a literal/fence opener, `--last-lines` move its window forward past the closer (with the `-n` line map derived from the truncator's real start, PF-019), and `--tokens` snap `best` back after the search. Degenerate case: marker first with `; cut inside a string literal` / `; cut inside a code fence`, then the raw cut, still N lines. Zero of the 76 golden snapshots move — which is evidence that nothing moved that nobody intended to move, not that the output is correct: the goldens' own header states they document current state and are not behavioural assertions, and re-blessing the whole set is one environment variable away (`INSTA_UPDATE=always`).
- **Corpus harness (A5)** re-verified at HEAD, two identical runs: 0 violations in all modes (default 68.2 % fallback, structure/full 97.2 %). Fixture regenerated from 200 `main`-reachable commits (40 of the old 200 were branch-only) and the CI Test Suite job now checks out full history — its depth-1 checkout is why CI failed with "corpus[0] has no parent".
- `--last-lines -n` source-line-number bug (PF-019) fixed.

### Inert (landed, no measurable effect)
- **B2** breadcrumb via `emit_source_line`: 0 of 528 corpus cells changed. Its commit message over-claims `(#512)`. Kept: it removes a documented anti-pattern.

### Reverted
- **B3** container gap fill shipped a structure-mode regression (line numbers and indent lost, `,` orphaned) that its own four tests could not see. Reverted; `has_changes` overlap and `assert_ast_rendered` kept.

### Retargeted
- **D2** egress chokepoint already existed (44 call sites). Dropped. **D5** `print_stdout = "deny"` → `disallowed-methods = ["std::io::stdout"]`. **E3** line map as provenance: premise refuted (one global gate on `line_numbers`); counts computed in source space instead. **E5** "marker is line N+1" inverted the real bug; the rule is N total.

### Dropped
- **D4** (no wrong count could be demonstrated), **E8** — dropped, but not to avoid a published-API change: three shipped here anyway (`truncate_to_token_budget` went from five parameters to seven, its documented empty-string contract became marker-wins, and `TransformConfig` became `#[non_exhaustive]`).

### Phase F — measured before implemented
- F1 docs corrected: five hook early returns skip the marker write; the "never byte loss" claim in `main.rs` was false (below).
- F2 the bare `golangci` alias deleted — a breaking CLI change: `skim golangci` is now `skim golangci-lint`, and the old name fails with a file-not-found error rather than an actionable message. F3 the last two stdout-destination cells pinned (18/18).
- F7: the plan said "1 of 3 predicates honoured, net-new 3"; measured 13 predicates, 4 honoured, net-new 2 — and one of those two (`env FOO=bar prog`) was refuted by its own RED test passing. Only the psql/mysql `require_flag` gap was real; added as the wrapper's D5 gate from the same rule table with a drift guard.
- F8 `Surface::{Explicit, Wrapper}`: the dispatch core is private and cannot be called without declaring a surface (signature pin). No misrouted caller existed; this is preventive.
- F9 `<cmd> | cat` (bare `cat`, sole consumer) is now rewritten, gated on `command_needs_exact_bytes`; `| cat > f`, `| cat | tee f`, `| cat -n`, `| head`, `$(… | cat)` and mixed `&&`/`|` compounds still refuse. All other compound shapes: `&&`/`;` were already rewritten segment-by-segment; pipes, redirects, substitutions, heredocs, process substitution refuse.
- F4 is a no-op (I6 cannot fire on the rewrite surface). F6 preserved: the test pins a deliberate design, not an invitation.
- **F5 — owner decision.** The force-raw marker's "a stale or missing marker costs compression, never bytes" is false, measured: after a same-tool `git status` clears the marker (#514), or with no hook at all, `git log -n 3 | tee f` writes 304 bytes instead of 6803, silently. Docs corrected here; the append-only redesign is byte-safe only in a monotone-OR variant that inverts two pinned tests and costs up to 300 s of compression per tool. Evidence and options are on #514.

### Performance (the `<50 ms / 1000 lines` MUST, never measured before)
criterion `transform_bench`, merge-base `a7d12b5` vs branch, machine idle: 53 benches — 25 improved, 27 unchanged, 1 regressed (`mode_comparison/typescript/Minimal` 25.9 µs → 30.6 µs, +18 %, a Θ(nodes) added traversal from the first-session literal-aware minimal pass `1ad93a8` — `collect_literal_ranges` walks the whole AST and did not exist at the merge-base, doubling TypeScript's traversal count; a linear added term reads as a roughly constant percentage at any one fixture size, so +18 % alone cannot separate O(1) from Θ(nodes), and the quadratic scaling guard discriminates linear from quadratic, not constant from linear). Absolutes at HEAD for 4000 generated TypeScript lines: structure 10.7 ms, minimal 13.7 ms, pseudo 14.6 ms, bounded 11.0 ms — 2.7 to 3.7 ms per 1000 lines. `scaling_minimal` / `scaling_pseudo` / `scaling_bounded` benches added. A CI gate is proposed in #518.

### Found by running the whole workspace
- `crates/rskim-tokens/tests/integration.rs` pinned the pre-E2 contract ("empty string when the budget is smaller than the marker"), as did the public doc on `truncate_to_token_budget`. Bisect: 33/33 at `a7d12b5`, 31/33 at `9058273` — introduced by the prior session's Phase E and never reported because `cargo test` stops at the first failing binary. Tests and doc now state the marker-wins rule.
- Every crate verified: rskim-core, rskim, rskim-tokens, rskim-contract 151/151, rskim-llm 86/86, rskim-compress 215/215, rskim-search 738/738, rskim-proxy 122/122, rskim-research 124/124, rskim-bench 196/196, doctests.

### Behaviour changes reviewers should know about
- Every Full/Degraded-tier `--json` run (`skim eslint --json`, `skim gh --json`, `skim log --json`) now prints one stderr line; `skim log` returns 141 without an analytics row on a closed pipe.
- `test_skim_test_cargo_in_this_repo` (a nested `cargo test` under a 120 s assert_cmd timeout) is timing-dependent on this machine after core changes; it passes once the inner build is warm.

### Not in this PR
The F5 marker redesign (#514); the multi-span AST truncator's `span_end` clamp (recorded in the E6 design); Snyk SAST (the MCP server cannot spawn headlessly — needs an interactive `snyk auth`). Follow-ups filed: #517 (`nextest --no-capture` elision), #518 (CI perf gate), #519 (ADR-001 vs node priority on small files).

### Related issues
Closes #510, closes #511, closes #513. Refs #512 (only the B1 part is addressed here), #514, #517, #518, #519.


dean0x:
## Code Review — Cycle 1

Full summary withheld (public repository).

| Category | CRITICAL | HIGH | MEDIUM | LOW | Total |
|----------|----------|------|--------|-----|-------|
| Blocking (in your changes) | 3 | 22 | 24 | - | 49 |
| Should Fix (code you touched) | - | 2 | 6 | - | 8 |
| Pre-existing (informational) | - | - | 4 | 1 | 5 |
| **Total** | **3** | **24** | **34** | **1** | **62** |

Full report: .devflow/docs/reviews/fix-fidelity-parity-overhaul/2026-09-03_2025/review-summary.md (not committed; ask the author)
*Posted by [devflow](https://github.com/dean0x/devflow) · cycle 1*

dean0x:
## Resolution Summary

**Verification gate result:** PASS  
**Publication mode:** STUB (public repository — full report withheld)

### Summary Statistics
| Metric | Count |
|--------|-------|
| Total issues triaged | 159 |
| Fixed | 103 |
| False positives | 1 |
| By design | 2 |
| Deferred (filed as issues) | 19 |
| Test-hygiene incomplete | 8 |
| Blocked | 0 |
| Duplicates collapsed | 26 |

### Key Outcomes

**All three CRITICALs are fixed:**
1. **env/printenv redaction bypass** (PF-012, second occurrence) — three new raw-serve paths re-exposed credentials; fixed via `redaction_is_mandatory` predicate. Owner decision: narrow fix for `env FOO=1 env` arm only; `env FOO=1 sh -c env` remains as documented accepted limitation.
2. **`--max-lines 1` served zero content lines** — reachable via `head -1 file.rs` rewrite hook; fixed by delegating bound arithmetic to `rskim-core`.
3. **`rskim-core` 2.11.0 published-API breaks** — undisclosed new parameter (now 7th), missing `#[non_exhaustive]`, inverted return contract. Fixed with annotations; major bump required.

### Verification Gate Results
Gate ran serially in the main loop after all agents finished:
- **rskim-core**: 773 unit tests + 16 doctests ✓
- **rskim**: 4801 tests (3 skipped) ✓  
- **rskim-tokens**: 33 tests ✓
- **rskim-contract**: 151 tests ✓
- **clippy --all-features**: 0 warnings ✓
- **cargo fmt --check**: clean ✓

**Gate caught 10 defects** that static reasoning missed:
- 2 compilation errors (typographic quotes, unicode-escape brace-doubling)
- 4 clippy violations (expect_err in deny-all module, hand-rolled div_ceil, redundant collapsible-if, to_string when Display is available)
- Real regressions: `-n` flag on pseudo view off-by-one in line annotations, line-map blank-drop normalization inverted
- Expansion regression: `gh` net-savings guard stop-firing (PF-011)
- 3 wrong test premises (META subcommand on wrapper surface excluded by construction, hard-coded marker wording, stale golden)

### Three Owner Decisions

1. **Narrow `env FOO=1` redaction fix** (testing-14) — guard only `env`/`printenv` children; other children still exec raw. Residual `env FOO=1 sh -c env` pinned as documented accepted limitation.
2. **Delegate bound arithmetic to `rskim-core`** (reliability-1) — single spelling avoids PF-033 parallel-array defect; freed reliability-3.
3. **Extend ADR-016 N=1 carve-out to `--last-lines`** (reliability-5) — head/tail symmetry; re-derived pinned test instead of correcting docs to match defect.

### Work Deferred to Follow-up Issues
Coherence grouping to reduce issue count and surface interdependencies:
- **#530** — Transform core: complexity-8, complexity-14, rust-6, performance-7, architecture-22, security-6, performance-1, architecture-5 (8 items)
- **#531** — Dispatch & registry: complexity-12, complexity-9, consistency-14, consistency-19, architecture-13, architecture-8 (6 items; architecture-8 is load-bearing for `gh` fidelity)
- **#532** — Test hygiene: testing-4, testing-5, testing-8 (relates to #514), testing-10, testing-12, testing-18, testing-19, testing-22 (8 in-flight items)
- **#533** — Contract decisions: complexity-3, complexity-6, consistency-16, reliability-9, documentation-23 (5 items needing policy calls, not edits)
- **#534** — git/diff/render.rs: architecture-19 (3382-line module decomposition; may interact with #509, #512)

Full report: `.devflow/docs/reviews/fix-fidelity-parity-overhaul/2026-09-03_2025/resolution-summary.md` (not committed; ask the author)

---

*Posted by [devflow](https://github.com/dean0x/devflow)*

dean0x:
## ⚠️ PR Description Corrections Needed

The PR body survives squash-merge, so these corrections should be applied by the author to prevent permanent record errors:

### Factual Corrections

1. **Performance Regression Attribution (performance-5)**
   - **Current:** "+18% Minimal benchmark regression" is attributed to "a constant per-file cost"
   - **Correction:** This is a **Θ(nodes) added traversal** — `collect_literal_ranges` walks the whole AST and did not exist at the merge-base, doubling TypeScript's traversal count. A linear added term shows as a roughly constant *percentage* at any single fixture size, so `+18%` cannot distinguish O(1) from Θ(nodes). The quadratic scaling guard (PF-020) discriminates linear from quadratic, not constant from linear, so it cannot falsify the claim.

2. **Exit Code Count (documentation-8)**
   - **Current:** "Seven exits declare a value"
   - **Correction:** **Eight** exits declare a value

3. **Generic Path Count (documentation-8)**
   - **Current:** "one generic path"
   - **Correction:** **Two** tier-derived paths

4. **Golden Snapshot Claim (testing-17)**
   - **Current:** "Zero of the 76 golden snapshots move" as correctness evidence
   - **Correction:** This is evidence of **no change**, not correctness. The goldens' own header states they document current state and are not behavioural assertions; a bulk re-bless is one `DEVFLOW_BLESS=1` environment variable away.

### Breaking CLI Changes

5. **Subcommand Rename**
   - `skim golangci` → `skim golangci-lint` is a **breaking CLI change** (the old name now fails with a file-not-found error rather than an actionable message)

### API Signature Changes

6. **New Public API Parameter**
   - `truncate_to_token_budget` now takes a **seventh** parameter (the description claims six; see the commit for the actual signature)

### Measurement Gaps

7. **`--tokens` Remedy Claim (documentation-2)**
   - **Current:** The `SKIM_PASSTHROUGH=1` remedy reaches stderr "unconditionally"
   - **Status:** This claim was measured false (2 of 9 mode×budget cells emitted nothing). The cascade half is fixed; CLAUDE.md's wording was deliberately left weaker pending re-measurement of the `{full,structure,pseudo} × {3,5,10}` matrix. Do not restore "unconditionally" without re-running that measurement.

---

These are documentation/wording corrections only, not code changes. They should be applied before the PR merges or recorded in a follow-up issue if the PR is already merged.

Issue #516, opened 2026-09-02
Make root-descending tree-sitter Node APIs unrepresentable in transform walkers (newtype facade)


dean0x:
## Context

\`rskim-core\`'s \`transform/minimal.rs\` and \`transform/pseudo.rs\` walkers were quadratic (one cubic) before #488 linearized them. The quadratic term was inside tree-sitter's C code — a \`TSNode\` carries no parent pointer, so \`parent()\`, \`prev_sibling()\`, \`next_sibling()\`, \`next_named_sibling()\`, and \`named_child(i)\` all re-derive their answer by descending from the tree root and scanning the parent's child list from position 0. The defective code ran the same number of Rust-level steps and produced byte-identical output; only the C-side work differed.

Because of that, no in-process mechanism (operation counter, visited-set, output equivalence) can catch a reintroduction: a counter has to live in the code a regression replaces, and the output does not change. The original guards used wall-clock ratios, and CI falsified four of them within a week (2.89x, 4.94x, 3.12x, 3.82x against a 2.8 gate). The follow-up replaced every wall-clock guard with a **source-level contract test** that forbids the root-descending API set in the production regions of those two files, plus exact-artifact assertions per test.

## Why this issue exists

The source contract is a lint, not a proof. It catches the enumerated construct; a novel quadratic shape (e.g. \`collect()\` + linear search per node) slips past it, and a future exception added for one legitimate call widens the hole for all. The strictly stronger fix is to make the construct **unrepresentable** at the type level.

## Proposed change

Introduce a facade in \`rskim-core::transform\`:

\`\`\`rust
/// A tree-sitter node restricted to O(1) accessors. No \`parent()\`, no sibling
/// navigation, no indexed child access — the walker must thread position via
/// \`WalkPosition\` / \`TreeCursor\` instead.
pub(crate) struct WalkNode<'t>(Node<'t>);

impl<'t> WalkNode<'t> {
    pub fn kind(&self) -> &'static str;
    pub fn start_byte(&self) -> usize;
    pub fn end_byte(&self) -> usize;
    pub fn is_named(&self) -> bool;
    pub fn child_by_field_name(&self, f: &str) -> Option<WalkNode<'t>>;  // descends INTO self: O(depth of field), fine
    pub fn children<'c>(&self, cursor: &'c mut WalkCursor<'t>) -> impl Iterator<Item = WalkNode<'t>>;
[...]

PR #520, opened 2026-09-02
test(rskim-core): replace wall-clock scaling guards with a deterministic source contract

dean0x:
## Problem

`crates/rskim-core` carried **ten** wall-clock timing assertions in tests — not the four that were flaking. All ten shared one pattern: build a synthetic source of size N (and usually 2N), take a **single un-warmed** `Instant::now()` / `elapsed()` sample per size, and assert either an absolute millisecond budget or a `t2/t1 < 2.8` doubling ratio. All ten ran inside the same ~1 s window of the same `rskim-core --lib` binary, so they competed with each other for the same cores.

Four of the ratio guards failed CI on GitHub-hosted 4-vCPU runners within a week of landing, with observed ratios of **2.89x, 4.94x, 3.12x and 3.82x against a 2.8 gate**; one failed twice consecutively on an isolated re-run. The ceilings were calibrated on developer hardware roughly **2x faster than the CI runner**, so the headroom the thresholds were designed with never existed in CI.

The ten (all now converted):

| # | Test | File |
|---|------|------|
| 1 | `test_large_header_block_linear_time` | `transform/minimal.rs` |
| 2 | `test_quadratic_scaling_guard` | `transform/minimal.rs` |
| 3 | `test_go_leading_comment_run_linear_time` | `transform/minimal.rs` |
| 4 | `test_go_leading_comment_run_scaling_guard` | `transform/minimal.rs` |
| 5 | `test_go_doc_block_scaling_guard` | `transform/minimal.rs` |
| 6 | `test_go_pseudo_leading_comment_run_scaling_guard` | `transform/minimal.rs` |
| 7 | `test_typescript_pseudo_semicolon_walk_scaling_guard` | `transform/pseudo.rs` |
| 8 | `test_python_pseudo_return_type_walk_scaling_guard` | `transform/pseudo.rs` |
| 9 | `test_cpp_pseudo_template_walk_scaling_guard` | `transform/pseudo.rs` |
| 10 | `test_python_minimal_large_header_linear_time` | `tests/integration.rs` |

Widening the ratio, adding warm-up, or taking best-of-N would all keep the clock — and therefore keep the flake. This removes the clock.

## Why not an operation counter

An instrumented counter cannot discriminate here, and this is the load-bearing fact of the design.

The quadratic term lives inside tree-sitter's C code, not ours. A `TSNode` carries no parent pointer, so `parent()`, `prev_sibling()`, `next_sibling()`, `next_named_sibling()` and `named_child(i)` each re-descend from the tree root and rescan the parent's child list from position 0. The defective and the fixed walker execute the **same number of Rust statements** and produce **byte-identical output**. Any counter able to tell them apart would have to live inside the very code a regression replaces.

So the checkable invariant is not *how many operations ran* but *which API the source calls* — a property of the text, available before anything executes.

## The fix

Three parts, and **zero production-code changes**: every diff hunk is inside a `#[cfg(test)] mod tests` or `tests/integration.rs`.

**1. A source-level complexity contract** (`transform/mod.rs`). `contract_transform_walkers_use_no_root_descending_node_apis` `include_str!`s `minimal.rs` and `pseudo.rs`, blanks comments and string literals, cuts the production region at the **last** `#[cfg(test)]`, and asserts eight root-descending tokens are absent:

```
.parent()  .prev_sibling(  .next_sibling(  .prev_named_sibling(
.next_named_sibling(  .named_child(  .child(  .rfind(
```

Blanking is not optional: both files *document* the banned APIs in rustdoc (`WalkPosition`'s fields are specified as "equivalent to `node.parent()`"), and `is_doc_comment` holds `"/**"`, `"/*!"` and `"///"` as string literals — an unsanitised scan would false-positive on the prose explaining the ban, and `"/**"` would open a comment swallowing live code.

The genuinely O(1) traversal set (`Node::walk`, `TreeCursor::goto_first_child` / `goto_next_sibling`, `Node::children`, `Node::named_children`, `Node::child_count`, `child_by_field_name` on a node already in hand) is deliberately excluded, and `blank_comments_and_strings_self_test` pins that none of those forms trips a token — the `.` prefix and `(` suffix are what keep `.child(` from matching `.child_by_field_name(`, `.named_child(` from matching `.named_children(`, and `.next_sibling(` from matching `goto_next_sibling(`.

Each failure names the file, line, token and reason, and tells the next engineer that a provably-O(1) call at that site may be added as an explicit exception *together with the measurement that proves it*.

**2. The ten timing bodies became exact-artifact assertions.** Every pre-existing behavioural assertion is kept verbatim. What replaced the clock:

- precomputed byte tables (`compute_header_end_byte`, `compute_go_doc_comment_starts`) compared against offsets derived independently from the fixture **text**;
- exact `(start, end)` removal-range sets from the walkers, compared against text-derived expectations;
- a **single-visit** `node_count`, compared against an independently-derived AST node count (`count_all_nodes`, an explicit stack over `Node::children`). This is the assertion a re-walk regression trips while leaving the output byte-identical.

N drops to 256/512 (128/256 for Go and C++). The cubic tripwires existed only to stop a reintroduced walk grinding CI for hours; the contract now forbids the construct outright.

Every test is renamed to say what it now asserts, and each carries a rustdoc paragraph stating explicitly that it does **not** discriminate O(N) from O(N²) on its own — the defect was output-preserving, and that job belongs to the contract test.

**3. Four now-dead helpers removed:** `time_go_minimal`, `time_go_pseudo`, `assert_no_cubic_regression` (`minimal.rs`), `time_pseudo` (`pseudo.rs`). The two cross-reference comments in `integration.rs` that named `test_quadratic_scaling_guard` now point at the contract test.

The empirical before/after series that motivated the original fixes is untouched — it already lives in the production rustdoc on `is_go_doc_comment` and `compute_header_end_byte`. `TransformResult::duration_ms` in `lib.rs` is untouched: that is production reporting, not a gate.

## Verification

All commands scoped to `-p rskim-core`, `-j 4`, `SKIM_PASSTHROUGH=1`.

```
$ cargo nextest run -p rskim-core -j 4 --no-fail-fast
    Starting 557 tests across 8 binaries
     Summary [   1.810s] 557 tests run: 557 passed, 0 skipped
exit 0

$ cargo test -p rskim-core --doc
test result: ok. 16 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
exit 0

$ cargo clippy -p rskim-core --all-targets -- -D warnings
exit 0

$ cargo fmt -- --check
exit 0

$ grep -rn "Instant\|elapsed()\|Duration::from" \
      crates/rskim-core/src/transform/ crates/rskim-core/tests/integration.rs
exit 1   (no matches — the only remaining clock in the crate is
          lib.rs:335 TransformResult::duration_ms, production reporting)
```

## Mutation results

Each mutation is a one-line edit to **production** code, applied, run, then reverted (`git checkout --`, tree verified clean by `git status --porcelain`).

| # | Mutation | Result | Caught by |
|---|----------|--------|-----------|
| M1 | `compute_header_end_byte`: `TreeCursor` loop → `named_child(i)` index loop | `557 tests run: 556 passed, 1 failed` | **contract only** — token `` `.named_child(` `` at `minimal.rs:540` |
| M1b | `is_module_header_comment`: threaded `depth` → `node.parent()` kind test | `290 tests run: 289 passed, 1 failed` | **contract only** — token `` `.parent()` `` |
| M2 | `is_go_doc_comment`: `binary_search` → per-node `next_named_sibling()` walk | `290 tests run: 281 passed, 9 failed` | contract (token `` `.next_named_sibling(` ``) **+ 8 behavioural** — see note |
| M3 | `collect_noise_ranges`: `pos.parent_kind` → `node.parent()` | `290 tests run: 289 passed, 1 failed` | **contract only** — token `` `.parent()` `` |
| M4 | C++ template start: `pos.prev_sibling_*` → `node.prev_sibling()` | `290 tests run: 289 passed, 1 failed` | **contract only** — token `` `.prev_sibling(` `` |
| M5 | `adjust_range_for_line_removal`: `partition_point` → `rfind('\n')` | `290 tests run: 289 passed, 1 failed` | **contract only** — token `` `.rfind(` `` |
| M6 | C++ template range starts at `node.start_byte()` unconditionally | `290 tests run: 286 passed, 4 failed` | `test_cpp_pseudo_template_ranges_are_exact_and_single_visit`, `test_cpp_pseudo_no_orphaned_template`, `test_cpp_pseudo_strips_template_function`, `test_cpp_template_parameter_list_skips_recursion` |
| M7 | `collect_noise_ranges`: descend into each child twice | `290 tests run: 286 passed, 4 failed` | `test_typescript_pseudo_semicolon_ranges_…`, `test_cpp_pseudo_template_ranges_…`, `test_python_pseudo_return_type_ranges_…`, `test_large_doc_blocks_sections_pseudo` |
| M8 | `collect_removable_comments`: descend into each child twice | `290 tests run: 284 passed, 6 failed` | `test_python_module_header_precompute_…`, `test_python_large_header_block_survives_transform`, `test_go_doc_comment_precompute_…`, `test_go_leading_comment_run_…` (×2), `test_large_doc_blocks_sections_minimal` |
| M9 | `go_gap_breaks_run` returns `true` unconditionally | `290 tests run: 284 passed, 6 failed` | `test_go_doc_comment_precompute_is_exact_and_single_visit` (`starts.len() == 3n`) + 5 pre-existing Go classification tests |
| M10 | `production_region`: `rfind` → `find` (cut at the FIRST `#[cfg(test)]`) | `290 tests run: 289 passed, 1 failed` | `production_region`'s own gate assertion: ``pseudo.rs: the last `#[cfg(test)]` does not gate `mod tests` (next item is Some("pub(crate) fn transform_pseudo("))`` |

**The headline result, stated plainly:** M1, M1b, M3, M4 and M5 are the historical defects reintroduced verbatim, and each is caught by **exactly one test — the source contract**. All ten exact-artifact tests stay green under those five mutations, because the output really is byte-identical. That is the empirical confirmation of the design's central claim, and it is why the contract exists rather than a counter or a wider ratio.

**One deviation from the predicted outcome, recorded rather than glossed:** M2 as specified was expected to be contract-only, but it also fails 8 behavioural tests. The replacement body walks forward siblings looking for `is_go_declaration` **without** the blank-line-gap rule that `go_gap_breaks_run` enforces, so it is not an output-preserving reproduction of the historical defect — it changes Go doc-comment classification. The contract still catches it on `.next_named_sibling(`; the extra failures are a property of that particular replacement body, not a counter-example to the design.

M6–M10 confirm the per-test assertions and the guard's own tripwires are not vacuous.

**A note on M10.** `production_region` asserts *both* that the tail after the cut contains `\nmod tests {` **and**, more tightly, that the marker it cut at is the one immediately gating `mod tests` (attribute lines allowed between). The containment check alone does not catch `find`: cutting at `pseudo.rs`'s earlier `#[cfg(test)]` (the one on `transform_pseudo`) leaves a tail that still contains `\nmod tests {` further down, so ~440 lines of production code would silently drop out of the policed region while the guard reported success. The tighter assertion is what M10 trips.

## Follow-up

The contract is a **lint, not a proof**: it pins the construct, not the complexity, and a novel super-linear pattern avoiding all eight tokens would pass. The stronger form is a type-level `WalkNode` facade exposing only the O(1) traversal set, which makes the forbidden construct *unrepresentable* rather than merely detectable. Tracked separately.

## Constants pinned

Every constant in the new assertions was derived by reading the grammar and the fixture shapes, then pinned by running. **None required adjustment** — all twelve tests passed on the first execution:

| Constant | Derived value | After running |
|----------|---------------|---------------|
| Python header end byte | `source.len() - PY_HEADER_TAIL.len() - 1`, `PY_HEADER_TAIL = "def f(x): return x\n"` | unchanged |
| Python header removable ranges | `[]` (whole block is the module header) | unchanged |
| Go doc-comment starts (doc-block fixture) | `source.match_indices("//")`, length `3n`, strictly ascending | unchanged |
| Go doc-block removable ranges | `[]` (every comment is a doc comment) | unchanged |
| Go leading-run doc starts | `[]` (`package_clause` is not an `is_go_declaration` kind) | unchanged |
| Go leading-run removal ranges | one `(start, start + line_len)` per `//`, `n` of them | unchanged |
| TypeScript semicolon ranges | one `(i, i+1)` per `;`, `n` of them, **and nothing else** | unchanged |
| Python parameter annotations | `": int"` → `(i, i + 5)`, `2n` of them; `-> int` preserved (ADR-007) | unchanged |
| C++ template prefix | `"template<typename T> "` (21 bytes, trailing space consumed, stops before `T`) | unchanged |
| Node-visit counts | `== count_all_nodes(root)` for the minimal walker; `<=` for the pseudo walker (it returns early on stripped kinds) | unchanged |
| Integration header count | 500 surviving `# Header comment ` lines, derived from the fixture rather than hardcoded | unchanged |

### DIFF
--- a/crates/rskim-core/src/transform/minimal.rs
+++ b/crates/rskim-core/src/transform/minimal.rs
@@ -816,6 +816,61 @@ mod tests {
         panic!("could not find comment #{n} in root named children");
     }
 
+    // Helper: parse Go source into a tree-sitter Tree.
+    fn parse_go(source: &str) -> Tree {
+        let mut parser = crate::Parser::new(Language::Go).unwrap();
+        parser.parse(source).unwrap()
+    }
+
+    /// Total AST node count, derived independently of the walkers under test.
+    ///
+    /// Explicit stack + `Node::children`, so the count never depends on the
+    /// traversal the walker performs. That independence is the whole point: it is
+    /// what lets `node_count == count_all_nodes(root)` detect a walker that visits
+    /// the same node twice, which is invisible in the output.
+    fn count_all_nodes(root: Node) -> usize {
+        let mut n = 0usize;
+        let mut stack = vec![root];
+        while let Some(node) = stack.pop() {
+            n += 1;
+            let mut c = node.walk();
+            for ch in node.children(&mut c) {
+                stack.push(ch);
+            }
+        }
+        n
+    }
+
+    /// Drive the production comment walker directly, composing `CommentWalkContext`
+    /// exactly the way `transform_minimal` does.
+    ///
+    /// Returns the raw (unadjusted) removal ranges and the number of nodes the
+    /// walker visited. The visit count is the instrument a re-walk regression trips:
+    /// descending twice leaves the ranges — and therefore the output — unchanged
+    /// while doubling the count.
+    fn collect_minimal_ranges(
+        source: &str,
+        tree: &Tree,
+        language: Language,
+    ) -> (Vec<(usize, usize)>, usize) {
+        let root = tree.root_node();
+        let header_end_byte = compute_header_end_byte(root, source, language);
+        let go_doc_comment_starts = compute_go_doc_comment_starts(root, source, language);
+        let mut ranges: Vec<(usize, usize)> = Vec::new();
+        let mut node_count: usize = 0;
+        let mut ctx = CommentWalkContext {
+            ranges: &mut ranges,
+            node_count: &mut node_count,
+            classification: CommentClassification {
+                header_end_byte,
+                go_doc_comment_starts: &go_doc_comment_starts,
+            },
+        };
+        collect_removable_comments(root, source, language, &mut ctx, 0, false)
+            .expect("comment walk must succeed on these fixtures");
+        (ranges, node_count)
+    }
+
     // ── compute_header_end_byte correctness ─────────────────────────────────
 
     #[test]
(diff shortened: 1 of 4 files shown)


dean0x:
Follow-up tracked in #516 — the `WalkNode` newtype facade that makes the forbidden root-descending calls a compile error rather than a contract-test failure. The source contract in this PR is the lint; #516 is the proof.

